PR #2, opened 2026-01-30
fix(docs): Update README

namrataghadi-galileo:


CONCEPTS.md
```diff
@@ -0,0 +1,285 @@
+## Concepts
```
lan17, line 1: should this be instead part of reference doc?

namrataghadi-galileo: We discussed in this morning's meeting to add to the main READMe. I think you missed the initial part of the meeting.
The decision was to add one liner and link to Concepts.md in main readme. Explain concepts in Concepts.md file. cc @nachiket-galileo

coverage-server.xml
```diff
@@ -0,0 +1,1917 @@
+<?xml version="1.0" ?>
```
lan17: please remove code cov xml files from this pr and also add them to .gitignore.

namrataghadi-galileo: done

PR #3, opened 2026-01-31
fix(docs): Add Example for Evaluator Extension

namrataghadi-galileo:
Deep Eval Evaluator example to show how Evaluator can be extended to create custom evaluators.


examples/deepeval/evaluator.py
```diff
@@ -0,0 +1,298 @@
...
+from typing import Any
+
+from agent_control_models import (
+    Evaluator,
+    EvaluatorMetadata,
+    EvaluatorResult,
+    register_evaluator,
+)
+
+# Import config - handle both relative and absolute imports
```
lan17: why do we need this?  we probably want to avoid having developers to deal with that.  

Lets fix example so its just using one or the other?  I prefer .config since its signaling local module

examples/deepeval/pyproject.toml
```diff
@@ -0,0 +1,43 @@
...
+
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+
+[tool.hatch.build.targets.wheel]
+packages = ["."]
+
+[tool.uv.sources]
+agent-control-models = { path = "../../models", editable = true }
```
lan17: example should be demonstrating depending on actual pypi pcakages

namrataghadi-galileo: Yea.. this was previously when we had not migrated from plugin to evaluator. I will make this change now

examples/deepeval/evaluator.py
namrataghadi-galileo: done

examples/deepeval/pyproject.toml
namrataghadi-galileo: done

PR #6, opened 2026-01-31
feat(server)!: add time-series stats and split API endpoints

abhinav-galileo:
BREAKING CHANGES:
- Response structure changed: `total_*` fields moved to `totals` object
- `control_id` param removed from GET /stats
- New endpoint GET /stats/controls/{id} for control-level stats

New features:
- Time-series data via `include_timeseries=true` parameter
- Auto bucket sizing based on time range (6-28 data points)
- Single optimized CTE query for stats + timeseries

API changes:
- GET /stats returns `totals` + `controls[]` (agent-level)
- GET /stats/controls/{id} returns `control_id`, `control_name`, `stats` (control-level)

Other changes:
- Remove MemoryEventStore (PostgreSQL-only testing)
- Add StatsTotals, ControlStatsResponse, TimeseriesBucket models
- Update docs with quick reference table and examples

### DIFF
--- a/models/src/agent_control_models/observability.py
+++ b/models/src/agent_control_models/observability.py
@@ -380,12 +380,12 @@ class StatsRequest(BaseModel):
 
     Attributes:
         agent_uuid: Agent to get stats for
-        time_range: Time range (1m, 5m, 15m, 1h, 24h, 7d)
+        time_range: Time range (1m, 5m, 15m, 1h, 24h, 7d, 30d, 180d, 365d)
         include_timeseries: Whether to include time-series data points
     """
 
     agent_uuid: UUID = Field(..., description="Agent UUID")
-    time_range: Literal["1m", "5m", "15m", "1h", "24h", "7d"] = Field(
+    time_range: Literal["1m", "5m", "15m", "1h", "24h", "7d", "30d", "180d", "365d"] = Field(
         default="5m", description="Time range"
     )
     include_timeseries: bool = Field(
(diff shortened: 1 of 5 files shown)


models/src/agent_control_models/observability.py
```diff
@@ -393,42 +393,121 @@ class StatsRequest(BaseModel):
...
         default_factory=dict,
         description="Action breakdown for matches: {allow, deny, warn, log}",
     )
+    timeseries: list[TimeseriesBucket] | None = Field(
+        default=None,
+        description="Time-series data points (only when include_timeseries=true)",
+    )
+
+
+class StatsResponse(BaseModel):
```
lan17, line 474: With the new `/stats` shape, we should also update `StatsRequest` to remove `control_id` and add `include_timeseries`. Right now the request model is still the legacy params, so OpenAPI/types will be inconsistent.

server/src/agent_control_server/observability/store/postgres.py
```diff
@@ -135,86 +192,162 @@ async def query_stats(
...
+                    SELECT generate_series(
+                        :cutoff,
+                        :now - CAST(:bucket_interval AS interval),
+                        CAST(:bucket_interval AS interval)
+                    ) as bucket
+                ),
+                -- Time-series aggregation (only buckets with data)
+                bucket_stats AS (
+                    SELECT
+                        date_bin(CAST(:bucket_interval AS interval), timestamp, :cutoff) as bucket,
```
lan17, line 233: `date_bin` is PostgreSQL 14+. If we intend to support older PG versions (or SQLite via `DB_URL`), we should either guard/document the minimum PG version here or add a fallback (e.g., `date_trunc` + arithmetic) for older installs.

abhinav-galileo: Looks like PostgreSQL 14 is supported till Nov 2026.

Will document the min requirement, and explore fallback options in a new PR.

models/src/agent_control_models/observability.py
abhinav-galileo: Done

PR #15, opened 2026-02-03
feat(sdk): Get Agent Controls from SDK Init

nachiket-galileo:
### Summary

agent_control.get_server_controls() returns the cached list of controls that were fetched from the Agent Control server.

#### Purpose:
Returns controls fetched during init() or the last refresh_controls() call
Used internally by the @control() decorator to determine which controls to apply
Useful for debugging/inspecting what controls are active for your agent

#### Returns:
list[dict[str, Any]] — List of control dictionaries if controls are cached
None — If init() hasn't been called yet or no controls were fetched

#### Usage 

```
import agent_control

# Initialize the SDK (fetches controls from server)
agent_control.init(
    agent_name="my-agent",
    agent_id="agent-123",
    server_url="http://localhost:8000"
)

# Inspect the cached controls
controls = agent_control.get_server_controls()
if controls:
    print(f"Loaded {len(controls)} controls")
    for c in controls:
        print(f"  - {c['name']} (execution: {c['control'].get('execution', 'server')})")
```

### DIFF
--- a/sdks/python/src/agent_control/control_decorators.py
+++ b/sdks/python/src/agent_control/control_decorators.py
@@ -463,7 +463,11 @@ def _get_server_controls() -> list[dict[str, Any]] | None:
     try:
         import agent_control
         return agent_control.get_server_controls()
-    except Exception:
+    except Exception as exc:
+        logger.debug(
+            "Unable to access cached server controls; proceeding without local cache.",
+            exc_info=exc,
+        )
         return None
 
 


sdks/python/src/agent_control/control_decorators.py
```diff
@@ -199,6 +204,80 @@ async def _evaluate(
...
+                    controls=controls,
+                )
+
+                # Convert result to dict format expected by process_result
+                return {
+                    "is_safe": result.is_safe,
+                    "confidence": result.confidence,
+                    "reason": result.reason,
+                    "matches": [
+                        {
```
lan17, line 235: you could just do m.dump_json here to shorten code, right?

sdks/python/src/agent_control/control_decorators.py
```diff
@@ -199,6 +204,80 @@ async def _evaluate(
...
+            except ImportError:
+                logger.warning(
+                    "Local evaluation not available (missing agent_control_engine). "
+                    "Falling back to server-only evaluation. "
+                    "Controls with execution='sdk' will be skipped."
+                )
+            except Exception as e:
+                logger.warning(
+                    "Local evaluation failed: %s. Falling back to server-only evaluation.",
+                    e,
```
lan17, line 277: pass exc_info=True to print stack trace

sdks/python/src/agent_control/control_decorators.py
```diff
@@ -379,6 +458,15 @@ def _log_single_control(
     )
 
 
+def _get_server_controls() -> list[dict[str, Any]] | None:
+    """Get the cached server controls from agent_control module."""
+    try:
+        import agent_control
+        return agent_control.get_server_controls()
+    except Exception:
+        return None
```
lan17, line 471: should definitely log here

sdks/python/src/agent_control/__init__.py
```diff
@@ -180,18 +179,126 @@ def __init__(
...
+            client,
+            _current_agent,
+            steps=[]
+        )
+        _server_controls = response.get('controls', [])
+        logger.info("Refreshed %d control(s) from server", len(_server_controls or []))
+        return _server_controls
+
+
+def refresh_controls() -> list[dict[str, Any]] | None:
```
lan17, line 242: should this be called in a loop somewhere ?

sdks/python/src/agent_control/control_decorators.py
nachiket-galileo: @lan17 I think we need to create very specific fields so returning entire json may not work ?

PR #23, opened 2026-02-06
feat(sdk): Auto-populate init() steps from @control() decorators

nachiket-galileo:
## Problem
`agent_control.init()` required manually passing `steps=[...]` even when that metadata already existed on `@control()`-decorated functions. This created duplication and drift risk.

## Goals
- Keep developer UX simple: no extra schema arguments on `@control()`.
- Auto-populate `init()` steps from registered controls.
- Improve schema inference quality and reliability.
- Preserve explicit override behavior for advanced users.

### 1) Auto step discovery from `@control()`
- `@control()` now registers function metadata at decoration time via `_control_registry.register(func, policy)`.
- Registered metadata includes:
  - step name (`.name` / `.tool_name` for tools, otherwise `func.__name__`)
  - step type (`tool` vs `llm`)
  - description (first docstring line)
  - optional policy metadata

### 2) Schema derivation isolated into `_schema_derivation.py`
- Added a dedicated module for input/output JSON schema inference:
  - unwraps callables with `inspect.unwrap`
  - resolves hints via `get_type_hints(..., include_extras=True)`
  - infers input schema from signature using dynamic Pydantic model (`create_model`)
  - infers output schema via `TypeAdapter(return_type).json_schema()`
- `args_schema` override remains first priority for input schema when available.

### 3) Best-effort fallback behavior (with warnings)
- If type-hint resolution or schema inference fails:
  - emits warning logs
  - falls back to permissive schemas
    - input: `{"type": "object", "additionalProperties": true}`
    - output: `{}`
- This keeps registration/init resilient instead of failing hard.

### 4) Deferred schema derivation timing
- Registry stores function metadata at registration time, but derives JSON schemas when steps are retrieved.
- This improves forward-reference behavior by deriving closer to `init()` time.

### 5) `init()` merge semantics
- `init()` now merges:
  - explicit `steps` passed by user
  - auto-discovered steps from registry
- Deduplication key is `(type, name)`.
- Explicit steps win on exact collisions.
- A warning is logged when auto steps are overridden.

### 6) Example added
- Added LangGraph example demonstrating auto-derived schemas without explicit `steps`:
  - `examples/langchain/langgraph_auto_schema_agent.py`
- Also fixed `examples/langchain/pyproject.toml` local evaluator path for `uv run` in that example directory.

## Behavior Summary
- Users can call `agent_control.init(...)` without `steps` and still register typed tool/llm step schemas automatically.
- Users can still pass explicit `steps` to override specific auto-derived entries when needed.
- No new schema arguments required on `@control()`.

### Schema derivation
- async callables
- wrapped/decorated callables (`inspect.unwrap` path)
- keyword-only params
- zero-parameter input schema shape
- `Annotated[...]` metadata preservation
- union/literal/any cases
- enum/dataclass input+output smoke coverage
- set/tuple/default-None edge cases
- JSON Schema validity checks (Draft 2020-12)
- fallback/error branches and warning assertions

### Registry/init wiring
- end-to-end `init()` wiring with real `@control()` auto-discovery into `register_agent(..., steps=...)`
- unresolved-hint fallback behavior through `init()` (warning + permissive schema)
- wrapper/decorator-ordering registration case
- explicit-vs-auto merge precedence by `(type, name)`

## Files Touched
- `sdks/python/src/agent_control/_control_registry.py`
- `sdks/python/src/agent_control/_schema_derivation.py`
- `sdks/python/src/agent_control/control_decorators.py`
- `sdks/python/src/agent_control/__init__.py`
- `sdks/python/tests/test_schema_derivation.py`
- `sdks/python/tests/test_control_registry.py`
- `sdks/python/tests/test_init_step_merge.py`
- `examples/langchain/langgraph_auto_schema_agent.py`
- `examples/langchain/pyproject.toml`
- `examples/README.md`
- `examples/langchain/README.md`

### DIFF
--- a/sdks/python/tests/test_control_registry.py
+++ b/sdks/python/tests/test_control_registry.py
@@ -2,19 +2,19 @@
 
 from __future__ import annotations
 
-from collections.abc import Generator
+import functools
 import logging
+from collections.abc import Generator
 from typing import Any
 
 import pytest
-from pydantic import BaseModel
-
 from agent_control._control_registry import (
     clear,
     get_registered_steps,
     merge_explicit_and_auto_steps,
     register,
 )
+from pydantic import BaseModel
 
 
 @pytest.fixture(autouse=True)
@@ -171,7 +171,7 @@ class TestRegister:
         self, caplog: pytest.LogCaptureFixture
     ) -> None:
         # Given a function registered before its forward-referenced model is available in globals.
-        def my_func(payload: "LaterModel") -> str:
+        def my_func(payload: LaterModel) -> str:
             ...
 
         register(my_func)
@@ -305,6 +305,34 @@ class TestDecoratorRegistration:
         assert steps[0]["name"] == "stacked"
         assert steps[0]["metadata"] == {"policy": "p1"}
 
+    def test_control_with_prior_wrapper_preserves_schema_fields(self) -> None:
+        # Given a user wrapper applied before @control() using functools.wraps.
+        from agent_control.control_decorators import control
+
+        def with_tracing(func: Any) -> Any:
+            @functools.wraps(func)
+            async def wrapped(*args: Any, **kwargs: Any) -> Any:
+                return await func(*args, **kwargs)
+
+            return wrapped
+
+        @control()
+        @with_tracing
+        async def wrapped_chat(message: str, include_context: bool = False) -> str:
+            return message
+
+        # When registered steps are materialized.
+        steps = get_registered_steps()
+
+        # Then registration keeps the original step identity and inferred schema fields.
+        assert len(steps) == 1
+        step = steps[0]
+        assert step["type"] == "llm"
+        assert step["name"] == "wrapped_chat"
+        assert step["input_schema"]["properties"]["message"]["type"] == "string"
+        assert step["input_schema"]["properties"]["include_context"]["type"] == "boolean"
+        assert step["output_schema"]["type"] == "string"
+
 
 class TestInitMerge:
     """Tests the explicit+auto merge behavior used by init()."""
@@ -345,7 +373,7 @@ class TestInitMerge:
         merge_result = merge_explicit_and_auto_steps(explicit_steps, auto_steps)
         merged = merge_result.steps
 
-        # Then the explicit version wins for the duplicate key while unrelated auto steps are retained.
+        # Then explicit wins for duplicates while unrelated auto steps are retained.
         my_step_entries = [s for s in merged if (s["type"], s["name"]) == ("llm", "my_step")]
         assert len(my_step_entries) == 1
         assert my_step_entries[0]["input_schema"] == {"custom": True}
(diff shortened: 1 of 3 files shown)


sdks/python/tests/test_control_registry.py
```diff
@@ -0,0 +1,396 @@
+"""Tests for the control step registry (_control_registry)."""
```
lan17: could you add Given/When/Then comments to the tests?

sdks/python/src/agent_control/__init__.py
```diff
@@ -435,6 +437,21 @@ async def handle(message: str):
     _server_url = server_url or os.getenv('AGENT_CONTROL_URL') or 'http://localhost:8000'
     _api_key = api_key
 
+    # Merge auto-discovered steps from @control() decorators with explicit steps.
+    # Explicit steps take precedence (by name) over auto-discovered ones.
```
lan17: do we even wanna take explicit steps?

sdks/python/src/agent_control/_control_registry.py
```diff
@@ -0,0 +1,191 @@
...
+        except Exception:
+            logger.debug("Failed to extract JSON schema from Pydantic model %s", annotation)
+            return None
+
+    # Primitive types
+    type_str = _PRIMITIVE_TYPE_MAP.get(annotation)  # type: ignore[arg-type]
+    if type_str is not None:
+        return {"type": type_str}
+
+    return None
```
lan17: Type coverage here is very narrow (`str/int/float/bool` + Pydantic models). Common annotations like `str | None`, `list[str]`, and `dict[str, int]` currently degrade to `None` and are dropped from `input_schema`/`output_schema`, so the auto-generated schema is incomplete for typical tool signatures.

sdks/python/src/agent_control/__init__.py
```diff
@@ -435,6 +437,21 @@ async def handle(message: str):
...
+    from agent_control._control_registry import get_registered_steps
+
+    auto_steps = get_registered_steps()
+    if auto_steps:
+        explicit_names = {s["name"] for s in (steps or [])}
```
lan17: Merge dedupe uses only `name`, but server identity is `(type, name)`. This can drop a valid auto-discovered step when explicit steps contain the same name under a different type. Consider deduping with `(type, name)` to match server behavior.

sdks/python/src/agent_control/_control_registry.py
```diff
@@ -0,0 +1,191 @@
...
+        hint = hints.get(name)
+        if hint is None:
+            continue
+        schema = _type_to_json_schema(hint)
+        if schema is not None:
+            properties[name] = schema
+
+    if not properties:
+        return None
+    return {"type": "object", "properties": properties}
```
lan17: This schema omits `required`, so parameters without defaults become optional in the generated JSON Schema. For example, `def f(query: str, limit: int = 5)` should include `required: ["query"]`.

PR #25, opened 2026-02-09
fix(sdk): Add step_name as parameter to control

namrataghadi-galileo:
Take step_name as parameter to control decorator.

### DIFF
--- a/server/src/agent_control_server/endpoints/agents.py
+++ b/server/src/agent_control_server/endpoints/agents.py
@@ -347,11 +347,11 @@ async def init_agent(
                 ],
             )
 
-    # Validate no duplicate step names within the same request
-    seen_step_keys: set[StepKeyTuple] = set()
+    # Build incoming_steps_by_key and validate no duplicates in single pass
+    incoming_steps_by_key: dict[StepKeyTuple, StepSchema] = {}
     for step in request.steps:
         step_key: StepKeyTuple = (step.type, step.name)
-        if step_key in seen_step_keys:
+        if step_key in incoming_steps_by_key:
             raise BadRequestError(
                 error_code=ErrorCode.VALIDATION_ERROR,
                 detail=f"Duplicate step name detected: type='{step.type}', name='{step.name}'. "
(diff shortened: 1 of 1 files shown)


sdks/python/src/agent_control/control_decorators.py
```diff
@@ -580,7 +613,7 @@ async def _execute_with_control(
         ctx.log_end()
 
 
-def control(policy: str | None = None) -> Callable[[F], F]:
+def control(policy: str | None = None, step_name: str | None = None) -> Callable[[F], F]:
```
lan17, line 616: Should we get rid of `policy` here?

sdks/python/src/agent_control/control_decorators.py
```diff
@@ -591,6 +624,10 @@ def control(policy: str | None = None) -> Callable[[F], F]:
         policy: Optional policy name for documentation. The agent's assigned
                 policy is automatically used. This parameter is for clarity
                 in code when multiple policies exist.
+        step_name: Optional custom name for this step. If not provided, uses
+                   the function name. Useful for grouping multiple functions
+                   under the same step name for control purposes, or when the
```
lan17: do we want to allow reusing step_names between different functions?  IMO we should probably not do that and fail fast if there's a name conflict.

The problem with reusing step_name is that things will get confusing if you have multiple functions with different signature reusing same step_name since we need to know their input/output schemas.

sdks/python/tests/test_control_decorators.py
```diff
@@ -461,3 +461,119 @@ def test_is_exception(self):
...
+# =============================================================================
+# STEP NAME TESTS
+# =============================================================================
+
+class TestStepName:
+    """Tests for custom step_name parameter."""
+
+    @pytest.mark.asyncio
+    async def test_custom_step_name_used_in_payload(self, mock_agent, mock_safe_response):
+        """Test that custom step_name is passed to evaluation payload."""
```
lan17, line 475: Could we please use Given/When/Then behavioral test style comments?

docs/REFERENCE.md
```diff
@@ -622,6 +622,30 @@ async def chat(message: str) -> str:
 
 **Works with both sync and async functions**.
 
+**Parameters**:
+
+- `policy` (str, optional): Policy name for documentation purposes. The agent's assigned policy is automatically used.
+- `step_name` (str, optional): Custom name for this step. If not provided, uses the function name. Useful for:
+  - Grouping multiple functions under the same step name for control purposes
```
lan17: mentioned elsewhere too, but not sure if we actually want to do this since the schemas would not align.

docs/REFERENCE.md
```diff
@@ -622,6 +622,30 @@ async def chat(message: str) -> str:
...
+  - Overriding auto-detected names when they don't match your control configuration
+  - Applying the same controls to functions with different names
```
lan17, line 630: This should be possible already via regex matching for stepnames

docs/REFERENCE.md
namrataghadi-galileo: Will take this out.

docs/REFERENCE.md
namrataghadi-galileo: Sure. And having custom step_names also achieves the same thing. So just calling it out here.

sdks/python/src/agent_control/control_decorators.py
namrataghadi-galileo: Its being worked on by @nachiket-galileo [here](https://app.shortcut.com/galileo/story/54639/policy-refactor)

sdks/python/src/agent_control/control_decorators.py
namrataghadi-galileo: But, there is nothing stopping someone from doing that even now!!!
I can take out the last statement. But, it should be called out explicitly in the DocString and Readme that `things will get confusing if there are multiple functions with different signature reusing same step_name since we need to know their input/output schemas`

sdks/python/tests/test_control_decorators.py
namrataghadi-galileo: done

sdks/python/src/agent_control/control_decorators.py
lan17: yea agree, we should discuss and figure out exact semantics we want here.

server/src/agent_control_server/endpoints/agents.py
```diff
@@ -347,6 +347,29 @@ async def init_agent(
                 ],
             )
 
+    # Validate no duplicate step names within the same request
+    seen_step_keys: set[StepKeyTuple] = set()
```
lan17: Perf ROI: while touching this section, consider skipping the second DB lookup by name in the common case where existing_by_uuid is already found below. You can validate name mismatch directly from existing_by_uuid and save one roundtrip per initAgent call.

server/src/agent_control_server/endpoints/agents.py
```diff
@@ -347,6 +347,29 @@ async def init_agent(
...
+    for step in request.steps:
```
lan17, line 352: Perf ROI: request.steps is also iterated again later to build incoming_steps_by_key. You could fold duplicate validation and map construction into a single pass here to reduce per-request Python work and allocations.

server/src/agent_control_server/endpoints/agents.py
```diff
@@ -492,8 +515,33 @@ async def init_agent(
         if key in incoming_steps_by_key:
             if key not in seen_steps:
                 incoming_step = incoming_steps_by_key[key]
-                if step.model_dump(mode="json") != incoming_step.model_dump(mode="json"):
-                    steps_changed = True
+                existing_schema = step.model_dump(mode="json")
```
lan17: Perf ROI: model_dump(mode="json") in this loop allocates temporary dicts on every iteration. Since both sides are already Pydantic models, comparing models directly (or only the schema-relevant fields) is cheaper. Similar optimization applies in evaluator comparison below.

server/src/agent_control_server/endpoints/agents.py
```diff
@@ -492,8 +515,33 @@ async def init_agent(
...
             if key not in seen_steps:
                 incoming_step = incoming_steps_by_key[key]
-                if step.model_dump(mode="json") != incoming_step.model_dump(mode="json"):
-                    steps_changed = True
+
+                # Compare only schema fields (type/name already matched by key)
+                # Avoid model_dump() allocations by direct field comparison
+                if (step.input_schema != incoming_step.input_schema or
+                    step.output_schema != incoming_step.output_schema):
+                    raise ConflictError(
```
lan17, line 523: talked to @soumya-io and it made sense to make this kind of validation optional on the server, so that user cal set a flag when calling initAgent which would make initAgent tolerate steps added/removed as well as schema changes.

filed this ticket to track this https://app.shortcut.com/galileo/story/55195/server-make-step-checks-in-initagent-optional

PR #29, opened 2026-02-13
fix(docs): (DO NOT MERGE) strands integration

namrataghadi-galileo:
**Strands Integration Example**
Demonstrates automatic safety controls for AWS Strands agents using a reusable hook pattern - no code decorators needed.

**What's Included**

**Core Integration**

- **AgentControlHook** - Reusable hook class that intercepts Strands agent events and enforces controls
- Automatic tool name extraction from Strands events for tool-specific controls
- Support for both pre and post-stage evaluation across LLM calls and tool executions

**Interactive Demo (interactive_demo/)**

- Customer support agent with Streamlit UI
- Real-time PII detection (SSN, email, credit card blocking)
- SQL injection prevention for database queries
- Shows control violations with user-friendly error messages

**Steering Demo (steering_demo/)**

- Banking email agent with PII redaction
- Combines AgentControl detection + Strands Steering correction
- Demonstrates layered governance: detect sensitive data, then intelligently redact it
- Shows integration between two safety systems

**Hook-Based Architecture**

- Attaches to Strands agents via the hooks parameter
- Intercepts BeforeToolCallEvent, AfterModelCallEvent, and other Strands lifecycle events
- Automatically constructs Step payloads and evaluates controls server-side
- No modification to agent code or tool definitions required

### DIFF
--- a/examples/strands_integration/common/agent_control_hook.py
+++ b/examples/strands_integration/common/agent_control_hook.py
@@ -125,6 +125,24 @@ class AgentControlHook(HookProvider):
     # Helper Methods
     # ============================================================================
 
+    def _invoke_callback(self, control_name: str, stage: str, result: EvaluationResult) -> None:
+        """Helper to invoke violation callback if configured."""
+        if self.on_violation_callback:
+            self.on_violation_callback(
+                {
+                    "agent": self.agent_name,
+                    "control_name": control_name,
+                    "stage": stage,
+                },
+                result,
+            )
+
+    def _raise_error(self, error: Exception, use_runtime_error: bool) -> None:
+        """Helper to raise error (RuntimeError wrapper if requested)."""
+        if use_runtime_error:
+            raise RuntimeError(str(error))
+        raise error
+
     async def _evaluate_and_enforce(
         self,
         step_name: str,
@@ -146,56 +164,44 @@ class AgentControlHook(HookProvider):
             agent_name=self.agent_name,
         )
 
+        # Check for explicit deny/steer actions
         action = _action_error(result)
         if action:
             _, err = action
             if isinstance(err, ControlSteerError):
                 print(f"\n🎯 STEER - {violation_type} needs correction")
                 print(f"   Reason: {err}")
-            if self.on_violation_callback:
-                self.on_violation_callback(
-                    {
-                        "agent": self.agent_name,
-                        "control_name": getattr(err, "control_name", "unknown"),
-                        "stage": stage,
-                    },
-                    result,
-                )
-            if use_runtime_error:
-                raise RuntimeError(str(err))
-            raise err
+
+            control_name = getattr(err, "control_name", "unknown")
+            self._invoke_callback(control_name, stage, result)
+            self._raise_error(err, use_runtime_error)
 
         # Fail closed if unsafe without explicit deny/steer match
         if not result.is_safe:
             control_name = "unknown"
             reason = result.reason
 
-            if result.matches and len(result.matches) > 0:
+            if result.matches:
                 first_match = result.matches[0]
                 control_name = first_match.control_name
 
                 if not reason:
-                    msg = getattr(getattr(first_match, "result", None), "message", None)
+                    # Extract message from first match result
+                    match_result = getattr(first_match, "result", None)
+                    msg = getattr(match_result, "message", None) if match_result else None
                     reason = msg or f"Control '{control_name}' triggered"
 
             print(f"\n🚫 CONTROL VIOLATION - {violation_type} blocked")
             print(f"   Control: {control_name}")
             print(f"   Reason: {reason}")
 
-            if self.on_violation_callback:
-                self.on_violation_callback(
-                    {
-                        "agent": self.agent_name,
-                        "control_name": control_name,
-                        "stage": stage,
-                    },
-                    result,
-                )
+            self._invoke_callback(control_name, stage, result)
 
             error_msg = f"Policy violation [{control_name}]: {reason}"
-            if use_runtime_error:
-                raise RuntimeError(error_msg)
-            raise agent_control.ControlViolationError(message=error_msg)
+            self._raise_error(
+                agent_control.ControlViolationError(message=error_msg),
+                use_runtime_error
+            )
     # ============================================================================
     # Hook Registration
     # ============================================================================
@@ -318,7 +324,7 @@ class AgentControlHook(HookProvider):
         print(f"{'='*70}")
 
         # Extract tool name and input from event
-        tool_name, tool_input = self._extract_tool_data(event, event_phase="pre")
+        tool_name, tool_input = self._extract_tool_data(event)
 
         print(f"🔧 Tool name: {tool_name}")
         # Format tool_input for display (might be dict or string)
@@ -360,7 +366,7 @@ class AgentControlHook(HookProvider):
             print(f"{'='*60}")
 
         # Extract tool name and output from event
-        tool_name, tool_output = self._extract_tool_data(event, event_phase="post")
+        tool_name, tool_output = self._extract_tool_data(event)
 
         if self.enable_logging:
             print(f"🔧 Tool name: {tool_name}")
@@ -378,7 +384,7 @@ class AgentControlHook(HookProvider):
     async def check_before_node(self, event: BeforeNodeCallEvent):
         """Check controls before node call (multi-agent graphs)."""
         input_text, _ = self._extract_messages(event)
-        node_id = event.node_id if hasattr(event, "node_id") else "unknown"
+        node_id = getattr(event, "node_id", "unknown")
 
         await self._evaluate_and_enforce(
             step_name=node_id,
@@ -391,7 +397,7 @@ class AgentControlHook(HookProvider):
     async def check_after_node(self, event: AfterNodeCallEvent):
         """Check controls after node call (multi-agent graphs)."""
         _, output_text = self._extract_messages(event)
-        node_id = event.node_id if hasattr(event, "node_id") else "unknown"
+        node_id = getattr(event, "node_id", "unknown")
 
         await self._evaluate_and_enforce(
             step_name=node_id,
@@ -405,13 +411,27 @@ class AgentControlHook(HookProvider):
     # Message Extraction Utilities
     # ============================================================================
 
+    def _extract_user_message_from_list(self, messages: list | None, reverse: bool = False) -> str:
+        """Extract user message content from Strands message list (TypedDict with role and content)."""
+        if not messages:
+            return ""
+
+        msg_iter = reversed(messages) if reverse else messages
+        for msg in msg_iter:
+            if isinstance(msg, dict) and msg.get("role") == "user":
+                return self._extract_content_text(msg.get("content", ""))
+        return ""
+
     def _extract_messages(self, event: Any) -> tuple[str, str]:
         """
         Extract input and output messages from event.
 
-        Handles different event types:
-        - BeforeInvocationEvent: has event.messages (user input)
-        - AfterModelCallEvent: has event.stop_response.message (model output)
+        Handles all Strands event types based on actual structure:
+        - BeforeInvocationEvent: messages (Optional[list[Message]])
+        - BeforeModelCallEvent: invocation_state (dict, has default_factory)
+        - AfterModelCallEvent: stop_response.message (Optional[ModelStopResponse])
+        - BeforeNodeCallEvent: invocation_state (Optional[dict], defaults to None)
+        - AfterNodeCallEvent: invocation_state (Optional[dict], defaults to None)
 
         Returns:
             Tuple of (input_text, output_text)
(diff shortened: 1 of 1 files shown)


examples/strands_integration/interactive_demo/interactive_support_demo.py
```diff
@@ -0,0 +1,965 @@
...
+
+# Load environment variables
+load_dotenv(Path(__file__).parent.parent / ".env", override=False)
+
+
+# ============================================================================
+# Safety Tracking Hook (Captures violations for UI display)
+# ============================================================================
+
+class SafetyTrackingHook(HookProvider):
```
yashsheth46: class AgentControlHook(HookProvider):
"""A hook that subscribes to all key events which map to a pre-defined step_name."""


    def __init__(self, agent_name, max_retries, even_control_list):
        self.event_control_list=even_control_list


    def register_hooks(self, registry: HookRegistry, **kwargs) -> None:
     """if even_control_list is empty, this registers callbacks for all key events that strands supports"""
        default_event_dict = {
            AgentInitializedEvent : self.agent_init,
            BeforeInvocationEvent : self.before_invocation,
            AfterInvocationEvent : self.after_invocation,
            BeforeModelCallEvent : self.before_model_call,
            AfterModelCallEvent : self.after_model_call,
            BeforeToolCallEvent : self.before_tool_call,
            AfterToolCallEvent : self.after_tool_call,
            BeforeNodeCallEvent : self.before_node_call,
            AfterNodeCallEvent : self.after_node_call
        }

        if not self.event_control_list:
            event_registration_list = self.event_control_list
        else:
            event_registration_list = default_event_dict.keys()

        for event in event_registration_list:
           if (event not in default_event_dict.keys())
               return Exception("Unknown event being registered")
            registry.add_callback(event, default_event_dict[event])

    def extract_event_messages(self, event) -> ControlParams
     """Extracts the event messages into a ControlParams json that abstracts any event related stuff into a standardized structure that goes into our evaluators"""
        # Check if we have messages
        if not event.messages:
            return "";
        # Extract messages here
        for msg in event.messages:
        

    @control(step_name="agent_init")
    async def agent_init(self, event: AgentInitializedEvent) -> ControlParams
        return extract_event_messages(event)


    @control(step_name="before_agent_invocation")
    async def before_invocation(self, event: BeforeInvocationEvent) -> ControlParams
        return extract_event_messages(event)

# Add more callbacks here

namrataghadi-galileo: For achieving something like this, we will need handling for hooks in control_decorator.py

```
class AgentControlHook:
    def __init__(self, agent_uuid, agent_name, server_url, event_control_list=None):
        self.agent_uuid = agent_uuid
        self.agent_name = agent_name
        self.server_url = server_url
        self.event_control_list = event_control_list or []
    
    def _base_request(self, event):
        """Build base request dict with common fields."""
        trace_id, span_id = get_trace_and_span_ids()
        return {
            "agent_uuid": str(self.agent_uuid),
            "server_url": self.server_url,
            "mode": "hook",
            "event": event
        }
    
    @agent_control.control()
    async def check_before_tool(self, event):
        return {
            **self._base_request(event),
            "step": {
                "type": "tool",
                "name": event.tool_call.name,
                "input": event.tool_call.input
            },
            "stage": "pre"
        }
    
    @agent_control.control()
    async def check_after_tool(self, event):
        return {
            **self._base_request(event),
            "step": {
                "type": "tool",
                "name": event.tool_call.name,
                "input": event.tool_call.input,
                "output": event.tool_result.output
            },
            "stage": "post"
        }

  @agent_control.control()
 async def check_before_model(self, event: BeforeModelCallEvent):
    # NO - Model events don't have a specific "function name"
    # This event fires when the LLM is about to be called, not a specific function
    # Use a static name to represent "LLM call"
    
    return {
        "mode": "hook",
        "agent_uuid": str(self.agent_uuid),
        "step": {
            "type": "llm",
            "name": "llm_call",  # Static name - always the same
            "input": {
                "messages": event.messages  # The actual LLM input
            }
        },
        "stage": "pre",
        "event": event,
    }
.......
Similarly other calls backs.

# In control_decorator.py

async def _execute_with_control(
    func: Callable,
    args: tuple,
    kwargs: dict,
    is_async: bool,
    step_name: str | None = None,
    controls: list[dict[str, Any]] | None = None,
) -> Any:
    # Get agent and create context
    agent = _get_agent()
    if not agent:
        # No agent initialized, execute without controls
        return await func(*args, **kwargs) if is_async else func(*args, **kwargs)
    
    # Create ControlContext (will work for normal functions, not hooks - that's OK)
    trace_id, span_id = get_trace_and_span_ids()
    ctx = ControlContext(
        agent_uuid=str(agent.agent_id),
        agent_name=agent.agent_name,
        server_url=_get_server_url(),
        func=func,
        args=args,
        kwargs=kwargs,
        trace_id=trace_id,
        span_id=span_id,
        start_time=time.perf_counter(),
        step_name=step_name,
    )
    ctx.log_start()
    
    try:
        # PRE-EXECUTION: Try to run (will fail silently for hooks)
        try:
            result = await _evaluate(
                ctx.agent_uuid, ctx.pre_payload(), "pre",
                ctx.server_url, ctx.trace_id, ctx.span_id,
                controls=controls,
            )
            ctx.process_result(result, "pre")
        except ControlViolationError:
            raise  # Always raise violations
        except Exception as e:
            # PRE check failed (expected for hooks) - log and continue
            logger.debug(f"Pre-execution check skipped: {e}")
        
        # Execute the function
        if is_async:
            output = await func(*args, **kwargs)
        else:
            output = func(*args, **kwargs)
        
        # NEW: Detect hook callback pattern
        if isinstance(output, dict) and output.get("mode") == "hook":
            # Hook callback - evaluate extracted Step at specified stage
            try:
                result = await _evaluate(
                    output["agent_uuid"],
                    output["step"],
                    output["stage"],  # "pre" or "post" from callback
                    output.get("server_url", ctx.server_url),
                    trace_id=output.get("trace_id", ctx.trace_id),
                    span_id=output.get("span_id", ctx.span_id),
                    controls=controls,
                )
                
                # If not safe, cancel the event
                if not result.get("is_safe", True):
                    event = output["event"]
                    violation_msg = result.get("reason", "Safety violation detected")
                    
                    # Cancel event based on type
                    if hasattr(event, "cancel_tool"):
                        event.cancel_tool = violation_msg
                    elif hasattr(event, "retry"):
                        event.retry = True
                        if hasattr(event, "retry_guidance"):
                            event.retry_guidance = violation_msg
                    
            except Exception as e:
                logger.error(f"Hook callback evaluation failed: {e}")
            
            return None  # Hook callbacks always return None to Strands
        
        # EXISTING: POST-EXECUTION for normal functions
        try:
            result = await _evaluate(
                ctx.agent_uuid, ctx.post_payload(output), "post",
                ctx.server_url, ctx.trace_id, ctx.span_id,
                controls=controls,
            )
            ctx.process_result(result, "post")
        except ControlViolationError:
            raise
        except Exception as e:
            logger.error(f"Post-execution control check failed: {e}")
        
        return output
    finally:
        ctx.log_end()
```

namrataghadi-galileo: cc @lan17 Any thoughts on this?

lan17: I think strands integration should not touch control decorator at all. 

We should just be calling `evaluate` directly to server/evaluation engine directly.

sdks/python/src/agent_control/evaluation.py
```diff
@@ -494,3 +494,126 @@ def _with_parse_errors(result: EvaluationResult) -> EvaluationResult:
...
+            "chat",
+            input="User message",
+            stage="pre",
+            agent_uuid="550e8400-e29b-41d4-a716-446655440000",
+            agent_name="customer-service-bot",
+            controls=my_controls
+        )
+    """
+    # Import here to avoid circular dependency
+    from . import _api_key, _current_agent, _server_controls, _server_url
```
lan17: no local imports please.

sdks/python/src/agent_control/evaluation.py
```diff
@@ -494,3 +494,126 @@ def _with_parse_errors(result: EvaluationResult) -> EvaluationResult:
...
+
+async def evaluate_controls(
+    step_name: str,
+    *,
+    input: Any = None,
+    output: Any = None,
+    context: dict[str, Any] | None = None,
+    step_type: Literal["tool", "llm"] = "llm",
+    stage: Literal["pre", "post"] = "pre",
+    agent_uuid: str | UUID | None = None,
```
lan17: don't allow  none here since we raise exception

sdks/python/src/agent_control/evaluation.py
```diff
@@ -494,3 +494,126 @@ def _with_parse_errors(result: EvaluationResult) -> EvaluationResult:
...
+    agent_name: str | None = None,
```
lan17: same here, lets not allwo none

sdks/python/src/agent_control/evaluation.py
```diff
@@ -494,3 +494,126 @@ def _with_parse_errors(result: EvaluationResult) -> EvaluationResult:
...
+    controls: list[dict[str, Any]] | None = None,
```
lan17: We should never have to pass this in here since we can always get the controls defined by the sever right?

examples/strands_integration/interactive_demo/interactive_support_demo.py
namrataghadi-galileo: This is done.

sdks/python/src/agent_control/evaluation.py
namrataghadi-galileo:
```
The Circular Dependency:

__init__.py imports from evaluation.py (lines 73, 88):


from . import evaluation
from .evaluation import check_evaluation_with_local, evaluate_controls
evaluation.py needs access to global variables defined in __init__.py (lines 228-233):


_current_agent, _server_controls, _server_url, _api_key
If evaluation.py imported these at the module level (top of file), Python would fail:

__init__.py starts importing → reaches from . import evaluation
evaluation.py starts loading → tries to from . import _api_key, ...
But __init__.py hasn't finished loading yet → ImportError: cannot import name '_api_key'
The Solution:
The local import inside the function (line 563) delays the import until after both modules are fully loaded. This is a standard Python pattern for breaking circular dependencies.
```

lan17: Let's just fix the circular dependency.

sdks/python/src/agent_control/evaluation.py
namrataghadi-galileo: done

sdks/python/src/agent_control/evaluation.py
namrataghadi-galileo: done

sdks/python/src/agent_control/evaluation.py
namrataghadi-galileo: done

sdks/python/src/agent_control/evaluation.py
namrataghadi-galileo: Its a medium amount of change. But, this is done.

sdks/python/src/agent_control/evaluation.py
```diff
@@ -494,3 +495,110 @@ def _with_parse_errors(result: EvaluationResult) -> EvaluationResult:
 
     # No controls at all - still include parse_errors if any
     return _with_parse_errors(EvaluationResult(is_safe=True, confidence=1.0))
+
+
+async def evaluate_controls(
+    step_name: str,
+    *,
+    input: Any = None,
```
lan17: this should be `Any | None`?  not sure if `Any` covers `None`

sdks/python/src/agent_control/evaluation.py
```diff
@@ -494,3 +495,110 @@ def _with_parse_errors(result: EvaluationResult) -> EvaluationResult:
...
+    output: Any = None,
+    context: dict[str, Any] | None = None,
+    step_type: Literal["tool", "llm"] = "llm",
+    stage: Literal["pre", "post"] = "pre",
+    agent_uuid: str | UUID,
```
lan17: wonder if we should get rid of it since im getting rid of it here: https://github.com/agentcontrol/agent-control/pull/44

sdks/python/src/agent_control/evaluation.py
lan17: Also, in what situations would this be null? 

Same question for the output.  

I suppose it would only be the case if there's no args to a tool call and a void output?

sdks/python/src/agent_control/evaluation.py
namrataghadi-galileo: We will have to uptake it once your PR is merged.

examples/strands_integration/common/agent_control_hook.py
```diff
@@ -0,0 +1,556 @@
...
+        - AfterModelCallEvent: has event.stop_response.message (model output)
+
+        Returns:
+            Tuple of (input_text, output_text)
+        """
+        input_text = ""
+        output_text = ""
+
+        # BeforeInvocationEvent - has messages attribute with user input
+        if hasattr(event, "messages") and event.messages:
```
zastrowm: Explicit event type checks would be preferable IMHO - for type safety purposes and readability

Same for below for tools

examples/strands_integration/common/agent_control_hook.py
```diff
@@ -0,0 +1,556 @@
...
+
+        tool_name = "unknown-tool"
+        tool_data = ""
+
+        # Extract tool name from selected_tool or tool_use
+        if hasattr(event, "selected_tool") and event.selected_tool:
+            # selected_tool is an AgentTool object with name attribute
+            if hasattr(event.selected_tool, "name"):
+                tool_name = event.selected_tool.name
+            elif hasattr(event.selected_tool, "__name__"):
```
zastrowm: Did you see a case where you got something that wasn't AgentTool? Same question for below and "get"

examples/strands_integration/common/agent_control_hook.py
```diff
@@ -0,0 +1,556 @@
...
+                tool_name = event.selected_tool.__name__
+        elif hasattr(event, "tool_use") and event.tool_use:
+            # tool_use is a dict with toolName key
+            if isinstance(event.tool_use, dict):
+                tool_name = event.tool_use.get("toolName", "unknown-tool")
+            elif hasattr(event.tool_use, "get"):
+                tool_name = event.tool_use.get("toolName", "unknown-tool")
```
zastrowm: Similar question to the need for "get" here

examples/strands_integration/common/agent_control_hook.py
```diff
@@ -0,0 +1,556 @@
...
+
+        if self.enable_logging:
+            print(f"🔍 Extracted tool name: {tool_name}")
+
+        # Extract input (pre-stage) or output (post-stage)
+        if event_phase == "pre":
```
zastrowm: If you did an isinstance check for BeforeToolCall and AfterToolCall, I think you could even omit the `event_phase` here?

examples/strands_integration/common/agent_control_hook.py
```diff
@@ -0,0 +1,556 @@
...
+        # Dict with 'content' key
+        if isinstance(content, dict):
+            content = content.get("content", content)
+
+        # Object with 'content' attribute
+        if hasattr(content, "content"):
+            content = content.content
+
+        # List of content blocks
+        if isinstance(content, list):
```
zastrowm, line 580: Do you ever want anything other than plain text?

[Our AgentResult](https://github.com/strands-agents/sdk-python/blob/694c4a7e135f8258f1138bf94efcc436d3db4896/src/strands/agent/agent_result.py#L38) also stringifies structured output and citation blocks

examples/strands_integration/common/agent_control_hook.py
namrataghadi-galileo: sure.

examples/strands_integration/common/agent_control_hook.py
namrataghadi-galileo: agreed. The original code had overly defensive `hasattr()` checks that don't correspond to any real case in Strands. Refactored code correctly removes these unnecessary checks and uses the actual types

examples/strands_integration/common/agent_control_hook.py
namrataghadi-galileo: Switching to an isinstance check let us remove a lot of code — great catch. This is handled there now.

examples/strands_integration/common/agent_control_hook.py
namrataghadi-galileo: Hooks intercept at a lower level (model/tool calls) BEFORE AgentResult is constructed. What we receive from the events is:
```
Message.content: list[ContentBlock]
ToolResult.content: list[ToolResultContent]
invocation_state values: Any (strings, dicts, etc.)
```

So we need more than plain text—thanks for flagging it. We now extract text, citations, tool info, and JSON from ContentBlock/ToolResultContent TypedDicts for control inspection, since those fields can contain content that needs policy enforcement.

examples/strands_integration/common/agent_control_hook.py
namrataghadi-galileo: type check refactor takes care of this redundancy

PR #34, opened 2026-02-19
fix(ui): selector subpaths issue

siddhant-galileo:


ui/src/core/page-components/agent-detail/modals/edit-control/control-definition-form.tsx
```diff
@@ -51,28 +58,32 @@ export const ControlDefinitionForm = ({ form }: ControlDefinitionFormProps) => {
...
-          { value: 'name', label: 'name' },
-          { value: 'type', label: 'type' },
+          '*',
+          'input',
+          'output',
+          'context',
+          'name',
+          'type',
+          'input.args',
+          'input.args.command',
```
lan17: this is not the right fix.

The input should allow arbitrary subpaths for input

siddhant-galileo: i mean you can add arbitrary paths, these are just suggestions. (you can use input, output and add arbitrary suffix paths)
backend enforces atleast the first key - `valid_roots = {"input", "output", "name", "type", "context", "*"}`

siddhant-galileo: https://github.com/user-attachments/assets/153d2dce-061d-4980-bfc0-398012bbdefb

video for reference.

PR #35, opened 2026-02-19
fix(sdk): (DO NO MERGE) add stage_override to control decorator inputs

namrataghadi-galileo:
Add stage_override parameter to the control decorator inputs. This allows users to override the original setting in the database thereby increasing the reusability of the control

### DIFF
--- a/sdks/python/src/agent_control/control_decorators.py
+++ b/sdks/python/src/agent_control/control_decorators.py
@@ -510,6 +510,7 @@ async def _execute_with_control(
     kwargs: dict,
     is_async: bool,
     step_name: str | None = None,
+    stage_override: str | None = None,
 ) -> Any:
     """
     Core control execution logic for both async and sync functions.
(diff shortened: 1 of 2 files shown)


sdks/python/src/agent_control/control_decorators.py
```diff
@@ -571,49 +577,60 @@ async def _execute_with_control(
...
+                result = await _evaluate(
+                    ctx.agent_uuid, ctx.post_payload(output), "post",
+                    ctx.server_url, ctx.trace_id, ctx.span_id,
+                    controls=controls,
+                )
+                ctx.process_result(result, "post")
+            except ControlViolationError:
+                raise
+            except Exception as e:
+                logger.error(f"Post-execution control check failed: {e}")
```
lan17, line 622: exc_info=True

sdks/python/src/agent_control/control_decorators.py
```diff
@@ -571,49 +577,60 @@ async def _execute_with_control(
...
         return output
     finally:
         ctx.log_end()
 
 
-def control(policy: str | None = None, step_name: str | None = None) -> Callable[[F], F]:
+def control(
+    policy: str | None = None,
+    step_name: str | None = None,
+    stage_override: str | None = None,
```
lan17, line 632: I don't understand the use case of this.

The whole point of agent control is to have controls defined at runtime in a server, so why would someone want to override this in code?  Wouldn't that defeat the purpose of runtime control?

PR #38, opened 2026-02-22
fix(sdk): 54253 add steer action and example

namrataghadi-galileo:
**Add "Steer" Action Type for Corrective Guidance**
Introduces steer as a 5th action type (alongside allow, deny, warn, log) to enable corrective guidance without hard-blocking execution.

**What is Steer?**
A middle ground between hard blocks (deny) and passive logging (warn/log). When a control with action="steer" matches:

Raises ControlSteerError (instead of ControlViolationError)
Includes guidance field with correction instructions
Allows applications to modify their approach and retry
Priority: deny > steer > warn/log

**Changes
Models (agent-control-models)**

Added "steer" to ControlAction.decision literal type
Added optional guidance: str | None field to ControlAction
Added steer_count: int to ControlStats for observability

**SDK (agent-control-sdk)**

New ControlSteerError exception with guidance, control_id, control_name, metadata
Updated @control() decorator to handle steer actions
Guidance fallback chain: control.action.guidance → metadata.guidance → evaluator message

**Engine (agent-control-engine)**
Treat steer same as deny for safety semantics (is_safe=False, early cancellation, fail-closed)

**Server**
Added steer_count to SQL aggregation queries in observability store
Fixed action_counts initialization to include "steer": 0

**UI**

Added "Steer" option to action dropdown
Conditional "Guidance" text field (optional, shown only when steer selected)
Regenerated TypeScript types from OpenAPI schema

**Documentation**

Updated OVERVIEW.md, REFERENCE.md, SDK README.md, CONCEPTS.md with steer semantics
Added ControlSteerError usage examples

**Tests**

Updated test fixtures to include steer_count
Added "steer" to action validation tests
All 354 server tests passing ✅

**Example Usage**

from agent_control import control, ControlSteerError

@control()
async def process_input(user_text: str):
    return user_text

try:
    result = await process_input("bad language here")
except ControlSteerError as e:
    print(f"Guidance: {e.guidance}")
    # Apply corrections and retry
    corrected = apply_guidance(user_text, e.guidance)
    result = await process_input(corrected)

**Breaking Changes: None - fully backward compatible**

### DIFF
--- a/engine/tests/test_core.py
+++ b/engine/tests/test_core.py
@@ -15,11 +15,15 @@ from agent_control_engine import clear_evaluator_cache
 from agent_control_engine.core import ControlEngine, _compile_regex
 from agent_control_evaluators import Evaluator, EvaluatorMetadata, register_evaluator
 from agent_control_models import (
+    ControlAction,
     ControlDefinition,
+    ControlScope,
+    ControlSelector,
     EvaluationRequest,
     EvaluatorResult,
     EvaluatorSpec,
     Step,
+    SteeringContext,
 )
 from pydantic import BaseModel
 
@@ -179,6 +183,7 @@ def make_control(
     step_names: list[str] | None = None,
     step_name_regex: str | None = None,
     execution: str = "server",
+    steering_context: SteeringContext | None = None,
 ) -> MockControlWithIdentity:
     """Create a mock control for testing."""
     selector: dict[str, Any] = {}
@@ -210,7 +215,11 @@ def make_control(
                 name=evaluator,
                 config={"value": config_value},
             ),
-            action={"decision": action},
+            action=(
+                ControlAction(decision=action, steering_context=steering_context)
+                if steering_context
+                else ControlAction(decision=action)
+            ),
         ),
     )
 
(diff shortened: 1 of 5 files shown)


engine/src/agent_control_engine/core.py
```diff
@@ -196,10 +196,10 @@ async def evaluate_control(eval_task: _EvalTask) -> None:
                         timeout=timeout,
                     )
 
-                    # Signal if this is a deny match
+                    # Signal if this is a deny or steer match
```
lan17: This sets the cancellation signal for both `deny` and `steer`, so a fast steer match can cancel slower deny controls before they finish. That can surface only `steer` matches even when a `deny` would also match, which breaks deny-first semantics. Consider canceling early only for `deny`, then resolving `steer` only after confirming no deny matched.

sdks/python/src/agent_control/control_decorators.py
```diff
@@ -463,6 +508,26 @@ def _handle_evaluation_result(result: dict[str, Any]) -> None:
                     message=message,
                     metadata=metadata
                 )
+            elif action == "steer":
```
lan17: This loop raises `ControlSteerError` as soon as it encounters a steer match. If a deny match is also present later in `matches`, deny is never enforced. To preserve deny precedence, pre-scan matches for any `deny` and raise `ControlViolationError` first, then handle steer/warn/log ordering.

models/src/agent_control_models/controls.py
```diff
@@ -211,9 +211,17 @@ def validate_evaluator_config(self) -> Self:
 class ControlAction(BaseModel):
     """What to do when control matches."""
 
-    decision: Literal["allow", "deny", "warn", "log"] = Field(
+    decision: Literal["allow", "deny", "steer", "warn", "log"] = Field(
         ..., description="Action to take when control is triggered"
     )
+    steering_context: str | None = Field(
```
lan17: should this be an object so that we can add more fields to it later?

i.e. 
```
class SteeringContext(BaseModel):
    message: str
   ... 
    new fields in future will go here
```

namrataghadi-galileo: keeping it simple for now. Lets wait for feedback and we can make changes later.

engine/src/agent_control_engine/core.py
namrataghadi-galileo: yes good point

models/src/agent_control_models/controls.py
lan17: I think we should still introduce this here as otherwise we would need a backward incompat change later if we want to add more fields, i.e. we would need to change the type of steering_context

sdks/python/src/agent_control/control_decorators.py
namrataghadi-galileo: done

PR #40, opened 2026-02-23
feat(server): add initAgent conflict_mode overwrite mode with SDK defaults

lan17:
## Summary
- add server-side `conflict_mode` on `initAgent` (`strict` default, `overwrite` opt-in)
- in overwrite mode, replace existing steps/evaluators with latest init payload and return overwrite metadata (`overwrite_applied`, `overwrite_changes`, `warnings`)
- keep strict mode semantics unchanged
- warn (do not block) when overwrite removes evaluators referenced by active controls
- update Python SDK init/register flows to send `conflict_mode`, defaulting SDK init flows to `overwrite`
- pin Python `refresh_controls_async()` to `conflict_mode="strict"` to avoid accidental schema removal on refresh
- regenerate TypeScript SDK models for new request/response schema and default `conflictMode` to `overwrite` in generated outbound schema
- add Python and TypeScript tests for conflict-mode defaults/wiring

## Story
- https://app.shortcut.com/galileo/story/55195/server-make-step-checks-in-initagent-optional

## Validation
- `make server-test`
- `make openapi-spec-check`
- `uv run --package agent-control-server ruff check --config pyproject.toml models/src/agent_control_models/agents.py server/src/agent_control_server/endpoints/agents.py server/tests/test_init_agent_conflict_mode.py`
- `uv run --package agent-control-server mypy --config-file pyproject.toml models/src/agent_control_models/agents.py server/src/agent_control_server/endpoints/agents.py`
- `make sdk-lint`
- `make sdk-typecheck`
- `make sdk-test`
- `make sdk-ts-typecheck`
- `make sdk-ts-test`
- `make sdk-ts-lint`
- `make sdk-ts-generate-check`

### DIFF
--- a/sdks/python/tests/test_init_step_merge.py
+++ b/sdks/python/tests/test_init_step_merge.py
@@ -167,6 +167,7 @@ def test_init_logs_fallback_warning_for_unresolved_type_hints(
 
 @pytest.mark.asyncio
 async def test_refresh_controls_uses_strict_conflict_mode() -> None:
+    # Given: an initialized SDK agent session with network-facing calls mocked.
     register_agent_mock = AsyncMock(return_value={"created": True, "controls": []})
     health_check_mock = AsyncMock(return_value={"status": "healthy"})
 
@@ -182,9 +183,11 @@ async def test_refresh_controls_uses_strict_conflict_mode() -> None:
             agent_id=str(uuid4()),
         )
 
+        # When: controls are refreshed through refresh_controls_async().
         register_agent_mock.reset_mock()
         await agent_control.refresh_controls_async()
 
+    # Then: refresh registration is non-destructive and forces strict conflict handling.
     assert register_agent_mock.await_count == 1
     assert register_agent_mock.await_args is not None
     assert register_agent_mock.await_args.kwargs["steps"] == []
--- a/server/tests/test_init_agent_conflict_mode.py
+++ b/server/tests/test_init_agent_conflict_mode.py
@@ -70,6 +70,7 @@ def _create_policy_with_agent_evaluator_control(
 
 
 def test_init_agent_overwrite_replaces_steps_and_evaluators(client: TestClient) -> None:
+    # Given: an existing agent registration with baseline steps and evaluators.
     agent_id = str(uuid.uuid4())
     agent_name = f"Agent-{uuid.uuid4().hex[:8]}"
 
@@ -100,6 +101,7 @@ def test_init_agent_overwrite_replaces_steps_and_evaluators(client: TestClient) 
     assert create_resp.status_code == 200
     assert create_resp.json()["created"] is True
 
+    # When: initAgent is called in overwrite mode with an updated registration payload.
     overwrite_payload = _init_payload(
         agent_id=agent_id,
         agent_name=agent_name,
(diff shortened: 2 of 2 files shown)


models/src/agent_control_models/server.py
```diff
@@ -23,6 +24,67 @@ class EvaluatorSchema(BaseModel):
     description: str | None = Field(None, max_length=1000, description="Optional description")
 
 
+class ConflictMode(StrEnum):
+    """Conflict handling mode for agent registration updates."""
+
+    STRICT = "strict"
```
namrataghadi-galileo, line 34: Good to have one line comment to state what `strict` means.

sdks/python/src/agent_control/__init__.py
```diff
@@ -281,7 +281,8 @@ async def refresh_controls_async() -> list[dict[str, Any]] | None:
         response = await agents.register_agent(
             client,
             _current_agent,
-            steps=[]
+            steps=[],
+            conflict_mode="strict",
```
namrataghadi-galileo, line 287: When should the mode be `strict` vs `overwrite`?

server/src/agent_control_server/endpoints/agents.py
```diff
@@ -152,6 +155,74 @@ async def _validate_policy_controls_for_agent(
...
+        parsed = parse_evaluator_ref_full(evaluator_ref)
+        if parsed.type != "agent":
+            continue
+        if parsed.namespace != agent.name:
+            continue
+        if parsed.local_name not in removed_evaluators:
+            continue
+        references_by_evaluator.setdefault(parsed.local_name, []).append((control.id, control.name))
+
+    removal_by_name: dict[str, InitAgentEvaluatorRemoval] = {}
```
namrataghadi-galileo: nitpick to simplify this:
```
result = []
for evaluator_name in sorted_removed:
    if evaluator_name in references_by_evaluator:
        references = references_by_evaluator[evaluator_name]
        result.append(InitAgentEvaluatorRemoval(
            name=evaluator_name,
            referenced_by_active_controls=True,
            control_ids=[ref.id for ref in references],
            control_names=[ref.name for ref in references],
        ))
    else:
        result.append(InitAgentEvaluatorRemoval(name=evaluator_name))
return result
```

sdks/python/src/agent_control/__init__.py
lan17: `strict` should be used for non-destructive calls where we do not want registration mutation (for example `refresh_controls_async`, which intentionally sends `steps=[]` just to refresh controls). `overwrite` is for full registration updates where latest init payload should replace previous steps/evaluators. Added an inline comment at this call site to make that explicit.

models/src/agent_control_models/server.py
lan17: Added this in the shared model docs: `strict` preserves existing compatibility checks/conflict errors, and `overwrite` applies latest-init-wins replacement for steps/evaluators.

server/src/agent_control_server/endpoints/agents.py
lan17: Applied this simplification in `_build_overwrite_evaluator_removals`: now we iterate once over `sorted(removed_evaluators)` and append either the referenced or default removal entry directly.

PR #43, opened 2026-02-25
feat(sdk): Refresh controls in a background loop

lan17:
## Summary
- add SDK wrappers for agent controls refresh (`agents.list_agent_controls` and typed `agents.list_agent_controls_typed`)
- change background refresh to call `GET /agents/{id}/controls` only (never `init_agent`)
- enforce swap-only cache publication and robust refresh thread lifecycle (start/stop/re-init)
- expand behavioral tests (GIVEN/WHEN/THEN) for refresh loop lifecycle, sync/async refresh paths, wrapper behavior, and typed validation
- harden test DB reset/setup for server and SDK test flows to prevent stale table/alembic-state failures during `make check`
- remove defensive `try/except ImportError` fallbacks for core SDK model/engine imports and use strict imports

## Why this is needed
The SDK process is often long-lived (workers, agent runtimes, service containers). Controls can be updated on the server after the process has started, but without a refresh path the SDK keeps using stale control state until restart. That creates a policy enforcement gap: newly added controls are not enforced promptly, and removed/modified controls are not reflected in runtime behavior.

The background refresh loop closes that gap by pulling current controls from the server and updating the local cache in-process. This gives us near-real-time policy propagation without requiring deploys or process restarts, which is important for both security response and day-to-day operations.

Using `GET /agents/{id}/controls` specifically is the right contract for refresh because it is a read-only controls fetch. `init_agent` is a registration/initialization API with broader semantics and side effects; refresh should not depend on initialization behavior. Keeping refresh on a dedicated controls endpoint makes behavior explicit, safer, and easier to reason about and test.

## Additional rationale: removing defensive import fallbacks
Historically, parts of the SDK had defensive `try/except ImportError` logic with runtime fallback classes for `agent_control_models`/engine types. In current supported packaging paths, these dependencies are expected to be present (workspace development setup and published SDK build flow). Keeping fallback paths adds maintenance cost and creates split behavior branches that are hard to reason about, type-check, and test.

This PR removes those fallback branches in core SDK paths and adopts fail-fast import behavior. That tightens API contracts, simplifies code paths, improves mypy signal quality, and prevents silent degradation into weakly-typed runtime behavior when packaging is misconfigured.

## Testing
- `make check`

## Ticket
- Shortcut story: https://app.shortcut.com/galileo/story/56313/sdk-background-policy-refresh-loop-for-control-cache
- Follow-up migration story: https://app.shortcut.com/galileo/story/56439/sdk-typed-response-model-migration-plan

### DIFF
--- a/sdks/python/tests/test_evaluation.py
+++ b/sdks/python/tests/test_evaluation.py
@@ -29,10 +29,8 @@ async def test_check_evaluation_requires_step_name_before_server_call():
 
 
 @pytest.mark.asyncio
-async def test_check_evaluation_returns_runtime_fallback_without_models(monkeypatch):
-    """Fallback path should return an object with expected attributes."""
-    monkeypatch.setattr(evaluation, "MODELS_AVAILABLE", False)
-
+async def test_check_evaluation_returns_result_model():
+    """check_evaluation returns a parsed EvaluationResult."""
     class DummyResponse:
         def raise_for_status(self) -> None:
             return None
@@ -58,7 +56,13 @@ async def test_check_evaluation_returns_runtime_fallback_without_models(monkeypa
         "/api/v1/evaluation",
         json={
             "agent_name": "agent-example_01",
-            "step": {"type": "llm", "name": "chat", "input": "hello"},
+            "step": {
+                "type": "llm",
+                "name": "chat",
+                "input": "hello",
+                "output": None,
+                "context": None,
+            },
             "stage": "pre",
         },
     )
--- a/sdks/python/tests/test_init_validation.py
+++ b/sdks/python/tests/test_init_validation.py
@@ -15,8 +15,7 @@ def test_init_rejects_invalid_agent_name() -> None:
 def test_init_rejects_negative_policy_refresh_interval() -> None:
     with pytest.raises(ValueError, match="policy_refresh_interval_seconds must be >= 0"):
         agent_control.init(
-            agent_name="Negative Interval Agent",
-            agent_id="550e8400-e29b-41d4-a716-446655440000",
+            agent_name="negative-interval-agent",
             policy_refresh_interval_seconds=-1,
         )
 
--- a/sdks/python/tests/test_integration_agents.py
+++ b/sdks/python/tests/test_integration_agents.py
@@ -144,7 +144,7 @@ async def test_list_agent_controls_typed_returns_model(
     # WHEN: controls are requested via the typed API wrapper.
     response = await agent_control.agents.list_agent_controls_typed(
         client,
-        test_agent["agent_id"],
+        test_agent["agent_name"],
     )
 
     # THEN: a typed model is returned.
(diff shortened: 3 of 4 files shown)


sdks/python/src/agent_control/__init__.py
```diff
@@ -435,6 +403,12 @@ async def handle(message: str):
     # Validate agent_id is a UUID (string or UUID instance)
     _agent_uuid = ensure_uuid(agent_id)
 
+    if policy_refresh_interval_seconds < 0:
+        raise ValueError("policy_refresh_interval_seconds must be >= 0")
+
+    # Re-init behavior: always stop existing loop before mutating shared agent/session globals.
+    _stop_policy_refresh_loop()
```
namrataghadi-galileo, line 404: There's a race condition in the init() function that can cause controls to be overwritten when re-initializing while the background refresh thread is still running.

The comment on [line 257] `# Re-init behavior: always stop existing loop before mutating shared agent/session globals.` acknowledges the intent

However, _stop_policy_refresh_loop() uses a 2-second timeout that may not be sufficient for slow network calls:

Better way to handle this is
```
def init(...):
    # Stop the loop first (outside lock)
    _stop_policy_refresh_loop()
    
    # Then protect the entire state mutation with the lock
    with _refresh_lock:
        # Verify we're not racing with a refresh
        if _refresh_thread is not None and _refresh_thread.is_alive():
            raise RuntimeError("Policy refresh thread still running")
        
        # Now safely mutate globals
        global _current_agent, _server_url, _api_key, _server_controls
        _current_agent = Agent(...)
        _server_url = server_url or ...
        # ... rest of init
```

sdks/python/src/agent_control/__init__.py
```diff
@@ -254,6 +151,88 @@ def get_server_controls() -> list[dict[str, Any]] | None:
...
 
 
+def _publish_server_controls(
+    controls: list[dict[str, Any]] | None,
+) -> list[dict[str, Any]] | None:
+    """Publish controls using swap-only snapshot semantics."""
+    global _server_controls
+
+    if controls is None:
+        _server_controls = None
```
namrataghadi-galileo, line 159: `_server_controls = None` ← Readers might see this mid-mutation
Use atomic updates or protect with a lock:
```

def _publish_server_controls(
    controls: list[dict[str, Any]] | None,
) -> list[dict[str, Any]] | None:
    global _server_controls
    
    with _refresh_lock:
        if controls is None:
            _server_controls = None
            return None
        _server_controls = list(controls)
        return _server_controls
```

lan17: These are valid concerns around race conditions but I feel like it's fine in our case since the source of truth is still the same.

So if we overwrite it's fine since it's coming from same source.

Hence the swap global control list semantics here and not worrying about potential race conditions.

It would be different if we were mutating shared list, but the logic is to just overwrite entire data structure, so it should be fine.

PR #45, opened 2026-02-26
fix(server)!: Feature/56688 fix docker and create bash

namrataghadi-galileo:
- Multi-platform builds (linux/amd64, linux/arm64) for Apple Silicon
- DATABASE_URL environment variable support
- Production docker-compose.yml with server service
- Documentation updates
- Automatic migrations on startup


server/src/agent_control_server/main.py
```diff
@@ -78,6 +78,28 @@ async def lifespan(app: FastAPI) -> AsyncGenerator[None, None]:
...
+        from alembic import command
+        from alembic.config import Config
+        import os
+        from pathlib import Path
+
+        # Find alembic.ini (in server directory or parent)
+        current_dir = Path(__file__).parent
+        alembic_ini = current_dir.parent.parent / "alembic.ini"
+
+        if alembic_ini.exists():
```
namrataghadi-galileo: If migrations fail (rare case):

Server logs a warning: "Failed to run migrations: {error}"
Server continues starting up: "Continuing without migrations - database may not be initialized"
This prevents the server from crashing, but users will know migrations didn't run

docker-compose.yml
```diff
@@ -19,5 +31,21 @@ services:
       retries: 10
     restart: unless-stopped
 
+  server:
+    image: galileoai/agent-control-server:latest
```
lan17, line 36: should this file be moved to root folder?

server/src/agent_control_server/main.py
lan17: we should not do this in main.py, we should do this via existing make file targets.  

We can update the entry point script ofr docker or elsewhere to make sure the alembic runs, but IMO this does not belong here in main.py

docker-compose.yml
namrataghadi-galileo: sure

server/src/agent_control_server/main.py
namrataghadi-galileo: sure

server/src/agent_control_server/config.py
```diff
@@ -86,8 +86,12 @@ class AgentControlServerDatabaseConfig(BaseSettings):
 
     def get_url(self) -> str:
         """Get database URL, preferring explicit url if set."""
-        if self.url:
-            return self.url
+        import os
```
lan17: can we avoid local imports? they are an anti pattern in python

namrataghadi-galileo: done

PR #49, opened 2026-02-27
fix(docs): add setup script

namrataghadi-galileo:
Add a setup.sh script that is complete setup in "one click".
Copy this script to local instead of cloning the repo and run it in terminal. You should be up and running with agent-control

Optionally:
Option 1: Use a GitHub Gist (Recommended)
Create a GitHub Gist and users can fetch it directly:

# Users run this one-liner
curl -fsSL https://gist.githubusercontent.com/YOUR_USERNAME/GIST_ID/raw/setup.sh | bash
Steps:

Go to https://gist.github.com/
Create a new gist with setup.sh content
Make it public
Get the raw URL
Then update the README to:

# Real Quick Start - One command setup
curl -fsSL https://gist.githubusercontent.com/.../setup.sh | bash
Option 2: Host in GitHub Releases
Attach the script to each release:

# Users download from latest release
curl -fsSL https://github.com/galileoai/agent-control/releases/latest/download/setup.sh -o setup.sh
chmod +x setup.sh
./setup.sh

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ Get up and running with Agent Control in **one command**:
 ### Option 1: Quick Install (Recommended)
 
 ```bash
-curl -fsSL https://raw.githubusercontent.com/agentcontrol/agent-control/main/setup.sh | sh
+curl -fsSL -H "Authorization: token YOUR_GITHUB_TOKEN" https://raw.githubusercontent.com/agentcontrol/agent-control/main/setup.sh | sh
 ```
 
 > ⚠️ **Note**: This will automatically download and run the setup script. [Review the script](https://github.com/agentcontrol/agent-control/blob/main/setup.sh) before running if you prefer.


setup.sh
```diff
@@ -1,126 +1,111 @@
...
-    local name=$2
-    
-    echo -e "${BLUE}Setting up $name...${NC}"
-    cd "$dir"
-    
-    if uv sync; then
-        echo -e "${GREEN}✓ $name installed successfully${NC}"
-    else
-        echo -e "${RED}✗ Failed to install $name${NC}"
+if [[ ! "$PYTHON_VERSION" =~ ^3\.12 ]]; then
```
lan17: check should be for 3.12+ so 3.14 should not trigger this

namrataghadi-galileo: done

PR #59, opened 2026-03-05
feat(sdk): 57143 strands extra

namrataghadi-galileo:
Summary
Added Strands integration as a vendor-owned subpackage in agent-control-sdk[strands-agents] (AgentControlHook + AgentControlSteeringHandler) and updated Strands examples to use the new import path with controls attached directly to agents.
Scope
User-facing/API changes:
New integration import path: agent_control.integrations.strands.
New optional extra: agent-control-sdk[strands-agents] (min Strands version 1.26.0).
Internal changes:
Added Strands integration modules to SDK.
Added evaluate_controls tests and restored missing convenience helpers.
Updated examples to remove common/ and policy wiring; controls are attached to agents.
Out of scope:
Strands repo changes beyond example integration.
Risk and Rollout
Risk level: medium
Rollback plan:
Revert SDK integration subpackage and example changes; release prior SDK version.
Testing
 Added or updated automated tests
 Ran make check (or explained why not)
 Manually verified behavior
Checklist
 Linked issue/spec (if applicable)
 Updated docs/examples for user-facing changes
 Included any required follow-up tasks

### DIFF
--- a/sdks/python/src/agent_control/integrations/strands/hook.py
+++ b/sdks/python/src/agent_control/integrations/strands/hook.py
@@ -35,27 +35,32 @@ logger = logging.getLogger(__name__)
 def _action_error(result: EvaluationResult) -> tuple[str, Exception] | None:
     """Return the first blocking action as an exception."""
 
-    match = next((m for m in (result.matches or []) if m.action in ("deny", "steer")), None)
-    if not match:
-        return None
-
-    msg = getattr(getattr(match, "result", None), "message", None) or result.reason
-    msg = msg or f"Control '{match.control_name}' triggered"
-
-    if match.action == "deny":
+    matches = result.matches or []
+    deny_match = next((m for m in matches if m.action == "deny"), None)
+    if deny_match:
+        msg = getattr(getattr(deny_match, "result", None), "message", None) or result.reason
+        msg = msg or f"Control '{deny_match.control_name}' triggered"
         deny_err = ControlViolationError(
-            control_id=match.control_id,
-            control_name=match.control_name,
+            control_id=deny_match.control_id,
+            control_name=deny_match.control_name,
             message=msg,
-            metadata=getattr(match.result, "metadata", None),
+            metadata=getattr(deny_match.result, "metadata", None),
         )
         return "deny", deny_err
 
-    ctx = getattr(match, "steering_context", None)
+    steer_match = next((m for m in matches if m.action == "steer"), None)
+    if not steer_match:
+        return None
+
+    msg = getattr(getattr(steer_match, "result", None), "message", None) or result.reason
+    msg = msg or f"Control '{steer_match.control_name}' triggered"
+    ctx = getattr(steer_match, "steering_context", None)
     ctx_msg = getattr(ctx, "message", None) if ctx else None
     steer_err = ControlSteerError(
-        control_name=match.control_name,
-        message=f"Steering required [{match.control_name}]: {msg}",
+        control_id=steer_match.control_id,
+        control_name=steer_match.control_name,
+        message=f"Steering required [{steer_match.control_name}]: {msg}",
+        metadata=getattr(steer_match.result, "metadata", None),
         steering_context=ctx_msg or msg,
     )
     return "steer", steer_err
@@ -102,6 +107,7 @@ class AgentControlHook(HookProvider):
         step_name: str,
         input: Any | None = None,
         output: Any | None = None,
+        context: dict[str, Any] | None = None,
         step_type: Literal["tool", "llm"] = "llm",
         stage: Literal["pre", "post"] = "pre",
         use_runtime_error: bool = False,
@@ -110,10 +116,23 @@ class AgentControlHook(HookProvider):
             step_name=step_name,
             input=input,
             output=output,
+            context=context,
             step_type=step_type,
             stage=stage,
             agent_name=self.agent_name,
         )
+
+        if result.errors:
+            error_names = ", ".join(
+                e.control_name for e in result.errors if getattr(e, "control_name", None)
+            )
+            self._raise_error(
+                RuntimeError(
+                    "Control evaluation failed; execution blocked for safety. "
+                    f"Errors: {error_names or 'unknown'}"
+                ),
+                use_runtime_error,
+            )
 
         action = _action_error(result)
         if action:
@@ -205,19 +224,24 @@ class AgentControlHook(HookProvider):
         )
 
     async def check_after_model(self, event: AfterModelCallEvent) -> None:
-        _, output_text = self._extract_messages(event)
+        input_text, output_text = self._extract_messages(event)
+        context = self._extract_context(event)
         await self._evaluate_and_enforce(
             step_name="check_after_model",
+            input=input_text,
             output=output_text,
+            context=context,
             step_type="llm",
             stage="post",
         )
 
     async def check_before_tool(self, event: BeforeToolCallEvent) -> None:
         tool_name, tool_input = self._extract_tool_data(event)
+        context = self._extract_context(event)
         await self._evaluate_and_enforce(
             step_name=tool_name,
             input=tool_input,
+            context=context,
             step_type="tool",
             stage="pre",
             use_runtime_error=True,
@@ -226,10 +250,12 @@ class AgentControlHook(HookProvider):
     async def check_after_tool(self, event: AfterToolCallEvent) -> None:
         tool_name, tool_output = self._extract_tool_data(event)
         tool_input = event.tool_use.get("input", {}) if event.tool_use else {}
+        context = self._extract_context(event)
         await self._evaluate_and_enforce(
             step_name=tool_name,
             input=tool_input,
             output=tool_output,
+            context=context,
             step_type="tool",
             stage="post",
             use_runtime_error=True,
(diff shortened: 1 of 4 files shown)


examples/strands_agents/README.md
```diff
@@ -0,0 +1,138 @@
+# AgentControl + Strands Integration
```
yashsheth46, line 1: Nit: Can we name the folder just strands-agents instead of strands_integration?

examples/strands_integration/README.md
```diff
@@ -0,0 +1,138 @@
...
+
+**Basic setup**:
+```python
+hook = AgentControlHook(
+    agent_name="my-customer-agent",
+    enable_logging=True
+)
+```
+
+**Steering integration (dual-hook)**:
```
yashsheth46: Remove (dual-hook) ? Can avoid adding more nomenclature since that will not reflect in Strands docs

examples/strands_integration/README.md
```diff
@@ -0,0 +1,138 @@
...
+- `AgentControlHook` for tool-stage deny (hard blocks)
+- `AgentControlSteeringHandler` for LLM steer → `Guide()` (corrective guidance)
+
+See `steering_demo/` for complete implementation.
```
yashsheth46: See 'steering_demo/README.md' for ....

sdks/python/src/agent_control/integrations/strands/steering.py
```diff
@@ -0,0 +1,141 @@
...
+                        self.agent_name,
+                        deny_match.control_name,
+                    )
+                raise ControlViolationError(message=error_msg)
+
+        except ControlViolationError:
+            raise
+        except Exception as exc:
+            logger.debug(
+                "agent=<%s>, error=<%s> | steering evaluation failed",
```
lan17: This failure path currently logs at `debug` and drops the stack trace, then returns `Proceed`. If evaluation fails because the server is unavailable, the SDK is misconfigured, or an evaluator crashes, this should be surfaced at least as `warning` and probably with `exc_info=True` so operators can actually diagnose why enforcement was skipped.

sdks/python/src/agent_control/integrations/strands/hook.py
```diff
@@ -0,0 +1,340 @@
...
+            step_name=tool_name,
+            input=tool_input,
+            step_type="tool",
+            stage="pre",
+            use_runtime_error=True,
+        )
+
+    async def check_after_tool(self, event: AfterToolCallEvent) -> None:
+        tool_name, tool_output = self._extract_tool_data(event)
+        await self._evaluate_and_enforce(
```
lan17, line 254: This post-stage hook passes only `output` into `evaluate_controls()`. That helper then synthesizes empty default input values, so any post-stage control that selects `input`, `context`, or `*` will no longer see the real request payload under the Strands integration. The decorator path preserves both input and output, so this looks like a silent behavior regression.

sdks/python/src/agent_control/__init__.py
```diff
@@ -707,9 +702,8 @@ async def remove_all_agent_policies(
 ) -> dict[str, Any]:
     """Remove all policy associations from an agent."""
     _final_server_url = server_url or os.getenv('AGENT_CONTROL_URL') or 'http://localhost:8000'
-
     async with AgentControlClient(base_url=_final_server_url, api_key=api_key) as client:
-        return await agents.remove_all_agent_policies(client, agent_name)
+        return await agents.remove_agent_policy(client, agent_name)
```
lan17: Why does `remove_all_agent_policies()` need to call the singular compatibility endpoint here instead of the dedicated remove-all `/policies` route? This changes the helper from idempotent success to a 404 when an agent has no policies, which looks like a public API behavior change rather than an internal refactor.

sdks/python/src/agent_control/integrations/strands/hook.py
```diff
@@ -0,0 +1,340 @@
...
+
+    match = next((m for m in (result.matches or []) if m.action in ("deny", "steer")), None)
+    if not match:
+        return None
+
+    msg = getattr(getattr(match, "result", None), "message", None) or result.reason
+    msg = msg or f"Control '{match.control_name}' triggered"
+
+    if match.action == "deny":
+        deny_err = ControlViolationError(message=f"Policy violation [{match.control_name}]: {msg}")
```
lan17: `ControlViolationError` is being constructed from a preformatted message here, but without passing `control_name`. That leaves `err.control_name` as `unknown`, so `_evaluate_and_enforce()` forwards the wrong control into `on_violation_callback` and any downstream logging or telemetry. Please pass the control metadata explicitly instead of only embedding it in the message string.

sdks/python/src/agent_control/integrations/strands/steering.py
```diff
@@ -0,0 +1,141 @@
...
+            if deny_match:
+                msg = getattr(getattr(deny_match, "result", None), "message", None) or result.reason
+                error_msg = f"Policy violation [{deny_match.control_name}]: {msg}"
+                if self.enable_logging:
+                    logger.debug(
+                        "agent=<%s>, control=<%s> | deny raised",
+                        self.agent_name,
+                        deny_match.control_name,
+                    )
+                raise ControlViolationError(message=error_msg)
```
lan17: Same issue here: the deny exception is created from a formatted message but without `control_name`, so callers only see `unknown` on the exception object even though the matched control is available. That makes the raised error string look right while callback and programmatic consumers lose the actual control identity.

examples/strands_agents/README.md
namrataghadi-galileo: done

examples/strands_integration/README.md
namrataghadi-galileo: sure.

examples/strands_integration/README.md
namrataghadi-galileo: done

sdks/python/src/agent_control/integrations/strands/steering.py
namrataghadi-galileo: thanks. done
```
logger.warning(
                "agent=<%s> | steering evaluation failed; proceeding without enforcement",
                self.agent_name,
                exc_info=True,
            )
```

sdks/python/src/agent_control/integrations/strands/hook.py
namrataghadi-galileo: fixed.

sdks/python/src/agent_control/__init__.py
namrataghadi-galileo: changed back to `remove_all`

sdks/python/src/agent_control/integrations/strands/hook.py
namrataghadi-galileo: thanks .. good catch. fixed

sdks/python/src/agent_control/integrations/strands/steering.py
namrataghadi-galileo: fixed

sdks/python/src/agent_control/integrations/strands/steering.py
```diff
@@ -0,0 +1,145 @@
...
+        except ControlViolationError:
+            raise
+        except Exception:
+            logger.warning(
+                "agent=<%s> | steering evaluation failed; proceeding without enforcement",
+                self.agent_name,
+                exc_info=True,
+            )
+
+        self.last_steer_info = None
```
lan17, line 130: The logging change helps, but this still fails open: any exception from `evaluate_controls()` other than an already-materialized `ControlViolationError` falls through to `Proceed`. From a business-logic standpoint we should still respect the evaluation outcome here instead of silently allowing execution when enforcement cannot complete. At minimum this needs an explicit fail-open/fail-closed policy, because right now server/sdk/evaluator failures bypass the control path.

sdks/python/src/agent_control/integrations/strands/steering.py
```diff
@@ -0,0 +1,145 @@
...
+
+        if self.enable_logging:
+            logger.debug(
+                "agent=<%s>, output_len=<%d> | checking output",
+                self.agent_name,
+                len(output_text),
+            )
+
+        try:
+            result = await agent_control.evaluate_controls(
```
lan17, line 60: This still evaluates post-model controls with only `output`. `evaluate_controls()` synthesizes missing `input` values, so any post-stage control that selects `input`, `context`, or `*` cannot see the original request payload on this path. The tool post hook was fixed, but the LLM/steering post path still appears to lose half the step state.

sdks/python/src/agent_control/integrations/strands/steering.py
namrataghadi-galileo: okay. Will keep it fail-closed for now.

sdks/python/src/agent_control/integrations/strands/steering.py
namrataghadi-galileo: Fixed the LLM/steering post path to pass input alongside output, so post-stage controls that select input, context, or * can see the original request payload.

namrataghadi-galileo: a few more gaps were caught by codex. so fixed them too. thanks

PR #60, opened 2026-03-05
feat(evaluators): add Cisco AI Defense evaluator and examples

rucpande:
- Add contrib package at evaluators/contrib/ai_defense
- Add examples under examples/ai_defense
- Update Makefile with AI Defense targets
- Update evaluators/builtin/pyproject.toml for discovery
- Update .gitignore

## Summary
- What changed and why.

## Scope
- User-facing/API changes:
- Internal changes:
- Out of scope:

## Risk and Rollout
- Rollback plan:

## Testing
- [x] Ran `make check` (or explained why not)

## Checklist


Makefile
```diff
@@ -232,3 +236,32 @@ galileo-typecheck:
 
 galileo-build:
 	$(MAKE) -C $(GALILEO_DIR) build
+
+## Single target for Cisco AI Defense contrib
+aidefense-build:
```
lan17: We should probably keep contrib specific package targets out of main `Makefile`.  What do you think @abhinav-galileo ?

evaluators/contrib/cisco/src/agent_control_evaluator_cisco/ai_defense/evaluator.py
```diff
@@ -0,0 +1,165 @@
...
+            strategy=self.config.messages_strategy,
+            payload_field=self.config.payload_field,
+        )
+        if not messages:
+            return EvaluatorResult(matched=False, confidence=1.0, message="No data to inspect")
+
+        # Resolve API key and endpoint
+        try:
+            api_key = _load_api_key(self.config.api_key_env)
+        except Exception as e:  # noqa: BLE001
```
abhinav-galileo: **Must fix: Missing API key ignores `on_error="deny"` — security bypass**

This exception handler always returns `matched=False`, ignoring `config.on_error`. When configured fail-closed (`on_error="deny"`), a missing API key silently fails open — the evaluator allows content through instead of blocking.

This should mirror the API-call exception handler at line 150 which correctly respects `on_error`:
```python
fallback = self.config.on_error
matched = fallback == "deny"
return EvaluatorResult(
    matched=matched,
    confidence=0.0,
    message=str(e),
    metadata={"error": str(e), "error_type": type(e).__name__, "fallback_action": fallback},
    error=None if matched else str(e),
)
```

No test covers this path — please add one with `on_error="deny"` and a missing env var.

examples/ai_defense/examples/ai_defense/register_agent.py
```diff
@@ -0,0 +1,80 @@
+"""Register an agent once and persist its ID locally for reuse.
```
abhinav-galileo: **Must fix: Duplicate file**

This file is an exact copy of `examples/ai_defense/register_agent.py`. Looks like an accidental nested directory — please delete `examples/ai_defense/examples/`.

evaluators/contrib/ai_defense/src/agent_control_evaluator_ai_defense/chat_inspect/evaluator.py
```diff
@@ -0,0 +1,165 @@
...
+        # Derive endpoint from region or explicit api_url
+        if self.config.api_url:
+            endpoint_url = self.config.api_url
+        else:
+            base_url = REGION_BASE_URLS.get(self.config.region or "us", REGION_BASE_URLS["us"])
+            endpoint_url = build_endpoint(base_url)
+
+        # Prepare or reuse client
+        timeout_s = max(0.001, float(self.config.timeout_ms) / 1000.0)
+        if self._client is None:
```
abhinav-galileo: **Should fix: Cached client won't pick up rotated API key**

The `AIDefenseClient` is created once here and cached, but `_load_api_key()` runs on every `evaluate()` call above. If the API key rotates in the environment, the cached client keeps using the old key.

Either recreate the client when the key changes (compare against `self._client.api_key`), or load the key once in `__init__` and skip re-reading it each call.

examples/ai_defense/chat_guarded.py
```diff
@@ -0,0 +1,87 @@
...
+    api_key = os.getenv("AGENT_CONTROL_API_KEY", "")
+    if not api_key:
+        print("❌ Missing AGENT_CONTROL_API_KEY")
+        return 2
+
+    try:
+        agent_control.init(
+            agent_name=args.agent_name,
+            agent_id=args.agent_id,
+            url=url,
```
abhinav-galileo: **Should fix: Incorrect kwarg — `url` vs `server_url`**

The SDK parameter is `server_url`, not `url`. This kwarg is silently ignored via `**kwargs`, so `init()` falls back to the `AGENT_CONTROL_URL` env var / localhost default. Fix for correctness and clarity.

Same issue in `register_agent.py`, `chat_guarded_all.py`, and `chat_guarded_post.py`.

evaluators/contrib/ai_defense/tests/test_chat_inspect_evaluator.py
```diff
@@ -0,0 +1,113 @@
...
+        captured["messages"] = messages
+        return {"is_safe": True}
+
+    monkeypatch.setattr(AIDefenseClient, "chat_inspect", capture, raising=True)
+
+    cfg = ChatInspectEvaluatorConfig(messages_strategy="single", payload_field="input")
+    ev = ChatInspectEvaluator(cfg)
+    _ = await ev.evaluate("hello world")
+    assert captured["messages"] == [{"role": "user", "content": "hello world"}]
+
```
abhinav-galileo: **Should fix: Test coverage gaps**

The existing tests cover the core paths well, but given this is a security-critical evaluator, please add tests for:
- Missing API key with `on_error="deny"` (the must-fix security bypass above)
- `api_url` override path
- Happy-path safe/unsafe `is_safe` mapping
- Client reuse behavior

evaluators/contrib/ai_defense/src/agent_control_evaluator_ai_defense/chat_inspect/client.py
```diff
@@ -0,0 +1,85 @@
...
+# Thin REST client for Cisco AI Defense Chat Inspection.
+# Uses httpx.AsyncClient and the OpenAPI-defined endpoint/header.
+from dataclasses import dataclass
+from typing import Any
+
+try:
+    import httpx
+
+    AI_DEFENSE_HTTPX_AVAILABLE = True
+except Exception:  # noqa: BLE001
```
abhinav-galileo: **Should fix: Narrow to `ImportError`**

Catching all exceptions here could mask real errors (e.g., a partially-loaded httpx with a broken transitive dependency). `ImportError` is the precise exception for missing packages.

examples/cisco_ai_defense/seed_policy.py
```diff
@@ -0,0 +1,207 @@
...
+    Notes:
+        The server API does not expose list/search by policy name. To be
+        idempotent on reruns, if creation returns HTTP 409 we attempt to
+        reuse the policy currently assigned to the target agent. This
+        covers the common rerun path where a prior invocation already
+        created and assigned the policy. If the policy exists but is not
+        assigned to the agent, the script will raise with guidance.
+    """
+    resp = await client.http_client.put("/api/v1/policies", json={"name": name}, headers=_headers())
+    if resp.status_code == 409:
```
abhinav-galileo: **Should fix: Policy conflict handling can mutate an unrelated policy**

On name conflict (409), this reuses whatever policy is currently assigned to the agent. If the agent already has a *different* policy assigned (e.g., `production-policy`), this script will attach AI Defense controls to that unrelated policy — silently modifying it.

Should verify the returned policy's name matches `name` before reusing, or bail with an error message instead of attaching controls to the wrong policy.

evaluators/contrib/cisco/pyproject.toml
```diff
@@ -0,0 +1,37 @@
+[project]
```
abhinav-galileo, line 1: **Should fix: Follow contrib template naming convention**

The [contrib template](https://github.com/agentcontrol/agent-control/blob/main/evaluators/contrib/template/README.md) uses the **org name** as the top-level namespace. We should follow this - e.g., with org=`cisco`:

| Convention | Expected | This PR |
|---|---|---|
| Directory | `evaluators/contrib/cisco/` | `evaluators/contrib/ai_defense/` |
| PyPI package | `agent-control-evaluator-cisco` | `agent-control-evaluator-ai-defense` |
| Python package | `agent_control_evaluator_cisco` | `agent_control_evaluator_ai_defense` |
| Entry point | `cisco.chat_inspect` | `ai_defense.chat_inspect` |
| Class name | `CiscoChatInspectEvaluator` | `ChatInspectEvaluator` |

This way, if Cisco contributes additional evaluators later, they'd naturally live under `evaluators/contrib/cisco/` and share the same package.

We should also add the `license`, `authors`, and `[tool.uv.sources]` fields from the template's `pyproject.toml`.

abhinav-galileo: cc: @lan17 @namrataghadi-galileo @yashsheth46

Makefile
abhinav-galileo: Agreed.

And should migrate galileo-* as well.

rucpande: removed aidefense related targets

evaluators/contrib/cisco/src/agent_control_evaluator_cisco/ai_defense/evaluator.py
rucpande: done

evaluators/contrib/ai_defense/src/agent_control_evaluator_ai_defense/chat_inspect/evaluator.py
rucpande: fixed

evaluators/contrib/ai_defense/tests/test_chat_inspect_evaluator.py
rucpande: done

examples/ai_defense/chat_guarded.py
rucpande: fixed

examples/cisco_ai_defense/seed_policy.py
rucpande: good to have robust example code. made the changes

evaluators/contrib/cisco/pyproject.toml
rucpande: Let me check what we want to do here. It'll probably be cisco_ai if anything

evaluators/contrib/ai_defense/src/agent_control_evaluator_ai_defense/chat_inspect/client.py
rucpande: fixed.

evaluators/contrib/cisco/pyproject.toml
rucpande: modified

examples/cisco_ai_defense/seed_policy.py
```diff
@@ -0,0 +1,209 @@
...
+        f"/api/v1/policies/{policy_id}/controls/{control_id}", headers=_headers()
+    )
+    resp.raise_for_status()
+
+
+async def _assign_policy(client: AgentControlClient, agent_name: str, policy_id: int) -> None:
+    resp = await client.http_client.post(
+        f"/api/v1/agents/{agent_name}/policy/{policy_id}", headers=_headers()
+    )
+    resp.raise_for_status()
```
abhinav-galileo: **Policy assignment has replace semantics.** `POST /agents/{name}/policy/{id}` deletes all existing agent-policy rows before inserting (see `server/src/agent_control_server/endpoints/agents.py:936`). Running this against an agent that already has policies will silently drop them.

Consider attaching controls directly to the agent instead of going through a policy. All existing examples on `main` use the direct approach (e.g. `steer_action_demo/setup_controls.py`, `langchain/setup_sql_controls.py`, `crewai/setup_content_controls.py`):

```python
for control_id in control_ids:
    try:
        await agents.add_agent_control(client, agent_name, control_id)
    except httpx.HTTPStatusError as exc:
        if exc.response.status_code != 409:
            raise
```

`add_agent_control` is additive and returns 409 on duplicate, so it's safe to rerun. Unless this example specifically needs to demonstrate policy-based management, switching to the direct pattern would be simpler, consistent with other examples, and eliminate the destructive behavior.

examples/cisco_ai_defense/seed_policy.py
```diff
@@ -0,0 +1,209 @@
...
+        "action": {"decision": "deny"},
+        "tags": ["ai-defense", "chat-inspect", stage, "llm"],
+    }
+
+    resp = await client.http_client.put(
+        f"/api/v1/controls/{control_id}/data",
+        json={"data": control_def},
+        headers=_headers(),
+    )
+    resp.raise_for_status()
```
abhinav-galileo: **Overwrites globally shared controls.** `ai-defense-pre` / `ai-defense-post` are looked up by fixed name and unconditionally overwritten. Controls are globally unique and referenced via many-to-many by other policies - rerunning this demo can silently change behavior for unrelated policies.

Consider using unique names per demo run (e.g. `ai-defense-pre-{agent_name}`) or skipping the update if the existing control already matches.

examples/cisco_ai_defense/register_agent.py
```diff
@@ -0,0 +1,79 @@
...
+
+import argparse
+import os
+import uuid
+
+import agent_control
+import httpx
+
+
+def _read_or_create_agent_id(path: str = ".agent_id") -> str:
```
abhinav-galileo: **`agent_id` is not a real identity concept.** Agents are keyed by `agent_name` in both the public model (`agent_control_models/agent.py:37`) and the server table. Passing `agent_id` to `agent_control.init()` silently stuffs it into `agent_metadata` as an unknown kwarg (see `sdks/python/src/agent_control/init.py:315`). The `.agent_id` file and README instructions create an illusion of per-instance identity that doesn't exist.

Recommend dropping `agent_id` entirely and keying on `agent_name` throughout.

evaluators/contrib/cisco/src/agent_control_evaluator_cisco/ai_defense/evaluator.py
```diff
@@ -0,0 +1,195 @@
...
+                return msgs
+        # Fallback to single
+
+    role = "assistant" if payload_field == "output" else "user"
+    content = "" if data is None else str(data)
+    return [{"role": role, "content": content}]
+
+
+@register_evaluator
+class CiscoAIDefenseEvaluator(Evaluator[CiscoAIDefenseConfig]):
```
abhinav-galileo, line 54: **Missing `is_available()` override.** `AI_DEFENSE_HTTPX_AVAILABLE` is set in `client.py` but never checked at registration time. If httpx isn't installed, the evaluator registers successfully and fails at runtime. The galileo contrib handles this:

```python
@classmethod
def is_available(cls) -> bool:
    from .client import AI_DEFENSE_HTTPX_AVAILABLE
    return AI_DEFENSE_HTTPX_AVAILABLE
```

evaluators/contrib/cisco/src/agent_control_evaluator_cisco/ai_defense/evaluator.py
```diff
@@ -0,0 +1,195 @@
...
+    )
+
+    config_model = CiscoAIDefenseConfig
+
+    def __init__(self, config: CiscoAIDefenseConfig) -> None:
+        self.config = config
+        self._client: AIDefenseClient | None = None
+        self._current_api_key: str | None = None
+        self._current_endpoint_url: str | None = None
+        self._current_timeout_s: float | None = None
```
abhinav-galileo: **Thread safety concern.** Evaluator instances are cached and reused across concurrent requests. `evaluate()` mutates `self._client` and the `_current_*` tracking fields without synchronization. Two concurrent calls during a key rotation can race on client creation/teardown. Consider an `asyncio.Lock` around the client-swap block.

evaluators/contrib/cisco/src/agent_control_evaluator_cisco/ai_defense/client.py
```diff
@@ -0,0 +1,86 @@
...
+            "Accept": "application/json",
+        }
+        if headers:
+            req_headers.update(headers)
+
+        payload: dict[str, Any] = {"messages": messages}
+        if metadata:
+            payload["metadata"] = metadata
+        if inspect_config:
+            payload["config"] = inspect_config
```
abhinav-galileo: Truthiness check skips empty dicts `{}`, which are valid per the OpenAPI spec.

```suggestion
        if metadata is not None:
            payload["metadata"] = metadata
        if inspect_config is not None:
            payload["config"] = inspect_config
```

evaluators/contrib/cisco/src/agent_control_evaluator_cisco/ai_defense/config.py
```diff
@@ -0,0 +1,32 @@
...
+        payload_field: Force single-message role: input→user, output→assistant
+        messages_strategy: "single" (synthesize) or "history" (pass-through messages)
+        metadata: Optional metadata object to include (OpenAPI spec)
+        inspect_config: Optional Inspect API config passthrough (see OpenAPI spec)
+    """
+
+    api_key_env: str = "AI_DEFENSE_API_KEY"
+    region: Literal["us", "ap", "eu"] | None = "us"
+    api_url: str | None = None
+    timeout_ms: int = 15_000
```
abhinav-galileo: Currently accepts 0 or negative values. The evaluator silently clamps with `max(0.001, ...)` but better to reject bad config at load time.

```suggestion
    timeout_ms: int = Field(default=15_000, ge=1)
```

(Requires adding `from pydantic import Field` to imports.)

evaluators/contrib/cisco/src/agent_control_evaluator_cisco/ai_defense/evaluator.py
```diff
@@ -0,0 +1,195 @@
...
+                confidence=0.0,
+                message=str(e),
+                metadata={
+                    "error": str(e),
+                    "error_type": type(e).__name__,
+                    "fallback_action": fallback,
+                },
+                # On fail-closed, expose details via metadata only
+                error=None if matched else str(e),
+            )
```
abhinav-galileo: **Bug: setting `error` field bypasses `on_error="allow"`.** The engine treats any `EvaluatorResult` with `error` set as a hard fail-closed, regardless of `matched` (see `engine/src/agent_control_engine/core.py:284-336`). So when `on_error="allow"`, the evaluator returns `matched=False, error=str(e)` - but the engine ignores `matched` and blocks anyway.

The Luna2 evaluator handles this correctly by never setting the `error` field when it handles the error itself via `on_error`. Error details go in `metadata` only:

```python
# Luna2 pattern (evaluator_galileo/luna2/evaluator.py:324-331):
return EvaluatorResult(
    matched=(error_action == "deny"),
    confidence=0.0,
    message=f"...: {error}",
    metadata={"error": str(error), ...},
    # no error= field - let matched drive the decision
)
```

The same issue exists in the other error handler at line 193. Both should drop the `error=` argument.

PR #63, opened 2026-03-05
fix(sdk): fix observability events not reaching the server

abhinav-galileo:
## Summary

Addresses the issue where control execution stats do not appear reliably in the UI for local SDK runs (see [sc-57369](https://app.shortcut.com/galileo/story/57369)).

This PR now includes four fixes:

- **Dedicated EventBatcher worker loop**: replaced caller-loop-dependent batching with a dedicated daemon worker thread + event loop, so flushing works consistently across sync callers, repeated `asyncio.run(...)`, and async apps.
- **Single-owner flush execution**: `_flush_loop` is now the sole owner of `_flush()` execution (signal-driven wakeups), avoiding fire-and-forget concurrent flush task races.
- **Graceful shutdown drain**: shutdown now wakes the worker, drains pending/in-flight batches safely, and then stops/joins the worker thread, preventing silent event loss during shutdown.
- **Safer defaults for observability**:
  - `observability_enabled` default changed to `True`
  - `flush_interval` default changed from `10.0` to `5.0` seconds

## Test plan

- [x] `make sdk-lint`
- [x] `make sdk-typecheck`
- [x] `make sdk-test` (`284 passed, 1 skipped`)
- [x] Added/ran targeted regression test for shutdown race:
  - `test_shutdown_drains_inflight_flush_without_data_loss`
- [ ] Manual verification: run a local example agent and confirm UI stats appear within ~5s

### DIFF
--- a/ui/src/core/page-components/agent-detail/monitor/index.tsx
+++ b/ui/src/core/page-components/agent-detail/monitor/index.tsx
@@ -64,7 +64,7 @@ export function AgentsMonitor({
     isLoading,
     error,
   } = useAgentMonitor(agentUuid, apiTimeRange, {
-    refetchInterval: 5000, // Poll every 5 seconds
+    refetchInterval: 2000, // Poll every 2 seconds
     includeTimeseries: true, // Always fetch timeseries for trend chart
   });
 


sdks/python/src/agent_control/observability.py
```diff
@@ -353,29 +420,51 @@ def add_event(self, event: ControlExecutionEvent) -> bool:
...
+            logger.error("Flush loop started without a worker flush signal")
+            return
+
         while self._running:
             try:
-                await asyncio.sleep(self.flush_interval)
-                await self._flush()
+                await asyncio.wait_for(signal.wait(), timeout=self.flush_interval)
+                signal.clear()
+            except TimeoutError:
```
namrataghadi-galileo, line 471: catch `asyncio.TimeoutError` .. This I believe is builtin `TimeoutError`

sdks/python/src/agent_control/observability.py
```diff
@@ -295,28 +304,86 @@ def __init__(
...
+        """Start the dedicated worker thread and flush loop."""
         if self._running:
             return
 
         self._running = True
-
-        # Try to get current event loop, create new if needed
+        self._graceful_shutdown = False
+        self._worker_ready.clear()
+        self._loop = asyncio.new_event_loop()
```
namrataghadi-galileo: `new_event_loop` is created in the main thread, but `loop.run_until_complete(self._flush_loop())` is executed in worker thread. This can raise `RuntimeError` or behave inconsistently since event loops are thread bound. 
Fix: create the loop inside _run_worker_loop() and assign it there.

sdks/python/src/agent_control/observability.py
abhinav-galileo: asyncio.TimeoutError is TimeoutError since Python 3.11 (and the project targets Python 3.12 per pyproject.toml's target-version = "py312")

PR #64, opened 2026-03-05
feat(sdk): ts sdk

siddhant-galileo:
## Summary
- What changed and why.

# TypeScript SDK Status

## Completed

- `init()` registers agents via `initAgent` and sends `sdk_language: "typescript"`.
- Step auto-discovery/merge is implemented via `_control_registry.ts`.
- `control()` HOF patterns are implemented.
- Control actions and errors are aligned (`allow`, `deny`, `steer`, `warn`, `log`).
- Server-side enforcement blocks `execution="sdk"` controls for TypeScript agents (direct attach, policy attach/assign, and control data update paths).
- TypeScript example flows and SDK tests are passing.

## Remaining Gaps

- Trace context propagation parity with Python (`withTrace`-style shared trace/span context).
- Env-based SDK config parity (`AGENT_CONTROL_*` defaults in TS SDK init path).
- Name-mangling will cause issue with selector
- Python-only features intentionally not ported:
  - local evaluator execution (`execution="sdk"` in SDK runtime)
  - runtime schema derivation from function signatures
  - sync function wrapping parity for `control()`
 

## Decision Notes

- Observability for TS server-evaluated controls is server-owned; no SDK-side event batching for now to avoid duplicate event ingestion.

## Scope
- User-facing/API changes: TS SDK updated with notes above
- Internal changes: 
- Out of scope:

## Risk and Rollout
- Rollback plan:

## Testing

## Checklist

### DIFF
--- a/server/src/agent_control_server/services/sdk_compat.py
+++ b/server/src/agent_control_server/services/sdk_compat.py
@@ -13,7 +13,7 @@ def is_typescript_agent_metadata(agent_metadata: dict[str, Any]) -> bool:
     nested = agent_metadata.get("agent_metadata")
     if not isinstance(nested, dict):
         return False
-    return nested.get("sdk_language", "").lower() == "typescript"
+    return str(nested.get("sdk_language", "")).lower() == "typescript"
 
 
 def is_local_execution_control(control_data: dict[str, Any]) -> bool:


sdks/typescript/src/generated/hooks/hooks.ts
```diff
@@ -45,6 +47,7 @@ export class SDKHooks implements Hooks {
         this.registerAfterErrorHook(hook);
       }
     }
+    initHooks(this);
```
abhinav-galileo: **blocking: generated SDK output diverges from committed output**

This file, `registration.ts`, and `validation-error.ts` are all Speakeasy-generated but have been hand-edited. CI's `generate-check` regenerates from the OpenAPI spec and diffs - these edits cause the `sdk-ts-ci` failure.

The sustainable fix is upstream: update the OpenAPI spec and/or Speakeasy config to produce the desired output, rather than editing generated files directly. Otherwise every regeneration will produce this drift.

server/src/agent_control_server/endpoints/agents.py
```diff
@@ -102,11 +103,18 @@ def _validate_controls_for_agent(agent: Agent, controls: list[Control]) -> list[
     except ValidationError:
         return [f"Agent '{agent.name}' has corrupted data"]
 
+    is_ts_agent = is_typescript_agent_metadata(agent_data.agent_metadata)
```
abhinav-galileo, line 132: **blocking: `initAgent` can convert an agent to TypeScript without revalidating existing controls**

`_validate_controls_for_agent` correctly checks for TS + `execution="sdk"` incompatibility here, but `initAgent` never calls it after updating `agent_metadata`. This means: create a Python agent, attach `execution="sdk"` controls, then re-init with `{"sdk_language": "typescript"}` - 200 OK, incompatible controls silently remain.

Suggested fix: after the metadata update and before the commit in `initAgent`, call `_validate_controls_for_agent(existing, controls)` and reject if incompatible controls are found - consistent with how the attach/assign paths behave.

examples/customer_support_agent_ts/src/setup-controls.ts
```diff
@@ -0,0 +1,224 @@
...
+ *
+ * Run once after starting the server:
+ *   npm run setup
+ */
+
+import { AgentControlClient, type ControlDefinitionInput } from "agent-control";
+
+const serverUrl = process.env.AGENT_CONTROL_URL ?? "http://localhost:8000";
+const apiKey = process.env.AGENT_CONTROL_API_KEY;
+const agentName = "customer-support-agent-ts-2";
```
abhinav-galileo: **medium: agent name mismatch between setup and runtime**

This is `"customer-support-agent-ts-2"` but `agent.ts:119` registers as `"customer-support-agent-ts"`. Running `pnpm run setup` then `pnpm start` as the README instructs won't work - controls land on the wrong agent.

sdks/typescript/src/client.ts
```diff
@@ -30,14 +32,29 @@ export class AgentControlClient {
   private options: AgentControlInitOptions | null = null;
   private sdk: AgentControlSDK | null = null;
 
-  init(options: AgentControlInitOptions): void {
+  async init(options: AgentControlInitOptions): Promise<void> {
```
abhinav-galileo, line 55: **medium: `init()` is now async but docs/examples don't reflect it**

This changed from sync `void` to `async Promise<void>` - a breaking change for existing consumers who call `init()` without `await`. The README snippet (line ~80) also shows non-awaited `init()`, so anyone copying it gets racey startup where `control()` wrappers can fire before registration completes.

Worth a semver-appropriate version bump and updating all doc snippets to use `await`.

examples/customer_support_agent_ts/src/setup-controls.ts
```diff
@@ -0,0 +1,224 @@
...
+  });
+  console.log(`Registered agent: ${agentName}`);
+
+  // Create or reuse policy
+  let policyId: number;
+  try {
+    const created = await client.policies.create({ name: policyName });
+    policyId = created.policyId;
+    console.log(`Created policy: ${policyName} (id=${policyId})`);
+  } catch {
```
abhinav-galileo: **low: silent catch blocks swallow all errors**

This and the catch blocks at lines 198 and 208 swallow everything including network failures and auth issues. Consider logging the caught error so users can distinguish "expected conflict" from "server unreachable."

server/src/agent_control_server/services/sdk_compat.py
```diff
@@ -0,0 +1,19 @@
+from __future__ import annotations
+
+from typing import Any
+
+
+def is_typescript_agent_metadata(agent_metadata: dict[str, Any]) -> bool:
+    """Return True when agent metadata identifies a TypeScript SDK agent."""
+    sdk_language = agent_metadata.get("sdk_language")
+    if not isinstance(sdk_language, str):
+        nested_metadata = agent_metadata.get("agent_metadata")
```
abhinav-galileo: **low: metadata shape ambiguity**

Checking both `agent_metadata.sdk_language` and `agent_metadata.agent_metadata.sdk_language` suggests the stored shape isn't formalized. If the double nesting is a real scenario from how `initAgent` stores data, a comment explaining why would help. If it's speculative, removing it keeps the contract clearer.

sdks/typescript/src/control.ts
```diff
@@ -1,17 +1,262 @@
...
+    type: stepType,
+    description: opts.description,
+    inputSchema: opts.inputSchema,
+    outputSchema: opts.outputSchema,
+    metadata: opts.metadata,
+  });
+
+  const wrapped = async (...args: TArgs): Promise<TResult> => {
+    const client = requireClient();
+    const { agentName } = client.config!;
```
abhinav-galileo: **low: non-null assertion**

`requireClient()` above checks `initialized` but the type system doesn't know `config` is non-null after that. Returning the config from `requireClient()` (e.g., `requireClient(): { client: AgentControlClient; agentName: string }`) would eliminate the assertion.

server/tests/test_policy_integration.py
```diff
@@ -24,6 +27,25 @@ def _create_agent(client: TestClient, name: str | None = None) -> tuple[str, str
     return agent_name, agent_name
 
 
+def _create_typescript_agent(client: TestClient, name: str | None = None) -> tuple[str, str]:
```
abhinav-galileo: **nit: duplicate helper**

Nearly identical to `_create_typescript_agent` in `test_controls_additional.py:36`. Consider extracting to `tests/utils.py`.

examples/customer_support_agent_ts/pnpm-lock.yaml
```diff
@@ -0,0 +1,359 @@
+lockfileVersion: '9.0'
```
abhinav-galileo, line 1: **nit: committed lock file for example**

359 lines of platform-specific esbuild entries. Examples typically don't need committed lock files - `pnpm install` generates this from `package.json`.

sdks/typescript/src/control.ts
```diff
@@ -1,17 +1,262 @@
...
+      { name: stepName, type: stepType, input: extractInput(args) },
+      () => fn(...args),
+    );
+  };
+
+  Object.defineProperty(wrapped, "name", { value: stepName });
+  return wrapped;
 }
+
+// NOTE: `guard()` and `check()` are temporarily disabled from the public API
```
abhinav-galileo: **nit: orphan comment**

If `guard()` and `check()` never shipped publicly, this note about them being "temporarily disabled" is unnecessary and may confuse readers.

sdks/typescript/src/generated/hooks/hooks.ts
siddhant-galileo: not sure why this is happening. I ran the required commands on local and pushed the changes. why is the diff not passing. cc: @lev

server/src/agent_control_server/endpoints/agents.py
siddhant-galileo: good callout. fixed here - https://github.com/agentcontrol/agent-control/pull/64/commits/7335a172ec9b6013e9a1fdb24deb4db53a0f90d6

sdks/typescript/src/client.ts
siddhant-galileo: yup. i have to do this. once I get a nod on sdk related changes, will update docs at all places in a single commit

examples/customer_support_agent_ts/pnpm-lock.yaml
siddhant-galileo: dont think this should be an issue. lockfiles are generally helpful

sdks/typescript/src/control.ts
siddhant-galileo: fixed https://github.com/agentcontrol/agent-control/pull/64/commits/c6237e57f0b3b34ccdf80877b7030e86837e37fc

server/tests/test_policy_integration.py
siddhant-galileo: fixed - https://github.com/agentcontrol/agent-control/pull/64/commits/c6237e57f0b3b34ccdf80877b7030e86837e37fc

sdks/typescript/src/control.ts
siddhant-galileo: fixed - https://github.com/agentcontrol/agent-control/pull/64/commits/c6237e57f0b3b34ccdf80877b7030e86837e37fc

server/src/agent_control_server/services/sdk_compat.py
siddhant-galileo: fixed - https://github.com/agentcontrol/agent-control/pull/64/commits/c6237e57f0b3b34ccdf80877b7030e86837e37fc

examples/customer_support_agent_ts/src/setup-controls.ts
siddhant-galileo: fixed - https://github.com/agentcontrol/agent-control/pull/64/commits/c6237e57f0b3b34ccdf80877b7030e86837e37fc

examples/customer_support_agent_ts/src/setup-controls.ts
siddhant-galileo: fixed - https://github.com/agentcontrol/agent-control/pull/64/commits/c6237e57f0b3b34ccdf80877b7030e86837e37fc

sdks/typescript/src/generated/hooks/hooks.ts
siddhant-galileo: fixed

sdks/typescript/src/control.ts
```diff
@@ -1,17 +1,260 @@
...
 
 export type AsyncFn<TArgs extends unknown[], TResult> = (...args: TArgs) => Promise<TResult>;
 
+// ---------------------------------------------------------------------------
+// Internal helpers
+// ---------------------------------------------------------------------------
+
+function extractInput(args: unknown[]): unknown {
+  if (args.length === 0) return null;
+  if (args.length === 1) return args[0];
```
lan17, line 56: I think this payload shape breaks tool calls. The server validates `tool` steps with `input` as an object, but this helper returns a primitive for one arg and an array for multiple args. That means wrappers like `lookup_customer(query: string)` or `search_knowledge_base(query: string)` will 422 during the pre-check and never execute. We probably want to build a named-argument object here, like the Python SDK does.

sdks/typescript/src/client.ts
```diff
@@ -26,18 +29,53 @@ export type ObservabilityApi = AgentControlSDK["observability"];
...
+
 export class AgentControlClient {
   private options: AgentControlInitOptions | null = null;
   private sdk: AgentControlSDK | null = null;
 
-  init(options: AgentControlInitOptions): void {
-    this.options = { ...options };
+  async init(options: AgentControlInitOptions): Promise<void> {
+    const agentName = validateAgentName(options.agentName);
+    this.options = { ...options, agentName };
```
lan17, line 57: This publishes initialized state before the async registration finishes. If callers keep using `init()` like the synchronous API on `main`, they can immediately run a controlled call against an agent that hasn’t been created yet, and a failed `agents.init()` still leaves `initialized === true`. I’d either delay assigning `sdk/options` until after registration succeeds or roll them back on error.

examples/customer_support_agent_ts/src/setup-controls.ts
```diff
@@ -0,0 +1,235 @@
...
+];
+
+async function main(): Promise<void> {
+  console.log(`Setting up controls on ${serverUrl} ...`);
+
+  const health = await client.system.healthCheck();
+  console.log(`Server: ${health.status} (${health.version})`);
+
+  // Register agent (steps are auto-registered when the agent app runs init() — same as Python)
+  await client.agents.init({
```
lan17, line 159: This re-registers the demo agent with only name/description. Because the SDK now defaults `initAgent` to `overwrite`, rerunning setup after the demo has already registered itself will wipe the stored step list and the `sdk_language` marker. That leaves the UI without step names and turns off the TS compatibility guard until the agent app runs again.

PR #70, opened 2026-03-06
feat(sdk): add shutdown() and ashutdown() lifecycle API

abhinav-galileo:
## Summary

Adds `agent_control.shutdown()` (sync) and `agent_control.ashutdown()` (async) for deterministic SDK teardown.

- Flushes all pending observability events
- Stops the policy refresh background thread and prevents in-flight refreshes from publishing after shutdown
- Resets global state
- Idempotent and safe to call without prior `init()`

This gives users a single, documented shutdown path instead of relying on best-effort `atexit` behavior - which is fragile for short-lived scripts, CLI tools, and demos.

### Implementation details

- Extracted `_fetch_controls_async()` (fetch-only, no publish) so the refresh worker can check the stop event between fetch and publish, preventing zombie state writes after shutdown.
- Added `sync_shutdown_observability()` as a public API in the observability module; `shutdown()` uses it instead of reaching into `_batcher` internals.
- Extracted `_reset_state()` helper to deduplicate state clearing across both shutdown paths.

Shortcut: [sc-57753](https://app.shortcut.com/galileo/story/57753)

## Test plan

- [x] `make sdk-lint`
- [x] `make sdk-typecheck`
- [x] `make sdk-test` (all passed, pre-existing integration failures only)
- [x] 10 new tests in `test_shutdown.py` covering sync/async paths, idempotency, no-init safety, and zombie refresh race regression
- [x] Updated `test_policy_refresh_loop.py` to match new `_fetch_controls_async` call path (14 tests pass)

### DIFF
--- a/sdks/python/src/agent_control/__init__.py
+++ b/sdks/python/src/agent_control/__init__.py
@@ -49,6 +49,7 @@ import asyncio
 import os
 import threading
 from collections.abc import Callable, Coroutine
+from dataclasses import dataclass
 from datetime import UTC, datetime
 from typing import Any, Literal, TypeVar
 
@@ -127,6 +128,20 @@ _policy_refresh_interval_seconds: int | None = None
 _refresh_thread: threading.Thread | None = None
 _refresh_stop_event: threading.Event | None = None
 
+# Session lifecycle state used to discard stale manual refresh results.
+_session_lock = threading.Lock()
+_session_generation = 0
+
+
+@dataclass(frozen=True)
+class _RefreshContext:
+    """Connection/session snapshot for a single refresh request."""
+
+    session_generation: int
+    agent_name: str
+    server_url: str
+    api_key: str | None
+
 
 def get_server_controls() -> list[dict[str, Any]] | None:
     """Get the cached server controls.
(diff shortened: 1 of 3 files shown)


sdks/python/src/agent_control/__init__.py
```diff
@@ -241,18 +276,11 @@ async def refresh_controls_async() -> list[dict[str, Any]] | None:
...
     try:
-        async with AgentControlClient(base_url=state.server_url, api_key=state.api_key) as client:
-            response = await agents.list_agent_controls(client, state.current_agent.agent_name)
-            refreshed_controls = _publish_server_controls(response.get("controls", []))
-            logger.info("Refreshed %d control(s) from server", len(refreshed_controls or []))
-            return refreshed_controls
+        controls = await _fetch_controls_async()
+        refreshed_controls = _publish_server_controls(controls)
+        logger.info("Refreshed %d control(s) from server", len(refreshed_controls or []))
+        return refreshed_controls
```
namrataghadi-galileo: `refresh_controls_async()` used to raise RuntimeError when `current_agent` or `server_url` was missing. It now calls _fetch_controls_async() inside a broad try/except and returns None on any exception. This silently changes behavior for callers who depended on the exception to detect misconfiguration.

sdks/python/src/agent_control/__init__.py
```diff
@@ -241,18 +276,11 @@ async def refresh_controls_async() -> list[dict[str, Any]] | None:
         controls = await agent_control.refresh_controls_async()
         print(f"Refreshed {len(controls)} controls")
     """
-    if state.current_agent is None:
-        raise RuntimeError("Agent not initialized. Call agent_control.init() first.")
-
-    if state.server_url is None:
-        raise RuntimeError("Server URL not set. Call agent_control.init() first.")
-
     try:
```
lan17, line 318: [P2] This broad `except` now swallows the `RuntimeError` from `_fetch_controls_async()` when the SDK is uninitialized. As a result, `refresh_controls_async()` and `refresh_controls()` silently return `None` or the previous cache before `init()` or after `shutdown()`, instead of surfacing the lifecycle misuse to the caller.

sdks/python/src/agent_control/__init__.py
```diff
@@ -241,18 +276,11 @@ async def refresh_controls_async() -> list[dict[str, Any]] | None:
...
-        async with AgentControlClient(base_url=state.server_url, api_key=state.api_key) as client:
-            response = await agents.list_agent_controls(client, state.current_agent.agent_name)
-            refreshed_controls = _publish_server_controls(response.get("controls", []))
-            logger.info("Refreshed %d control(s) from server", len(refreshed_controls or []))
-            return refreshed_controls
+        controls = await _fetch_controls_async()
```
lan17: [P2] This still publishes whatever `_fetch_controls_async()` returns without checking that the SDK session is still current. If a manual refresh is in flight while another task calls `shutdown()` or re-runs `init()`, the late response can repopulate `state.server_controls` after teardown or overwrite the new agent's controls with stale data from the old session.

PR #83, opened 2026-03-10
feat(sdk): add Google ADK framework integration

abhinav-galileo:
## Summary
- add a shared framework integrations core for Agent Control evaluation and enforcement
- add a packaged Google ADK integration with lazy imports, multi-agent step naming, bind-based registration, async fallback step sync, and cleanup support
- add SDK tests for the core and Google ADK integration surface

## Verification
- make check
- make sdk-lint
- make sdk-typecheck
- make sdk-test

### DIFF
--- a/sdks/python/src/agent_control/integrations/google_adk/_extractors.py
+++ b/sdks/python/src/agent_control/integrations/google_adk/_extractors.py
@@ -27,6 +27,11 @@ def _extract_text_from_parts(parts: Any) -> str:
             chunks.append(text)
             continue
 
+        structured = _extract_structured_part(part)
+        if structured is not None:
+            chunks.append(structured)
+            continue
+
         if isinstance(part, dict):
             dict_text = part.get("text")
             if isinstance(dict_text, str) and dict_text:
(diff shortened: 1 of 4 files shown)


sdks/python/src/agent_control/integrations/google_adk/_extractors.py
```diff
@@ -0,0 +1,137 @@
...
+        return ""
+
+    chunks: list[str] = []
+    for part in parts:
+        text = getattr(part, "text", None)
+        if isinstance(text, str) and text:
+            chunks.append(text)
+            continue
+
+        if isinstance(part, dict):
```
lan17, line 35: I think `_extract_text_from_parts` is only pulling `part.text` (plus dict `text/json`), but ADK parts can also contain structured payloads like function-call / function-response objects. In tool-calling flows that can collapse evaluated model output to `""`, so post-model controls may never see unsafe content. We should serialize structured part fields before evaluation so controls still run on non-text outputs.

sdks/python/src/agent_control/integrations/google_adk/plugin.py
```diff
@@ -0,0 +1,902 @@
...
+            return
+
+        result_like = {
+            "action": "steer" if isinstance(exc, agent_control.ControlSteerError) else "deny",
+            "message": exc.message,
+            "steering_context": (
+                exc.steering_context if isinstance(exc, agent_control.ControlSteerError) else None
+            ),
+        }
+        self.on_violation_callback(
```
lan17: We should probably treat `on_violation_callback` as best-effort. Right now if user callback code throws (telemetry/network/etc.), that exception can bubble out of `_handle_llm_exception` / `_handle_tool_exception` and prevent us from returning the blocked replacement response. That turns a handled deny/steer into a runtime failure; wrapping this callback in `try/except` with logging would keep enforcement behavior reliable.

namrataghadi-galileo (approved):

PR #85, opened 2026-03-10
fix(examples): shut down SDK in short-lived demos

abhinav-galileo:
Updated docs/examples for user-facing changes

### DIFF
--- a/examples/agent_control_demo/demo_agent.py
+++ b/examples/agent_control_demo/demo_agent.py
@@ -251,6 +251,15 @@ The controls are evaluated SERVER-SIDE:
 """)
 
 
+async def run_demo_session() -> None:
+    """Initialize the agent and run the scripted demo."""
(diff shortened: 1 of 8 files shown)


examples/agent_control_demo/demo_agent.py
```diff
@@ -158,90 +158,91 @@ async def run_demo():
     print(f"   Server: {SERVER_URL}")
 
     try:
-        logger.info(f"Initializing agent: {AGENT_NAME}")
-        agent_control.init(
-            agent_name=AGENT_NAME,
-            agent_description="Demo chatbot for testing controls",
-            server_url=SERVER_URL,
+        try:
+            logger.info(f"Initializing agent: {AGENT_NAME}")
```
lan17: suggest to extract this out so that try/finally is around a short amount of code.

PR #93, opened 2026-03-11
fix(ui): create flow - name trimming issue

siddhant-galileo:
## Summary
- What changed and why.
   Fix - https://galileoteamworkspace.slack.com/archives/C09TKBCRABG/p1773172188238399?thread_ts=1773165596.478009&cid=C09TKBCRABG
  Also login modal icon fixed

- Fixed logo color on login modal. 

## Scope
- User-facing/API changes:
- Internal changes:
- Out of scope:

## Risk and Rollout
- Rollback plan:

## Testing

## Checklist

### DIFF
--- a/ui/src/core/page-components/agent-detail/modals/edit-control/edit-control-content.tsx
+++ b/ui/src/core/page-components/agent-detail/modals/edit-control/edit-control-content.tsx
@@ -87,8 +87,7 @@ export const EditControlContent = ({
       execution: 'server',
     },
     validate: {
-      name: (value) =>
-        !value?.trim() ? 'Control name is required' : null,
+      name: (value) => (!value?.trim() ? 'Control name is required' : null),
       selector_path: (value) => {
         if (!value?.trim()) {
           return 'Selector path is required';
@@ -290,8 +289,7 @@ export const EditControlContent = ({
             color: 'green',
           });
         } else {
-          const nameChanged =
-            values.name.trim() !== control.name.trim();
+          const nameChanged = values.name.trim() !== control.name.trim();
 
           if (nameChanged) {
             try {
@@ -323,13 +321,18 @@ export const EditControlContent = ({
                       evaluatorForm
                     );
                     setUnmappedErrors(
-                      unmapped.map((e) => ({ field: e.field, message: e.message }))
+                      unmapped.map((e) => ({
+                        field: e.field,
+                        message: e.message,
+                      }))
                     );
                   }
                 } else {
                   notifications.show({
                     title: 'Failed to rename control',
-                    message: problemDetail.detail || 'An unexpected error occurred while renaming',
+                    message:
+                      problemDetail.detail ||
+                      'An unexpected error occurred while renaming',
                     color: 'red',
                   });
                 }
--- a/models/src/agent_control_models/server.py
+++ b/models/src/agent_control_models/server.py
@@ -10,7 +10,7 @@ from .policy import Control
 
 
 def _strip_slug_name(v: str) -> str:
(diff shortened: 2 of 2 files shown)


ui/src/core/page-components/agent-detail/modals/edit-control/edit-control-content.tsx
```diff
@@ -270,31 +288,39 @@ export const EditControlContent = ({
...
           notifications.show({
             title: 'Control created',
-            message: `"${values.name}" has been added to this agent.`,
+            message: `"${normalizedName}" has been added to this agent.`,
             color: 'green',
           });
         } else {
-          const trimmedName = values.name.trim();
-          const nameChanged = trimmedName && trimmedName !== control.name;
+          const nameChanged = normalizedName && normalizedName !== control.name;
```
lan17: The validator above explicitly allows an existing non-slug name as long as the user leaves it unchanged, but this branch compares the trimmed value against the raw stored name. For a legacy control stored as `" bad name "`, saving only the description now triggers a rename to `"bad name"`; the new PATCH schema rejects that with `422`, and the control data update never runs. That breaks the intended legacy-edit path for any old records with leading/trailing whitespace.

ui/src/core/components/login-modal.tsx
```diff
@@ -67,8 +67,8 @@ export function LoginModal({ opened }: LoginModalProps) {
             <Image
               src={
                 colorScheme === 'dark'
```
lan17: This conditional is reversed. `ac-logo-dark.svg` is the white-on-dark variant already used elsewhere in the app, so dark mode now renders the black logo on the dark login modal and light mode renders the white logo on the light modal. The flow still works, but the branding becomes hard to see in the matching theme.

ui/src/core/page-components/agent-detail/modals/edit-control/edit-control-content.tsx
siddhant-galileo: fixed - https://github.com/agentcontrol/agent-control/pull/93/commits/9ce4c72d81f28f83cd0857816c33d9ff58828f87

ui/src/core/components/login-modal.tsx
siddhant-galileo: fixed - https://github.com/agentcontrol/agent-control/pull/93/commits/9ce4c72d81f28f83cd0857816c33d9ff58828f87

We don't need colorScheme here actually. Login modal exists outside app layout. so we dont have provision for light/dark mode here. kept it simple to be dark mode

siddhant-galileo:
> I think the main unresolved issue here is the contract boundary.
> 
> This repo already has a clear pattern for agent names: normalization is defined once and applied consistently across shared models, server endpoints, and SDK entry points. Control names still do not have that. In this PR, the only canonicalization being added is inside one UI flow, while the server create/patch endpoints and both SDKs still treat `name` as caller-provided input.
> 
> That matters because non-UI clients are real clients here. A browser user now gets trimming/canonicalization in the modal, but Python and TypeScript SDK callers still need to know to pre-trim or they will just hit API validation. I think we should make that choice explicit: either control-name canonicalization belongs at the server/shared boundary, or this remains validation-only and every client is responsible for sending canonical names. I would avoid implicitly defining canonical control names in one modal.
> 
> Separately, the legacy-name policy should be explicit and covered by a regression test. The code reads like “existing invalid names remain editable unless the user actually renames them”, which seems reasonable, but it should be stated and tested directly.

Fixed this - moved logic to server now. UI only uses trim to check if name has updated or not


> The login logo swap also looks unrelated to the trimming fix. If it’s intentional, I’d explain it in the PR; otherwise I’d split it out.

Added a line in PR description

PR #100, opened 2026-03-11
feat(docs): simplify main README quick start v2

abhinav-galileo:


README.md
```diff
@@ -22,282 +22,144 @@
...
----
-
-**Alternatively**, for local development with the Agent Control repository, clone the repo and follow all steps below.
+### 4. Add controls
 
-### Step 1: Start the Agent Control server
+Choose the path you want:
 
-Startup Agent Control server manually for local development.
+- Visual setup: follow the [UI Quickstart](https://docs.agentcontrol.dev/core/ui-quickstart)
```
namrataghadi-galileo: I still think here, we should add link to main Quick Start as it has link to UI-Quick start too. 

`- Working code examples: browse [examples/README.md](examples/README.md)` feels misplaced in this section

abhinav-galileo: - added main quick start
- working "code" examples is fine, no?

namrataghadi-galileo: It just feels misplaced in the Add Controls section.
If we can add it outside this section following right after, that will be great

README.md
```diff
@@ -22,172 +22,81 @@
...
-## Quick start
+- **Centralized safety** - define controls once, apply across agents, update without redeploying
+- **Runtime configuration** - manage controls via API or UI, no code changes needed
+- **Pluggable evaluators** - built-in (regex, list, JSON, SQL) or bring your own
+- **Framework support** - works with LangChain, CrewAI, Google ADK, AWS Strands, and more
 
-Protect your AI agent in 4 simple steps.
+![Agent Control Overview](docs/images/AgentControlDiagram.png)
 
+## Quick Start
```
lan17, line 42: What about the docker compose start if you just wanna run agent control server ?

README.md
```diff
@@ -22,172 +22,81 @@
...
+Prerequisites: Docker and Python 3.12+.
 
-- Python 3.12+
-- Docker
+The fastest way to try Agent Control is:
 
----
+1. Start the server
+2. Install the SDK
+3. Wrap a model or tool call with `@control()`
```
lan17: We should point out that this is only needed when integration is not available.

README.md
namrataghadi-galileo: it is still there in readme

PR #116, opened 2026-03-12
fix(sdk): support async generator functions in control() decorator

### LINKED ISSUE #113: control() decorator can break on streaming (async generator) functions for agents that have stream=true for latency.
### Summary
@agent_control.control() silently breaks when applied to streaming (async generator) functions — the standard pattern for LLM response streaming. Controls are either completely bypassed or the decorator crashes at runtime.

### Motivation
Most LLM applications stream responses via async generators. If @control() can't protect these, output controls (toxicity, PII, content policy) are ineffective for the majority of real-world usage. Harmful content streams directly to users unevaluated.

### Current behavior
Async generator → crash: Decorator misclassifies it as sync, calls asyncio.run() inside a running loop RuntimeError
Async generator → identity lost: inspect.isasyncgenfunction(decorated) returns False, breaking framework introspection
Returns async iterator → garbage evaluation: Post-check evaluates "<async_generator object at 0x...>" instead of actual content

### Expected behavior
Async generators remain async generators after decoration
Pre-check runs before first chunk; chunks yield in real-time
Post-check runs after stream completes on full accumulated output
Deny/steer actions raise appropriate errors after stream ends

### Reproduction
Apply @agent_control.control() to an async def function with yield
Call it from async context
Get RuntimeError: asyncio.run() cannot be called from a running event loop
Existing tests: TestStreamingLimitations in sdks/python/tests/test_control_decorators.py

### Proposed solution
Add an async generator wrapper path in control():

if inspect.isasyncgenfunction(func):
    return async_gen_wrapper   # new path
elif inspect.iscoroutinefunction(func):
    return async_wrapper
return sync_wrapper
Where async_gen_wrapper runs pre-check, yields chunks while accumulating, then runs post-check on the joined output.

KazChe:
## Summary
- adds async generator (streaming) support to the `control()` decorator, fixing crashes and silent bypasses when applied to `async def` functions with `yield` — the standard pattern for LLM response streaming.

## Scope
- User-facing/API changes: `@control()` now correctly wraps async generator functions, preserving `inspect.isasyncgenfunction()` identity
- Internal changes: New `async_gen_wrapper` path in `control()` that runs pre-check before first chunk, yields chunks in real-time while accumulating, and runs post-check on full accumulated output after stream completes
- Out of scope: Sync generator support, per-chunk evaluation

## Risk and Rollout
- Risk level: low
- Rollback plan: Revert commit — the change is additive (new code path) with no modifications to existing sync/async wrapper logic

## Testing
- [x] Ran `make check` (lint + typecheck pass; server tests have pre-existing postgres failures unrelated to this change)

## Checklist
- [x] Linked issue/spec (if applicable) — Closes #113
- [x] Included any required follow-up tasks

### DIFF
--- a/sdks/python/src/agent_control/control_decorators.py
+++ b/sdks/python/src/agent_control/control_decorators.py
@@ -829,11 +829,63 @@ def control(policy: str | None = None, step_name: str | None = None) -> Callable
                     pass
 
         @functools.wraps(func)
+        async def async_gen_wrapper(*args: Any, **kwargs: Any) -> Any:
+            agent = _get_current_agent()
+            if agent is None:
+                logger.warning(
+                    "No agent initialized. Call agent_control.init() first. "
+                    "Running without protection."
+                )
+                async for chunk in func(*args, **kwargs):
+                    yield chunk
+                return
+
+            controls = _get_server_controls()
+
+            existing_trace_id = get_current_trace_id()
+            if existing_trace_id:
+                trace_id = existing_trace_id
+                span_id = _generate_span_id()
+            else:
+                trace_id, span_id = get_trace_and_span_ids()
+
+            ctx = ControlContext(
+                agent_name=agent.agent_name,
+                server_url=_get_server_url(),
+                func=func,
+                args=args,
+                kwargs=kwargs,
+                trace_id=trace_id,
+                span_id=span_id,
+                start_time=time.perf_counter(),
+                step_name=step_name,
+            )
+            ctx.log_start()
+
+            try:
+                # PRE-EXECUTION: Check controls with check_stage="pre"
+                await _run_control_check(ctx, "pre", ctx.pre_payload(), controls)
+
+                # Yield chunks while accumulating full output for post-check
+                accumulated: list[str] = []
+                async for chunk in func(*args, **kwargs):
+                    accumulated.append(str(chunk))
+                    yield chunk
+
+                # POST-EXECUTION: Check controls on full accumulated output
+                full_output = "".join(accumulated)
+                await _run_control_check(ctx, "post", ctx.post_payload(full_output), controls)
+            finally:
+                ctx.log_end()
+
+        @functools.wraps(func)
         def sync_wrapper(*args: Any, **kwargs: Any) -> Any:
             return asyncio.run(
                 _execute_with_control(func, args, kwargs, is_async=False, step_name=step_name)
             )
 
+        if inspect.isasyncgenfunction(func):
+            return async_gen_wrapper  # type: ignore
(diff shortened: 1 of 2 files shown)


sdks/python/src/agent_control/control_decorators.py
```diff
@@ -828,12 +828,64 @@ async def async_wrapper(*args: Any, **kwargs: Any) -> Any:
...
+            ctx.log_start()
+
+            try:
+                # PRE-EXECUTION: Check controls with check_stage="pre"
+                await _run_control_check(ctx, "pre", ctx.pre_payload(), controls)
+
+                # Yield chunks while accumulating full output for post-check
+                accumulated: list[str] = []
+                async for chunk in func(*args, **kwargs):
+                    accumulated.append(str(chunk))
```
abhinav-galileo, line 872: `str(chunk)` is lossy - if chunks are dicts, dataclasses, or framework message objects, the original data can't be recovered for the post-check. Can we handle non-string chunks here?

sdks/python/src/agent_control/control_decorators.py
```diff
@@ -828,12 +828,64 @@ async def async_wrapper(*args: Any, **kwargs: Any) -> Any:
...
+                start_time=time.perf_counter(),
+                step_name=step_name,
+            )
+            ctx.log_start()
+
+            try:
+                # PRE-EXECUTION: Check controls with check_stage="pre"
+                await _run_control_check(ctx, "pre", ctx.pre_payload(), controls)
+
+                # Yield chunks while accumulating full output for post-check
```
namrataghadi-galileo, line 869: This path only runs the post-stage check after the async for finishes. If the caller breaks early, cancels the task, or the client disconnects mid-stream, Python closes the wrapper and runs finally, but never reaches the post-check. That leaves a real bypass for streaming consumers, which is exactly where partial reads are common. This path needs explicit handling/documentation and tests for break, cancellation, and aclose().

sdks/python/src/agent_control/control_decorators.py
```diff
@@ -828,12 +828,64 @@ async def async_wrapper(*args: Any, **kwargs: Any) -> Any:
...
+                accumulated: list[str] = []
+                async for chunk in func(*args, **kwargs):
```
namrataghadi-galileo, line 871: Unlike the existing non-streaming wrapper, this implementation yields every chunk to the caller before enforcing the post-stage result. A post-stage deny or steer therefore raises only after the full response has already been delivered, so it no longer provides the same fail-closed behavior as control() on normal async functions. If that tradeoff is intentional, it needs to be called out very clearly; otherwise the stream must be buffered or evaluated chunk-by-chunk.

Issue #120, opened 2026-03-12
ListEvaluator: values=[""] builds match-all regex, can cause complete agent denial-of-service. Can be handled from UI/Backend.


Ritvik777:
### Summary
When you set values=[""] (a list with one empty string) in the list evaluator config, it accidentally blocks every single input. Any message — "Book a flight", "Tell me a joke", anything — gets denied. One typo in the config = your entire agent stops working. ested end-to-end on a running server (FastAPI + PostgreSQL). The server accepts values=[""] without complaint and stores it in the database. The bug triggers at evaluation time.
The existing guard (if not self._values) catches an empty list [] but misses a list containing an empty string [""] — because [""] has length 1, so it looks non-empty.

### Motivation
This is an open-source, self-hosted project. Anyone can configure controls through the API or a UI form. It's really easy to accidentally end up with an empty string in the values list — a blank form field, a trailing comma, or just "" instead of []. When that happens, there's no error, no warning. The control silently starts blocking everything and you have no idea why your agent stopped responding.

### Current behavior
You set values=[""] in a list control config.
The server accepts it — no validation error.
Under the hood, the evaluator builds a regex from that empty string: \b()\b.
That regex matches at every word boundary in every string — so it matches everything.
Every evaluation returns matched: true, every request gets denied.
The agent is completely shut down.

### Expected behavior
Empty strings should be filtered out from the values list before the regex is built.
values=[""] should behave the same as values=[] — the evaluator should say "Empty control values - control ignored" and return matched: false.

### Reproduction
Create a control: PUT /api/v1/controls → PUT /api/v1/controls/{id}/data with values=[""], match_mode="contains", action.decision="deny".
Attach it to an agent through a policy.
Send any message to POST /api/v1/evaluation.
Result: Every message is denied. 5 out of 5 normal user requests blocked.
Minimal reproduction (3 lines):

from agent_control_evaluators.list.evaluator import ListEvaluator
from agent_control_evaluators.list.config import ListEvaluatorConfig
import asyncio
evaluator = ListEvaluator(ListEvaluatorConfig(values=[""], match_mode="contains"))
result = asyncio.run(evaluator.evaluate("Tell me a joke"))
print(result.matched)  # True — should be False

### Proposed solution
One-line fix in evaluator.py line 38 — filter out empty strings:

### # Before:
self._values = [str(v) for v in config.values]
### # After:
self._values = [str(v) for v in config.values if str(v).strip()]

### Additional context
[...]

PR #121, opened 2026-03-12
fix(server): reject empty string list evaluator values

lan17:
## Summary
- reject empty-string list entries at list evaluator config validation time
- defensively filter exact empty strings in the evaluator so legacy invalid configs do not compile into a match-all regex
- add evaluator and server regression tests for both validation paths

## Testing
- make evaluators-test
- make server-test
- make lint
- make typecheck

Fixes https://github.com/agentcontrol/agent-control/issues/120

### DIFF
--- a/evaluators/builtin/src/agent_control_evaluators/list/config.py
+++ b/evaluators/builtin/src/agent_control_evaluators/list/config.py
@@ -28,7 +28,7 @@ class ListEvaluatorConfig(EvaluatorConfig):
     @field_validator("values")
     @classmethod
     def validate_values(cls, values: list[str | int | float]) -> list[str | int | float]:
-        """Reject empty-string entries that would compile into a match-all regex."""
-        if any(isinstance(value, str) and value == "" for value in values):
-            raise ValueError("values must not contain empty strings")
+        """Reject blank string entries that would compile into pathological regexes."""
+        if any(isinstance(value, str) and value.strip() == "" for value in values):
+            raise ValueError("values must not contain empty or whitespace-only strings")
         return values
--- a/evaluators/builtin/tests/list/test_list.py
+++ b/evaluators/builtin/tests/list/test_list.py
@@ -13,8 +13,20 @@ class TestListEvaluatorConfig:
         """Test that empty-string list entries are rejected at config validation time."""
         # Given: a list evaluator config with an empty-string value
         # When: constructing the config model
-        with pytest.raises(ValidationError, match="values must not contain empty strings"):
+        with pytest.raises(
+            ValidationError, match="values must not contain empty or whitespace-only strings"
+        ):
             ListEvaluatorConfig(values=[""])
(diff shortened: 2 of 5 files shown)


evaluators/builtin/src/agent_control_evaluators/list/config.py
```diff
@@ -24,3 +24,11 @@ class ListEvaluatorConfig(EvaluatorConfig):
         description="'exact' for full string match, 'contains' for keyword/substring match",
     )
     case_sensitive: bool = Field(False, description="Whether matching is case sensitive")
```
abhinav-galileo, line 26: This catches exact `""` but allows whitespace-only strings like `" "` or `"\t"` through. Those produce similarly pathological regexes - e.g. `\b(\ )\b` in contains mode.

Consider using `strip()`:
```python
if any(isinstance(value, str) and value.strip() == "" for value in values):
    raise ValueError("values must not contain empty or whitespace-only strings")
```

Same change needed in the defensive filter in `evaluator.py:38` (`if value != ""` -> `if value.strip()`).

evaluators/builtin/tests/list/test_list.py
```diff
@@ -0,0 +1,42 @@
...
+            case_sensitive=False,
+        )
+        evaluator = ListEvaluator(config)
+
+        # When: evaluating normal text against the legacy config
+        result = await evaluator.evaluate("Tell me a joke")
+
+        # Then: the evaluator ignores the empty control values
+        assert result.matched is False
+        assert result.message == "Empty control values - control ignored"
```
abhinav-galileo, line 54: This tests the denylist path (`match_on="match"`), but the original report was about an allowlist blocking all calls. Could we add a companion case with `match_on="no_match"` that asserts `matched is False` for normal input? That encodes the failure mode that prompted this fix.

Something like:
```python
async def test_legacy_empty_string_allowlist_does_not_block_all(self) -> None:
    config = ListEvaluatorConfig.model_construct(
        values=[""],
        logic="any",
        match_on="no_match",
        match_mode="exact",
        case_sensitive=False,
    )
    evaluator = ListEvaluator(config)
    result = await evaluator.evaluate("legitimate_value")
    assert result.matched is False
```

lan17: Addressed in `319eede`.

I added a companion legacy allowlist regression covering `match_on="no_match"`, plus a whitespace-only legacy case. The evaluator tests now cover both the original failure mode and the broadened blank-string hardening.

evaluators/builtin/src/agent_control_evaluators/list/config.py
lan17: Addressed in `319eede`.

I broadened the guard from exact `""` to any blank/whitespace-only string in both the config validator and the evaluator fallback. That now rejects values like `" "` and `"\t"` at validation time and defensively drops them for legacy invalid configs.

Issue #123, opened 2026-03-12
Skip evaluation entirely when no controls apply to the current step


lan17:
## Summary
When an agent step has no applicable controls, evaluation should be a no-op. Today the SDK still enters the evaluation path, and if the agent has any server-executed controls at all, it will still call `/api/v1/evaluation` even when no control is defined for the current tool/step or the relevant controls are disabled.

## Motivation
This affects the common case where only a subset of tools or stages are protected. Teams expect unprotected steps to behave like plain function calls with no guardrail overhead. Making unnecessary evaluation calls adds avoidable latency, creates noise in traces/logs, and makes it harder to reason about whether a step is actually protected.

## Current behavior
In the SDK local-evaluation path, controls are first partitioned by `execution`. For server controls, the SDK sets `has_server_controls = true` based on the presence of any `execution="server"` control for the agent, before checking whether that control is:
- enabled
- scoped to the current stage
- scoped to the current step type
- scoped to the current step name or regex

As a result:
- disabled or non-matching local controls are skipped by the engine and do not reach evaluator execution
- but the SDK still makes a server `/api/v1/evaluation` call whenever any server control exists for the agent, even if none apply to the current step
- the server then repeats applicability filtering and often returns a result with no applicable controls

This means “no control for this tool/step” is not a true no-op today.

## Expected behavior
If there are no applicable controls for the current step, evaluation should be skipped entirely.

Concretely:
- if no enabled control applies to the current tool/step/stage, the SDK should not run local evaluation
- if no enabled server control applies to the current tool/step/stage, the SDK should not call `/api/v1/evaluation`
- disabled controls and controls scoped to different tools/steps should behave as if they are absent for that invocation

## Reproduction (if bug)
1. Create an agent with a server control scoped to a specific tool or step, for example `step_names=["send_email"]`, or a control that is disabled.
2. Initialize the SDK so those controls are cached locally.
3. Invoke a different `@control()`-wrapped tool or step, for example `lookup_order`, at a stage where no control applies.
4. Observe that the SDK still enters evaluation and may call `POST /api/v1/evaluation`, even though no control applies to that invocation.

## Proposed solution (optional)
Add an SDK-side applicability prefilter before local or server evaluation is triggered. The prefilter should mirror the existing engine applicability checks:
- `enabled`
- `scope.stages`
- `scope.step_types`
- `scope.step_names`
- `scope.step_name_regex`
- `execution`

Only controls that pass that applicability check for the current invocation should:
[...]

PR #124, opened 2026-03-13
fix(sdk): skip evaluation when no controls apply

lan17:
## Summary
- Make the SDK evaluation path a true no-op when no enabled control applies to the current step/stage.
- Prefilter both local (`execution="sdk"`) and server (`execution="server"`) controls using the same applicability rules before any evaluation work starts.
- Preserve the existing fail-open behavior for malformed server controls by still deferring those to the server.
- Add regression coverage for disabled and non-matching controls across both local and server execution.

## Scope
- User-facing/API changes:
  - If no control applies to a step invocation, the SDK does nothing: no local control engine execution and no `POST /api/v1/evaluation` request.
- Internal changes:
  - Reuse the SDK control engine's applicability logic to determine whether local or server evaluation should run at all.
  - Extend local-evaluation tests to cover disabled, stage-mismatched, step-type-mismatched, step-name-mismatched, and regex-mismatched controls for both execution modes.
- Out of scope:
  - Server-side evaluation behavior when applicable controls do exist.
  - Broader control loading or caching changes.

## Risk and Rollout
- Risk level: low
- Rollback plan:
  - Revert this PR to restore the previous behavior where the SDK entered evaluation as long as local or server controls existed for the agent.

## Testing
  - Ran `make test`, plus targeted `ruff` and `mypy` on the touched SDK files.

## Checklist
- [x] Linked issue/spec (if applicable)
  - Closes https://github.com/agentcontrol/agent-control/issues/123
  - Not needed; behavior change is covered by tests.
  - None.

### DIFF
--- a/sdks/python/src/agent_control/evaluation.py
+++ b/sdks/python/src/agent_control/evaluation.py
@@ -148,21 +148,25 @@ def _get_applicable_controls(
     return cast(list[_ControlAdapter], applicable_controls)
 
 
-def _has_applicable_server_controls(
-    controls: list[dict[str, Any]],
+def _has_applicable_prefiltered_server_controls(
+    server_control_payloads: list[dict[str, Any]],
     request: EvaluationRequest,
 ) -> bool:
-    """Return whether any well-formed server control applies to this request.
+    """Return whether any partitioned server control applies to this request.
 
-    If any server control cannot be parsed locally, this returns True so the SDK
-    still defers to the server for authoritative handling.
+    The caller is responsible for partitioning raw control payloads by
+    ``execution`` before calling this helper. This function only inspects the
+    server-control subset and does not re-check ``execution`` itself.
+
+    If any server control payload cannot be parsed locally, this returns True so
+    the SDK still defers to the server for authoritative handling.
     """
-    server_controls: list[_ControlAdapter] = []
+    parsed_server_controls: list[_ControlAdapter] = []
 
-    for control in controls:
+    for control in server_control_payloads:
         try:
             control_def = ControlDefinition.model_validate(control["control"])
-            server_controls.append(
+            parsed_server_controls.append(
                 _ControlAdapter(
                     id=control["id"],
                     name=control["name"],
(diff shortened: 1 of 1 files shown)


sdks/python/src/agent_control/evaluation.py
```diff
@@ -134,6 +134,51 @@ class _ControlAdapter:
...
+    request: EvaluationRequest,
+) -> bool:
+    """Return whether any well-formed server control applies to this request.
+
+    If any server control cannot be parsed locally, this returns True so the SDK
+    still defers to the server for authoritative handling.
+    """
+    server_controls: list[_ControlAdapter] = []
+
+    for control in controls:
```
namrataghadi-galileo:
```
for control in controls:
        control_data = control.get("control")
        if not isinstance(control_data, dict):
            _logger.warning(f"Malformed control payload: {control}")
            return True

        if control_data.get("execution") != "server":
            continue

        try:
            control_def = ControlDefinition.model_validate(control_data)
            server_controls.append(
                _ControlAdapter(
                    id=control["id"],
                    name=control["name"],
                    control=control_def,
                )
            )
        except Exception:
            _logger.warning(f"Failed to parse server control: {control}")
            return True

    if not server_controls:
        return False

    return bool(_get_applicable_controls(server_controls, request, context="server"))
```

namrataghadi-galileo: NOTE: Adding `server` check, to ensure we are looking at server controls. If the intention is to look at both `sdk` as well as `server` controls, then we should rename this function appropriately.
```
if control_data.get("execution") != "server":
            continue
```

namrataghadi-galileo: or atleast call out the details in the DocString

lan17: Good call. The helper only ever receives the server subset after the execution partitioning in `check_evaluation_with_local()`, so I did not want to add a second `execution == "server"` check inside it and make the contract even less explicit.\n\nI pushed a small follow-up in 91f4596 that makes that assumption clearer instead: renamed the helper to `_has_applicable_prefiltered_server_controls(...)`, renamed the input to `server_control_payloads`, and expanded the docstring to say that the caller has already partitioned by `execution` before calling it.\n\nSo behavior is unchanged, but the boundary should be much less surprising now.

Issue #118, opened 2026-03-12
Feature: Longitudinal behavioral drift evaluator (temporal policy compliance)


nanookclaw:
## Context

Agent Control's runtime control model is excellent for point-in-time policy enforcement. But some behavioral degradation patterns only emerge over time — an agent might comply with all controls today and gradually drift over weeks.

## The Problem

Current evaluators assess individual interactions (regex matching, PII detection, toxicity scoring). They answer: "Is this response safe right now?" But they don't answer: "Is this agent becoming less reliable over time?"

In our empirical work on behavioral measurement across 13 LLM agents (calibration, adaptation, robustness dimensions), we observed:
- Agents scoring 1.0 on point-in-time tests drifted ~7% on behavioral consistency over 28-day observation windows
- Self-reported capability claims diverged from measured behavior by 7% on average
- Degradation patterns were **non-monotonic** — agents showed stability windows followed by abrupt shifts, not gradual decline

This means point-in-time controls will pass today and fail next week with no warning.

## Proposed Solution

A **temporal behavioral drift evaluator** — a custom evaluator that:

1. **Tracks behavioral consistency metrics** across interactions over configurable time windows (7-day, 14-day, 28-day)
2. **Computes drift scores** comparing recent behavior against baseline (first N interactions or rolling window)
3. **Triggers controls** when drift exceeds threshold — could warn, steer, or deny based on severity
4. **Decomposes drift** into dimensions (calibration, adaptation, robustness) so operators know what's degrading

### Integration with Agent Control Architecture

This fits naturally as a pluggable evaluator:

```python
from agent_control import control

@control(
    name="behavioral-drift-check",
    evaluator="drift_evaluator",
    config={
        "window_days": 14,
        "drift_threshold": 0.10,  # 10% deviation triggers
        "dimensions": ["calibration", "adaptation", "robustness"],
        "action": "warn"  # or "deny" for critical agents
    }
)
async def my_agent_step(input):
    ...
```

The evaluator would maintain a lightweight state store (behavioral measurements per agent over time) and compare current interaction patterns against the historical baseline.

## Why This Matters for Enterprise

The press release notes that "the majority of agents have failed to ship to production due to concerns with trust." Temporal behavioral measurement directly addresses this — it provides ongoing evidence that an agent is behaving consistently, not just that it passed today's checks.

## References

- PDR paper (Zenodo record / DOI): https://zenodo.org/records/19028012
[...]

PR #140, opened 2026-03-20
feat(contrib): add drift.temporal evaluator for longitudinal behavioral monitoring

nanookclaw:
## Summary

Adds `drift.temporal` — a new contrib evaluator that detects gradual behavioral degradation patterns that point-in-time evaluators (regex, list, SQL, JSON) miss.

This follows from the discussion in #118, where @lan17 suggested implementing the evaluator in a standalone repository first. I built it ([nanookclaw/agent-control-drift-evaluator](https://github.com/nanookclaw/agent-control-drift-evaluator), 2 ⭐) and am now integrating it into the contrib ecosystem using the galileo pattern.

## The Gap

Built-in evaluators answer: *"Is this response OK right now?"*

They don't answer: *"Is this agent getting worse over time?"*

This evaluator fills that gap.

## How It Works

1. Records a numeric score (0.0–1.0) per agent per interaction
2. Compares the **recent window** (last N observations) to a **baseline** (first M observations)
3. Returns `matched=True` when recent average drops below baseline by more than the configured threshold
4. Stores history as local JSON — **no external API or service required**

## Usage

```yaml
controls:
  - name: "drift-check"
    selector: "$.quality_score"
    evaluator: "drift.temporal"
    config:
      agent_id: "sales-agent-prod"
      window_size: 10
      baseline_size: 20
      drift_threshold: 0.10
    action: alert
```

## Design Decisions (Empirically Grounded)

Two findings from [published longitudinal research (DOI: 10.5281/zenodo.19028012)](https://doi.org/10.5281/zenodo.19028012) shaped the defaults:

1. **`min_observations ≥ 5`** — Drift signals are noisy below 5 observations. Default prevents early false positives.
2. **Windowed comparison, not cumulative** — Agents can drift and recover without intervention. A rolling window captures current state; a cumulative average masks it.

Both patterns were independently replicated by a production deployment (NexusGuard fleet, v0.5.36, 48 passing tests).

## Package Structure

Follows the galileo contrib pattern exactly:

```
evaluators/contrib/drift/
├── pyproject.toml           # agent-control-evaluator-drift
├── Makefile                 # test / lint / typecheck / build
├── README.md
└── src/
    └── agent_control_evaluator_drift/
        └── drift/
            ├── config.py    # DriftEvaluatorConfig (Pydantic)
            └── evaluator.py # DriftEvaluator (@register_evaluator)
```

## Tests

31 tests covering:
- Config validation (bounds, window vs baseline, `on_error` modes)
- Core drift computation (insufficient data, baseline building, stable, drift detected, threshold boundaries)
- File I/O helpers (load/save roundtrip, missing file, corrupt JSON, auto-created directories)
- Full evaluator integration (persistence across instances, independent `agent_id` tracking, score clamping, fail-open/closed, metadata completeness)

## Checklist

- [x] Tests pass (verified locally against the builtin evaluator interface)
- [x] No external API required (`requires_api_key = False`)
- [x] Follows contrib package structure (galileo pattern)
- [x] `pyproject.toml` entry point registered: `drift.temporal = agent_control_evaluator_drift.drift:DriftEvaluator`
- [x] README with config reference, research background, and usage examples
- [x] `on_error` fail-open/closed behavior consistent with galileo evaluator

Relates to: #118

### DIFF
--- a/evaluators/contrib/drift/pyproject.toml
+++ b/evaluators/contrib/drift/pyproject.toml
@@ -1,14 +1,14 @@
 [project]
 name = "agent-control-evaluator-drift"
-version = "1.0.0"
+version = "8.0.0"
 description = "Temporal behavioral drift evaluator for agent-control"
 readme = "README.md"
 requires-python = ">=3.12"
 license = { text = "Apache-2.0" }
 authors = [{ name = "Nanook (nanookclaw)" }]
 dependencies = [
-    "agent-control-evaluators>=3.0.0",
-    "agent-control-models>=3.0.0",
+    "agent-control-evaluators>=7.5.0",
+    "agent-control-models>=7.5.0",
     "pydantic>=2.12.4",
 ]
 
--- a/evaluators/contrib/drift/pyproject.toml
+++ b/evaluators/contrib/drift/pyproject.toml
@@ -1,0 +1,37 @@
+[project]
+name = "agent-control-evaluator-drift"
+version = "1.0.0"
+description = "Temporal behavioral drift evaluator for agent-control"
+readme = "README.md"
+requires-python = ">=3.12"
+license = { text = "Apache-2.0" }
+authors = [{ name = "Nanook (nanookclaw)" }]
+dependencies = [
+    "agent-control-evaluators>=3.0.0",
+    "agent-control-models>=3.0.0",
+    "pydantic>=2.12.4",
+]
+
+[project.optional-dependencies]
+dev = [
+    "pytest>=8.0.0",
+    "pytest-asyncio>=0.23.0",
+    "pytest-cov>=4.0.0",
+    "ruff>=0.1.0",
+    "mypy>=1.8.0",
+]
+
+[project.entry-points."agent_control.evaluators"]
+"drift.temporal" = "agent_control_evaluator_drift.drift:DriftEvaluator"
+
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+
+[tool.hatch.build.targets.wheel]
+packages = ["src/agent_control_evaluator_drift"]
+
+# For local dev, use override to resolve from workspace
+[tool.uv.sources]
+agent-control-evaluators = { path = "../../builtin", editable = true }
+agent-control-models = { path = "../../../models", editable = true }
--- a/evaluators/contrib/drift/src/agent_control_evaluator_drift/drift/evaluator.py
+++ b/evaluators/contrib/drift/src/agent_control_evaluator_drift/drift/evaluator.py
@@ -1,0 +1,286 @@
+"""Temporal behavioral drift evaluator for Agent Control.
+
+Detects gradual behavioral degradation that point-in-time evaluators miss
+by tracking numeric scores over time and comparing recent windows to baselines.
+"""
+
+from __future__ import annotations
+
+import json
+import logging
+import os
+from pathlib import Path
+from typing import Any
+
+from agent_control_evaluators import Evaluator, EvaluatorMetadata, register_evaluator
+from agent_control_models import EvaluatorResult
+
+from agent_control_evaluator_drift.drift.config import DriftEvaluatorConfig
+
+logger = logging.getLogger(__name__)
+
+
+def _load_history(path: Path) -> list[float]:
+    """Load observation history from a JSON file.
+
+    Args:
+        path: Path to the history file.
+
+    Returns:
+        List of float scores, or empty list if file doesn't exist.
+    """
+    if not path.exists():
+        return []
+    try:
+        with path.open("r") as fh:
+            data = json.load(fh)
+            scores = data.get("scores", [])
+            return [float(s) for s in scores if isinstance(s, (int, float))]
+    except (json.JSONDecodeError, OSError, ValueError) as exc:
+        logger.warning("Failed to load drift history from %s: %s", path, exc)
+        return []
+
+
+def _save_history(path: Path, scores: list[float]) -> None:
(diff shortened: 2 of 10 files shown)


lan17 (commented):
Thanks for putting this together. I like the direction overall, and the package structure is easy to follow. I left a few comments on things I think we should tighten up before merging.

evaluators/contrib/drift/src/agent_control_evaluator_drift/drift/evaluator.py
```diff
@@ -0,0 +1,286 @@
...
+        }
+
+    baseline_scores = scores[:baseline_size]
+    recent_scores = scores[-window_size:]
+
+    baseline_avg = sum(baseline_scores) / len(baseline_scores)
+    recent_avg = sum(recent_scores) / len(recent_scores)
+    drift_magnitude = baseline_avg - recent_avg  # positive = drop
+
+    matched = drift_magnitude >= drift_threshold
```
lan17: Nice catch to compare recent vs baseline here. One thing I think will bite us is raw float precision on the exact-threshold boundary. For the added test case with a `1.0` baseline, `0.9` recent window, and `0.10` threshold, this ends up as `0.099999...`, so the evaluator returns `False` even though the documented behavior says `>=` should trigger. That also lines up with why `test_exactly_at_threshold_triggers` is currently failing. I think we should compare with a small tolerance or round before the threshold check.

evaluators/contrib/drift/pyproject.toml
```diff
@@ -0,0 +1,37 @@
+[project]
```
lan17, line 1: Happy to see this added as a standalone contrib package. I think we still need the repo-level release wiring though. Right now semantic-release, `scripts/build.py`, `test-extras`, and the release workflow still only know about the Galileo contrib package, so I do not think this one will actually get versioned, tested, and published from this repo yet.

evaluators/contrib/drift/src/agent_control_evaluator_drift/drift/evaluator.py
```diff
@@ -0,0 +1,286 @@
...
+                    "fallback_action": self.config.on_error,
+                },
+            )
+
+        # Append new score
+        scores.append(score)
+
+        # Persist updated history
+        try:
+            _save_history(history_path, scores)
```
lan17: I think this needs some synchronization around the history update path. Right now each call does load, append, and overwrite with no lock, so if two workers hit the same agent at once, the last writer wins and we drop observations. I was able to reproduce that with a small multiprocess harness. Since the drift result depends on having a complete history, this feels worth fixing before merge.

evaluators/contrib/drift/src/agent_control_evaluator_drift/drift/evaluator.py
nanookclaw: Good catch. Fixed with `round(drift_magnitude, 10) >= drift_threshold` — rounds at 10 decimal places which eliminates the ULP-level noise (1.0 - 0.9 = 0.09999999... → rounds to exactly 0.1) while preserving all meaningful precision. The test_exactly_at_threshold_triggers case should pass now.

evaluators/contrib/drift/pyproject.toml
nanookclaw: Agreed — the drift package was orphaned from the release pipeline. Fixed in the same commit:
- Added `DRIFT_DIR := evaluators/contrib/drift` to the Makefile and `drift-{test,lint,lint-fix,typecheck,build}` targets matching the galileo pattern
- Wired `drift-test` into `test-extras` (was galileo-only)
- Added `build_evaluator_drift()` to `scripts/build.py` with `'drift'` and `'all'` targets
The drift package will now get versioned, tested, and built alongside galileo on every release.

evaluators/contrib/drift/src/agent_control_evaluator_drift/drift/evaluator.py
nanookclaw: Good reproduction. Fixed by replacing the separate `_load_history()` / `_save_history()` calls with a single `_load_and_append_history()` that holds `fcntl.LOCK_EX` for the full read-modify-write cycle. The lock is scoped to a per-agent `.lock` file (e.g. `customer-support.lock` next to `customer-support.json`), so different agents remain fully parallel while same-agent concurrent calls serialize correctly.

No new dependency — `fcntl` is stdlib on POSIX. The lock file is never deleted (advisory locks don't need cleanup; the file descriptor releasing the lock is enough). Let me know if you'd prefer an explicit `try/finally` delete or if there's a project convention I should follow.

nanookclaw:
Rebased this on current `main` and force-pushed `110c9f8`.

Conflict resolution kept the drift evaluator fixes while using the repo's newer generic contrib-package wiring instead of reintroducing hardcoded `galileo`/`drift` build targets. I also added drift to the contrib verification contract and cleaned the package lint/test compatibility after the rebase.

Local checks run:
- `make contrib-verify`
- `python -m py_compile scripts/build.py scripts/contrib_packages.py`
- `uv run python scripts/contrib_packages.py names`
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run --extra dev python -m pytest -p pytest_asyncio.plugin tests -q` from `evaluators/contrib/drift` (`32 passed`)
- `uv run --extra dev ruff check --config ../../../pyproject.toml src/ tests/`
- `uv run --extra dev mypy --config-file ../../../pyproject.toml src/`
- `uv build` from `evaluators/contrib/drift`
- `git diff --check`

PR #141, opened 2026-03-20
feat: add financial governance evaluators (spend limits + transaction policy)

### LINKED ISSUE #129: Feature: Add financial governance evaluator for agent spend limits
## Problem

Agent Control supports behavioral governance (toxicity, PII, compliance) but has no concept of financial spend limits. As agents transact autonomously via x402 and other payment protocols, enterprises need governance over what agents spend.

## Proposed Solution

Add a `financial-governance` evaluator type that can:
- Track cumulative agent spend per policy period
- Enforce per-transaction and per-period financial limits
- Interface with wallet SDKs to validate transactions before execution

## Reference Implementation

[agentpay-mcp](https://github.com/up2itnow0822/agentpay-mcp) (npm: `agentpay-mcp`, v1.2.0) is an open-source MCP server implementing non-custodial x402 payments with on-chain spend limits. Could serve as a reference evaluator.

Key features mapping to Agent Control policies:
- Per-transaction spend caps
- Rolling period budgets
- Allowlisted recipient addresses
- Transaction queuing when limits exceeded

Happy to contribute if interested.

up2itnow0822:
## Summary

Implements the financial governance evaluator proposed in #129, following @lan17's technical guidance.

Two contrib evaluators for enforcing financial spend limits on autonomous AI agent transactions:

### `financial_governance.spend_limit`
Tracks cumulative agent spend and enforces rolling budget limits:
- Per-transaction caps
- Rolling period budgets (configurable window, default 24h)
- Context-aware overrides (different limits per channel/agent/session via evaluate metadata)
- Pluggable `SpendStore` protocol — `InMemorySpendStore` included, bring your own PostgreSQL/Redis

### `financial_governance.transaction_policy`
Static policy checks with no state tracking:
- Currency allowlist
- Recipient blocklist/allowlist  
- Min/max amount bounds

## Design Decisions (per #129 discussion)

1. **Decoupled from data source** — `SpendStore` protocol means no new tables in core. Custom backends implement two methods: `record_spend()` and `get_spend()`.
2. **No changes to core** — Self-contained contrib package under `evaluators/contrib/financial-governance/`.
3. **Context-aware limits** — Override keys in the evaluate data dict (`channel_max_per_transaction`, `channel_max_per_period`) allow per-context policies without separate evaluator instances.
4. **Python SDK compatible** — Standard `Evaluator` interface, works with both server and SDK evaluation engine.

## Tests

53 tests passing — covers both evaluators, the `InMemorySpendStore`, config validation, edge cases, and context overrides.

```
pytest tests/ -v
53 passed in 2.79s
```

## Related

- Closes #129
- Reference implementation: [agentpay-mcp](https://github.com/up2itnow0822/agentpay-mcp) (non-custodial x402 payments)
- Protocol: [x402](https://github.com/coinbase/x402)

### DIFF
--- a/evaluators/contrib/financial-governance/README.md
+++ b/evaluators/contrib/financial-governance/README.md
@@ -127,11 +127,18 @@ rolling = BudgetLimit(
     window=BudgetWindow(kind="rolling", seconds=86400),
 )
 
-# Calendar-day budget (UTC)
+# Calendar-day budget (UTC by default)
 daily = BudgetLimit(
     amount=Decimal("500"),
     currency="USDC",
     window=BudgetWindow(kind="fixed", unit="day"),
+)
+
+# Calendar-day budget aligned to New York local midnight
+ny_daily = BudgetLimit(
+    amount=Decimal("500"),
+    currency="USDC",
+    window=BudgetWindow(kind="fixed", unit="day", timezone="America/New_York"),
 )
 
 config = SpendLimitConfig(limits=[cap, rolling, daily])
@@ -143,7 +150,7 @@ evaluator = SpendLimitEvaluator(config)
 | kind | Required fields | Notes |
 |------|----------------|-------|
 | `"rolling"` | `seconds` | Sliding window from `now - seconds` |
-| `"fixed"` | `unit` (`"day"`, `"week"`, or `"month"`) | Calendar-aligned, UTC by default |
+| `"fixed"` | `unit` (`"day"`, `"week"`, or `"month"`), optional `timezone` | Calendar-aligned in the configured IANA timezone, UTC by default |
 
 ### scope_by semantics
 
@@ -197,7 +204,13 @@ result = await evaluator.evaluate({
 
 ## Custom SpendStore
 
-The `SpendStore` protocol requires three methods. Implement them for your backend:
+The `SpendStore` protocol requires four methods for full evaluator compatibility:
+- `record_spend()`
+- `get_spend()`
+- `check_and_record()`
+- `check_and_record_many()`
+
+Implement them for your backend:
 
 ```python
 from decimal import Decimal
@@ -268,12 +281,29 @@ class PostgresSpendStore:
             self.record_spend(amount, currency, metadata)
             return True, current
 
+    def check_and_record_many(
+        self,
+        amount: Decimal,
+        currency: str,
+        checks: list[BudgetCheck],
+        metadata: dict | None = None,
+    ) -> tuple[bool, int | None, list[Decimal]]:
+        with self._conn.transaction():
+            current_spends: list[Decimal] = []
+            for idx, check in enumerate(checks):
+                current = self.get_spend(currency, check.start, check.end, check.scope)
+                current_spends.append(current)
+                if current + amount > check.limit:
+                    return False, idx, current_spends
+            self.record_spend(amount, currency, metadata)
+            return True, None, current_spends
+
 # Use it:
 store = PostgresSpendStore("postgresql://...")
 evaluator = SpendLimitEvaluator(config, store=store)
 ```
 
-> **Single-process atomicity note:** `InMemorySpendStore.check_and_record()` uses a `threading.Lock` to atomically check-and-record within a single process. For multi-process or distributed deployments, your custom store must implement true database-level atomics (e.g., PostgreSQL `SELECT ... FOR UPDATE`, Redis Lua scripts).
+> **Single-process atomicity note:** `InMemorySpendStore.check_and_record()` and `check_and_record_many()` use a `threading.Lock` to atomically validate and record within a single process. For multi-process or distributed deployments, your custom store must implement true database-level atomics (e.g., PostgreSQL `SELECT ... FOR UPDATE`, Redis Lua scripts).
 
 ## Running Tests
 
--- a/evaluators/contrib/financial-governance/README.md
+++ b/evaluators/contrib/financial-governance/README.md
@@ -10,10 +10,12 @@ As agents transact autonomously via protocols like [x402](https://github.com/coi
 
 Tracks cumulative agent spend and enforces rolling budget limits. Stateful — records approved transactions and checks new ones against accumulated spend.
 
-- **Per-transaction cap** — reject any single payment above a threshold
-- **Rolling period budget** — reject payments that would exceed a time-windowed budget
-- **Context-aware overrides** — different limits per channel, agent, or session via evaluate metadata
+- **Per-transaction cap** — reject any single payment above a threshold (`BudgetLimit` with no window)
+- **Rolling period budget** — reject payments that would exceed a time-windowed budget (`BudgetWindow(kind="rolling", ...)`)
+- **Calendar-aligned budget** — reject payments that exceed a day/week/month budget (`BudgetWindow(kind="fixed", ...)`)
+- **Scoped budgets** — independent counters per channel, agent, or session via `scope_by`
 - **Pluggable storage** — abstract `SpendStore` protocol with built-in `InMemorySpendStore`; bring your own PostgreSQL, Redis, etc.
+- **Atomic enforcement** — `check_and_record()` prevents TOCTOU races in single-process deployments
 
 ### `financial_governance.transaction_policy`
 
@@ -35,16 +37,25 @@ pip install -e ".[dev]"
 
 ### Spend Limit
 
+The `spend_limit` evaluator is configured via a list of `BudgetLimit` objects. Each limit is evaluated independently — the first violation wins.
+
 ```yaml
 controls:
   - name: spend-limit
     evaluator:
       type: financial_governance.spend_limit
       config:
-        max_per_transaction: 100.0    # Max USDC per single payment
-        max_per_period: 1000.0        # Rolling 24h budget
-        period_seconds: 86400         # Budget window (default: 24 hours)
-        currency: USDC                # Currency to govern
+        limits:
+          # Per-transaction cap: single payment ≤ 100 USDC
+          - amount: "100.00"
+            currency: USDC
+          # Per-channel rolling 24h budget: each channel limited to 1000 USDC/day
+          - amount: "1000.00"
+            currency: USDC
+            scope_by: [channel]
+            window:
+              kind: rolling
+              seconds: 86400
     selector:
       path: input                     # Extract step.input (transaction dict)
     action: deny
@@ -61,8 +72,8 @@ controls:
         allowed_currencies: [USDC, USDT]
         blocked_recipients: ["0xDEAD..."]
         allowed_recipients: ["0xALICE...", "0xBOB..."]
-        min_amount: 0.01
-        max_amount: 5000.0
+        min_amount: "0.01"
+        max_amount: "5000.00"
     selector:
       path: input
     action: deny
@@ -82,15 +93,69 @@ The transaction dict (from `step.input`) should contain:
 ```python
 # step.input — transaction payload
 {
-    "amount": 50.0,              # required — transaction amount
-    "currency": "USDC",          # required — payment currency
-    "recipient": "0xABC...",     # required — payment recipient
+    "amount": "50.00",             # required — Decimal or numeric string
+    "currency": "USDC",            # required — payment currency
+    "recipient": "0xABC...",       # required — payment recipient
+    # optional context fields (used for scope_by)
+    "channel": "slack",
+    "agent_id": "agent-42",
+    "session_id": "sess-1",
 }
 ```
 
+> **Note:** Use `Decimal` or string representations for `amount` — never raw `float`. Floating-point arithmetic is imprecise for money. The evaluator internally converts to `Decimal`.
+
+## BudgetLimit Model
+
+```python
+from decimal import Decimal
+from agent_control_evaluator_financial_governance.spend_limit import (
+    BudgetLimit, BudgetWindow, SpendLimitConfig, SpendLimitEvaluator,
+)
+
+# Per-transaction cap (no window)
+cap = BudgetLimit(amount=Decimal("100"), currency="USDC")
+
+# Rolling 24-hour budget, scoped per channel
+rolling = BudgetLimit(
+    amount=Decimal("1000"),
+    currency="USDC",
+    scope_by=("channel",),
+    window=BudgetWindow(kind="rolling", seconds=86400),
+)
+
+# Calendar-day budget (UTC)
+daily = BudgetLimit(
+    amount=Decimal("500"),
+    currency="USDC",
+    window=BudgetWindow(kind="fixed", unit="day"),
+)
+
+config = SpendLimitConfig(limits=[cap, rolling, daily])
+evaluator = SpendLimitEvaluator(config)
+```
+
+### BudgetWindow
+
+| kind | Required fields | Notes |
+|------|----------------|-------|
+| `"rolling"` | `seconds` | Sliding window from `now - seconds` |
+| `"fixed"` | `unit` (`"day"`, `"week"`, or `"month"`) | Calendar-aligned, UTC by default |
+
+### scope_by semantics
+
+`scope_by` lists the context dimension keys to isolate spend buckets. Each dimension is **independent**:
+
+- `scope_by=()` (default) — global budget: all spend in that currency shares one counter
+- `scope_by=("channel",)` — one counter per unique `channel` value
+- `scope_by=("agent_id",)` — one counter per unique `agent_id`
+- `scope_by=("channel", "agent_id")` — one counter per unique `(channel, agent_id)` pair
+
+Spend in `channel-A` does **not** count against `channel-B`'s budget.
+
 ## Context-Aware Limits
 
-Context fields (`channel`, `agent_id`, `session_id`) and per-context limit overrides can be provided in two ways:
+Context fields (`channel`, `agent_id`, `session_id`) can be provided in two ways:
 
 **Option A: Via `step.context`** (recommended for engine integration)
 
@@ -98,42 +163,36 @@ Context fields (`channel`, `agent_id`, `session_id`) and per-context limit overr
 step = Step(
     type="tool",
     name="payment",
-    input={"amount": 75.0, "currency": "USDC", "recipient": "0xABC"},
+    input={"amount": "75.00", "currency": "USDC", "recipient": "0xABC"},
     context={
         "channel": "experimental",
         "agent_id": "agent-42",
-        "channel_max_per_transaction": 50.0,
-        "channel_max_per_period": 200.0,
     },
 )
 ```
 
-When using `selector.path: "*"`, the evaluator merges `step.context` fields into the transaction data automatically. When using `selector.path: "input"`, context fields must be included directly in `step.input`.
+When using `selector.path: "*"`, the evaluator merges `step.context` fields into the transaction data automatically. Fields already present in `step.input` are never overwritten by context.
 
 **Option B: Inline in the transaction dict** (simpler, for direct SDK use)
 
 ```python
 result = await evaluator.evaluate({
-    "amount": 75.0,
+    "amount": "75.00",
     "currency": "USDC",
     "recipient": "0xABC",
     "channel": "experimental",
-    "channel_max_per_transaction": 50.0,
-    "channel_max_per_period": 200.0,
+    "agent_id": "agent-42",
 })
 ```
 
-Spend budgets are **scoped by context** — spend in channel A does not count against channel B's budget. When no context fields are present, budgets are global.
-
 ## Custom SpendStore
 
-The `SpendStore` protocol requires two methods. Implement them for your backend:
+The `SpendStore` protocol requires three methods. Implement them for your backend:
 
 ```python
+from decimal import Decimal
 from agent_control_evaluator_financial_governance.spend_limit import (
-    SpendStore,
-    SpendLimitConfig,
-    SpendLimitEvaluator,
+    SpendStore, SpendLimitConfig, SpendLimitEvaluator,
 )
 
 class PostgresSpendStore:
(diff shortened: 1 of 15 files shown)


evaluators/contrib/financial-governance/src/agent_control_evaluator_financial_governance/spend_limit/evaluator.py
```diff
@@ -0,0 +1,269 @@
...
+            )
+
+        # ---- Extract required fields ----
+        amount = _extract_float(data, "amount")
+        if amount is None:
+            return EvaluatorResult(
+                matched=False,
+                confidence=1.0,
+                message="Transaction data missing required field 'amount'",
+                error="Missing required field: amount",
```
lan17: I think this is conflating malformed runtime payload with evaluator failure. The config itself is already validated via the Pydantic config model; what this code is checking here is the selector output passed into `evaluate()`. `EvaluatorResult.error` is supposed to be for crashes/timeouts/missing deps, and the engine treats any non-null `error` as an errored control. So with a deny rule, a malformed transaction now fail-closes as an evaluator error instead of behaving like a normal policy result. I think these branches should stay in the regular matched/non-matched path and reserve `error` for actual evaluator failures. Same issue shows up in `transaction_policy`.

evaluators/contrib/financial-governance/tests/test_spend_limit.py
```diff
@@ -0,0 +1,294 @@
...
+    # Base config allows up to 1000 per tx, but channel caps at 50
+    ev = _make_evaluator(max_per_transaction=1000.0)
+    result = await ev.evaluate(_tx(amount=75.0, channel_max_per_transaction=50.0))
+    assert result.matched is True
+    assert result.metadata and result.metadata["violation"] == "per_transaction_cap"
+    assert result.metadata["max_per_transaction"] == pytest.approx(50.0)
+
+
+@pytest.mark.asyncio
+async def test_context_override_channel_max_per_period() -> None:
```
lan17: I’d add a stronger test around the context-aware budget story here. Right now this only proves that the override number is read; it doesn’t prove spend is actually isolated by `channel` / `agent_id` / `session_id`. A case like “90 USDC in channel A, then 20 USDC in channel B with `channel_max_per_period=100`” would catch whether the store lookup is scoped or just summing all USDC spend. I’d also like at least one engine-level test for the invalid-input path so we don’t accidentally lock in `result.error` as a policy outcome.

evaluators/contrib/financial-governance/src/agent_control_evaluator_financial_governance/spend_limit/store.py
```diff
@@ -0,0 +1,158 @@
...
+        """Persist a completed (or pending) spend record.
+
+        Args:
+            amount: Positive monetary amount that was spent.
+            currency: ISO-4217 or token symbol (e.g. ``"USDC"``).
+            metadata: Optional key-value bag for agent_id, session_id, etc.
+        """
+        ...
+
+    def get_spend(self, currency: str, since_timestamp: float) -> float:
```
lan17: I think there’s a deeper issue with the API shape here. `get_spend()` only lets the evaluator query by `(currency, since_timestamp)`, but the evaluator is also recording `channel` / `agent_id` / `session_id` metadata and the README/tests are positioning this as context-aware budgeting. As written, `channel_max_per_period` is still checking against all spend in that currency, not spend for the current channel/agent/session.

I reproduced it locally with 90 USDC in channel A, then a 20 USDC transaction in channel B with `channel_max_per_period=100`, and the second transaction gets denied because the lookup is summing global USDC spend. Feels like the store API needs a scoped query (or an atomic check+record API) before this behavior is actually correct.

evaluators/contrib/financial-governance/src/agent_control_evaluator_financial_governance/__init__.py
```diff
@@ -0,0 +1,46 @@
...
+  blocklists, amount bounds, and permitted currencies.
+
+Both evaluators are registered automatically when this package is installed and
+the ``agent_control.evaluators`` entry point group is discovered.
+
+Example usage in an agent-control control config::
+
+    {
+      "condition": {
+        "selector": {"path": "*"},
```
lan17: If this is meant to be a real Agent Control config example, I think `selector.path: "*"` is misleading here. In the engine, `*` passes the whole `Step` object into the evaluator, not the raw transaction dict, so `amount` / `currency` / `recipient` won’t be top-level fields.

`input` seems closer to what this evaluator actually expects, but then the context-aware override story probably needs to be spelled out separately since step context lives under `context`, not inside `input`.

lan17 (commented):
Latest pass looks much better overall. The original evaluator-error and scoped-budget issues seem fixed. I just left a few follow-ups on packaging/docs/test coverage.

evaluators/contrib/financial-governance/pyproject.toml
```diff
@@ -0,0 +1,55 @@
+[project]
+name = "agent-control-evaluator-financial-governance"
```
lan17, line 2: Nice to have this as a standalone package, but I do not think it is actually reachable for end users yet. As-is, I do not think `pip install "agent-control-evaluators[financial-governance]"` will pull this in, since `agent-control-evaluators` only exposes `galileo` and `cisco` extras today, and I do not see release wiring to publish this contrib package either. If the goal is for this to be installable the same way as the other optional evaluators, I think we still need the extra in `evaluators/builtin/pyproject.toml` plus the publish/release wiring.

evaluators/contrib/financial-governance/tests/test_spend_limit.py
```diff
@@ -0,0 +1,467 @@
...
+    step_data = {
+        "type": "tool",
+        "name": "payment",
+        "input": {"amount": 10.0, "currency": "USDC", "recipient": "0xABC", "channel": "from-input"},
+        "context": {"channel": "from-context"},
+    }
+    result = await ev.evaluate(step_data)
+    assert result.matched is False
+    # input's channel should win (not clobbered)
+    assert result.metadata and result.metadata.get("channel") is None or True  # just verify no crash
```
lan17: I do not think this test is asserting the intended behavior right now. `... or True` makes it always pass, and on the success path `result.metadata` does not even carry `channel`, so this will not catch a regression in input-vs-context precedence. I would tighten this to assert against the store state or another observable that proves `input["channel"]` won over `context["channel"]`.

evaluators/contrib/financial-governance/README.md
```diff
@@ -0,0 +1,185 @@
...
+    def __init__(self, connection_string: str):
+        self._conn = connect(connection_string)
+
+    def record_spend(self, amount: float, currency: str, metadata: dict | None = None) -> None:
+        self._conn.execute(
+            "INSERT INTO agent_spend (amount, currency, metadata, recorded_at) VALUES (%s, %s, %s, NOW())",
+            (amount, currency, json.dumps(metadata)),
+        )
+
+    def get_spend(self, currency: str, since_timestamp: float) -> float:
```
lan17: The custom store example is already stale relative to the protocol above. `SpendStore.get_spend()` now takes `scope`, so anyone copying this signature will implement the wrong interface and miss the context-scoped budget behavior. I would update the example to include `scope: dict[str, str] | None = None` and show how the metadata filter is applied.

evaluators/contrib/financial-governance/README.md
```diff
@@ -0,0 +1,185 @@
...
+pip install -e ".[dev]"
+pytest tests/ -v
+```
+
+## Design Decisions
+
+1. **Decoupled from data source** — The `SpendStore` protocol means no new tables in core Agent Control. Bring your own persistence.
+2. **Context-aware limits** — Override keys in the evaluate data dict allow per-channel, per-agent, or per-session limits without multiple evaluator instances.
+3. **Python SDK compatible** — Uses the standard evaluator interface; works with both the server and the Python SDK evaluation engine.
+4. **Fail-open on errors** — Missing or malformed data returns `matched=False` with an `error` field, following Agent Control conventions.
```
lan17: This line does not match the implementation anymore. After the fix, malformed runtime payload returns `matched=False` with `error=None`, which is the right shape. Leaving this as written is going to send people back toward the old behavior.

evaluators/contrib/financial-governance/src/agent_control_evaluator_financial_governance/spend_limit/evaluator.py
```diff
@@ -0,0 +1,329 @@
...
+            if any(k in data for k in ("channel", "agent_id", "session_id")):
+                scope = {
+                    k: str(data[k])
+                    for k in ("channel", "agent_id", "session_id")
+                    if k in data and data[k] is not None
+                }
+                if not scope:
+                    scope = None
+
+            period_spend = self._store.get_spend(tx_currency, since, scope=scope)
```
lan17: I think the store contract is still racy for real budget enforcement. We read `get_spend(...)`, decide, and only then `record_spend(...)`, so two concurrent requests can both pass this check and both record, overshooting the cap. If this is only meant as a lightweight contrib example that may be fine, but if we want hard spend limits I think the store API needs an atomic `check_and_record` or reservation-style primitive.

evaluators/contrib/financial-governance/src/agent_control_evaluator_financial_governance/spend_limit/evaluator.py
```diff
@@ -0,0 +1,329 @@
...
+
+        # ---- Rolling period budget ----
+        if effective_max_per_period > 0:
+            since = time.time() - self.config.period_seconds
+
+            # Build scope for context-aware budget isolation.
+            # When channel/agent/session overrides are present, query only
+            # spend matching that context — not global spend.
+            scope: dict[str, str] | None = None
+            if any(k in data for k in ("channel", "agent_id", "session_id")):
```
lan17: The scoping semantics here are a little different from how I first read the docs. If a transaction carries both `channel` and `agent_id`, we scope to the exact tuple, not to an independent per-channel budget and per-agent budget. That means spend can effectively bypass a "channel budget" just by varying agent/session within the same channel. If tuple-scoped budgets are the intent, I would make that explicit.

lan17 (commented):
One more pass on the store shape. I think it is a reasonable MVP, but I still have two concerns around enforceability and scoping semantics.

evaluators/contrib/financial-governance/src/agent_control_evaluator_financial_governance/spend_limit/config.py
```diff
@@ -0,0 +1,68 @@
...
+
+        {
+          "max_per_transaction": 500.0,
+          "max_per_period": 5000.0,
+          "period_seconds": 86400,
+          "currency": "USDC"
+        }
+    """
+
+    max_per_transaction: float = Field(
```
lan17: I would be pretty hesitant to use `float` for money here. If this is meant to be a real spend-control API, I think we should either use `Decimal` or make amounts explicit integer minor/atomic units before more code starts depending on the current shape.

evaluators/contrib/financial-governance/src/agent_control_evaluator_financial_governance/spend_limit/store.py
```diff
@@ -0,0 +1,187 @@
...
+        """Persist a completed (or pending) spend record.
+
+        Args:
+            amount: Positive monetary amount that was spent.
+            currency: ISO-4217 or token symbol (e.g. ``"USDC"``).
+            metadata: Optional key-value bag for agent_id, session_id, etc.
+        """
+        ...
+
+    def get_spend(
```
lan17, line 88: The config/store API feels a bit too tied to one budget shape: one rolling window defined by `max_per_period + period_seconds`, queried as `get_spend(..., since_timestamp, ...)`. That works for this MVP, but if we ever want fixed daily/weekly windows or multiple concurrent limits, I think this will get awkward fast. I would consider moving toward structured limit definitions plus a range-based query (`start`, `end`) instead of baking rolling-window semantics into the store contract.

lan17 (commented):
A couple more API-design thoughts here. The current shape works for the narrow rolling-budget case, but I think there are two places where we may be locking ourselves in more than we want.

evaluators/contrib/financial-governance/src/agent_control_evaluator_financial_governance/spend_limit/store.py
lan17: Concrete sketch of what I had in mind, roughly:

```python
class BudgetWindow(BaseModel):
    kind: Literal["rolling", "fixed"]
    seconds: int | None = None          # rolling window
    unit: Literal["day", "week", "month"] | None = None
    timezone: str | None = None         # fixed calendar windows

class BudgetLimit(BaseModel):
    amount: Decimal
    currency: str
    scope_by: tuple[Literal["channel", "agent_id", "session_id"], ...] = ()
    window: BudgetWindow | None = None  # None => per-transaction cap
```

Then config can just be `limits: list[BudgetLimit]`, and the store/query side can move toward `get_spend(currency, start, end, scope=...)` instead of baking rolling-window semantics into `since_timestamp`. Not saying this exact API, just the shape.

evaluators/contrib/financial-governance/src/agent_control_evaluator_financial_governance/spend_limit/evaluator.py
```diff
@@ -0,0 +1,377 @@
...
+                        },
+                    )
+                # Per-tx cap passes → no need to "record" a cap (it's per-call)
+
+            else:
+                # Rolling / fixed period budget
+                win_start = _window_start(limit)
+                period_limits_to_record.append((limit, scope, win_start))
+
+                period_spend = self._store.get_spend(tx_currency, win_start, scope=scope)
```
lan17: I think this is still bypassing the atomic path you added to the store. We read current spend here and only record later, so two concurrent requests can still both pass and both write. I reproduced it locally with two concurrent 60 USDC requests under a 100 USDC cap: both evaluations returned allow and the ledger ended at 120. If we want hard spend limits, I think the evaluator needs to call `check_and_record()` here rather than `get_spend()` and a later `record_spend()`.

evaluators/contrib/financial-governance/src/agent_control_evaluator_financial_governance/spend_limit/evaluator.py
```diff
@@ -0,0 +1,377 @@
...
+        """Build the scope filter for *limit* from transaction *data*.
+
+        For each key in ``limit.scope_by``, extract the value from ``data``
+        (if present).  Returns ``None`` (global query) when scope_by is empty
+        or none of the specified keys are present in data.
+        """
+        if not limit.scope_by:
+            return None
+
+        scope: dict[str, str] = {}
```
lan17, line 219: I do not think dropping missing `scope_by` keys is safe here. For a limit configured as `(channel, agent_id)`, a transaction that only carries `channel` turns into a broader per-channel lookup, because `matches_scope()` treats the shorter dict as a subset match. That means a partially populated scope can inherit spend from other agents in the same channel, which is different from the `("channel", "agent_id")` pair semantics the README describes.

evaluators/contrib/financial-governance/src/agent_control_evaluator_financial_governance/spend_limit/store.py
```diff
@@ -0,0 +1,303 @@
...
+    read-decide-write sequence, making it atomic within a single process.
+    ``get_spend()`` + ``record_spend()`` called separately are *not* atomic
+    and may overshoot the cap under concurrent load.
+
+    Args:
+        max_age_seconds: Records older than this are eligible for pruning.
+            Defaults to 7 days (604 800 s).
+    """
+
+    def __init__(self, max_age_seconds: int = 604_800) -> None:
```
lan17: I think this default makes the built-in store incompatible with the `fixed month` window the docs advertise. Anything older than 7 days gets pruned, so by day 8 of the month the in-memory store starts undercounting monthly spend and can allow overshoot. If the in-memory store is only meant for tests/local dev that is fine, but then I would either derive retention from the longest configured window or document that month-level enforcement is not actually sound with the default store.

lan17 (changes requested):
Latest pass looks a lot better overall. The malformed-input handling and the basic scoped lookup story are much closer to what I had in mind.

I still do not think the spend-limit side is quite enforceable enough for merge, though, so I left three blocking comments inline:

1. the evaluator still bypasses the atomic store path and can overshoot under concurrency;
2. composite `scope_by` keys widen if one dimension is missing, which breaks the documented tuple semantics;
3. the default in-memory retention window is only 7 days, so `fixed month` budgets undercount by construction.

If this were framed as a lightweight contrib sketch I would be less worried, but the current docs are presenting it as real financial governance, so I think those need to be tightened before we merge.

up2itnow0822 (commented):
@lan17 - appreciate the thorough review. These are real issues, not style nits. Taking the three blockers:

**1. Concurrency: `get_spend()` then `record_spend()` is racy**

You are right - the read-then-write pattern is fundamentally unsound for hard budget enforcement. Two concurrent 60 USDC requests under a 100 USDC cap will both pass because neither sees the other's pending spend. I'll replace the two-step path with an atomic `check_and_record()` that does the budget check and ledger write in a single operation:

```python
class SpendStore(Protocol):
    async def check_and_record(
        self,
        scope: dict[str, str],
        amount: Decimal,
        currency: str,
        limit: Decimal,
        since: datetime,
    ) -> tuple[bool, Decimal]:
        """Atomically check budget and record if within limit.
        Returns (allowed, new_total). If not allowed, nothing is recorded."""
        ...
```

The in-memory store uses a lock. Production stores (Redis, Postgres) use their native atomic primitives (INCRBY with conditional, SELECT FOR UPDATE). The evaluator calls this single method instead of the read/decide/write sequence.

**2. Composite `scope_by` keys widen on missing dimensions**

Agreed - a transaction with only `channel` matching against a `(channel, agent_id)` limit should not fall through to a broader per-channel lookup. The fix: if any dimension in the configured `scope_by` tuple is missing from the transaction metadata, the limit does not apply to that transaction at all. Strict tuple matching, not subset matching.

```python
def matches_scope(limit_scope: tuple[str, ...], tx_meta: dict[str, str]) -> bool:
    # ALL dimensions must be present - missing any = no match
    return all(dim in tx_meta for dim in limit_scope)
```

This means a partially-scoped transaction gets evaluated against limits whose scope it fully satisfies, and is unconstrained by limits requiring dimensions it does not carry. If that's too permissive for some deployments, the evaluator could optionally reject transactions missing required scope dimensions entirely (fail-closed on incomplete metadata).

**3. 7-day retention breaks `fixed month` windows**

Clear bug. The retention window should be derived from the longest configured budget window, not a static default. For a monthly budget, the store needs at least 31 days of history. I'll change the default retention to `max(configured_window_seconds) * 1.5` (with a floor of 7 days for rolling-only configs), and add a note in the docs that the in-memory store is intended for development/testing - production deployments with monthly windows should use a persistent store.

**On `float` for money:**

Also agreed. I'll switch to `Decimal` throughout the config and store layer. The evaluator already computes comparisons, so using `Decimal` from the config level down prevents any floating-point accumulation errors before they start.

I'll push these fixes. Should have an updated commit within 24 hours.

up2itnow0822:
@lan17 - appreciate the latest pass. These three blockers are all valid and I agree they need to be fixed before merge. Here is how I plan to address each one:

**1. Concurrency: `get_spend()` then `record_spend()` is racy**

You reproduced the exact failure mode - two concurrent 60 USDC requests both passing a 100 USDC cap. The read-then-write pattern is fundamentally unsound for hard budget enforcement. I will replace the two-step path with an atomic `check_and_record()`:

```python
class SpendStore(Protocol):
    async def check_and_record(
        self,
        scope: dict[str, str],
        amount: Decimal,
        currency: str,
        limit: Decimal,
        since: datetime,
    ) -> tuple[bool, Decimal]:
        """Atomically check budget and record if within limit.
        Returns (allowed, new_total). If not allowed, nothing is recorded."""
        ...
```

The in-memory store uses a lock. Production stores (Redis, Postgres) use their native atomic primitives. The evaluator calls this single method instead of the read/decide/write sequence.

**2. Composite `scope_by` keys widen on missing dimensions**

Agreed. A transaction with only `channel` should not match a `(channel, agent_id)` limit via subset matching. The fix: strict tuple matching - if any dimension in the configured `scope_by` tuple is missing from the transaction metadata, the limit does not apply.

```python
def matches_scope(limit_scope: tuple[str, ...], tx_meta: dict[str, str]) -> bool:
    return all(dim in tx_meta for dim in limit_scope)
```

Transactions missing required scope dimensions get evaluated only against limits whose scope they fully satisfy.

**3. 7-day retention breaks `fixed month` windows**

Clear bug. The retention window should be derived from the longest configured budget window, not a static default. I will change the default to `max(configured_window_seconds) * 1.5` (floor of 7 days for rolling-only), and document that the in-memory store is for dev/testing only - production deployments with monthly windows should use a persistent backend.

**Also switching `float` to `Decimal`** throughout config and store, per your earlier feedback.

Will push these fixes within 24 hours.

up2itnow0822:
Thanks again for the detailed review. I pushed the follow-up packaging/docs pass and re-validated the contrib package.

What changed in the latest revision:
- wired the financial-governance package into packaging/release flow so it is buildable/installable, instead of only living as source in the repo
- added the contrib `pyproject.toml` and package entrypoints
- updated the README/docs to match the actual evaluator shape and installation path
- kept the earlier enforcement fixes in place: atomic `check_and_record()`, strict scope matching, 31-day default retention, and `Decimal`-based money handling

What I intentionally did not change:
- I kept this as a contrib package rather than moving it into core. That still feels like the right boundary for an opinionated financial-control surface while the API settles.
- I did not add a production store backend in this PR. The store contract is there, but shipping a Redis/Postgres implementation felt like separate scope from getting the evaluator semantics correct.

Verification from this pass:
- `python3 scripts/build.py` completed successfully
- fresh venv install of `-e evaluators/contrib/financial-governance` plus test deps succeeded
- `pytest evaluators/contrib/financial-governance/tests -q` passed in that clean env

Remaining follow-up / risk:
- real deployments still need a backend with true cross-process atomics, because the in-memory store is only a single-process dev/test implementation
- if you want a bundled Redis/Postgres store, I can split that into a follow-on PR instead of growing this one further

If this looks aligned now, I’d appreciate another pass on the packaging/installability side and the overall merge-readiness of the contrib shape.

up2itnow0822:
Addressed the new Codex findings in `e633cd9`.

Changes in this pass:
- spend-limit now validates all applicable period budgets first, then records exactly once only after every budget passes
- added `check_and_record_many()` to the in-memory store so multi-budget evaluation stays atomic within the single-process store path
- reject non-finite amounts (`NaN`, `Infinity`, `-Infinity`) in both evaluators before any comparisons
- added regression coverage for early-recording, duplicate-recording, partial-write, and non-finite amount cases

Verification:
- `uv run pytest tests -q`
- `uv run ruff check src tests`

This keeps the earlier atomicity fix, but avoids mutating ledger state before the full evaluator decision is known.

up2itnow0822:
Pushed follow-up commit `57ef71d` for the two remaining Codex findings.

Changes in this pass:
- fixed `fixed` budget windows to honor the configured IANA timezone instead of always anchoring to UTC
- updated the contrib README/custom store contract to include `check_and_record_many()` and the full evaluator-compatible method set
- added a regression test for timezone-aligned day windows

Verification:
- `uv run pytest tests/test_spend_limit.py -q`
- `uv run ruff check src tests`

Issue #130, opened 2026-03-16
feat(evaluators): add built-in budget evaluator for per-agent cost tracking


amabito:
## Summary
Built-in budget evaluator for per-agent cumulative cost tracking.

## Motivation
Retries and recursive tool chains pile up fast -- a 3-layer retry loop is 64 API calls from one user request. Current evaluators are stateless, so there's no way to express "deny after $X total spend" without maintaining a separate counter service outside the control plane.

## Current behavior
Controls evaluate step content (regex, list, JSON, SQL) but can't track cumulative state across evaluations. Cost enforcement requires a custom evaluator with external state management.

## Expected behavior
A built-in `budget` evaluator that:
1. Tracks cumulative cost per agent (in-memory, or via PostgreSQL for persistence)
2. Config: `max_cost_usd`, `cost_per_1k_input_tokens`, `cost_per_1k_output_tokens`
3. On post-stage evaluation, reads token counts from `step.output` or `step.context`, accumulates
4. Returns `matched=True` when ceiling is hit
5. Pairs with existing actions -- `deny` for hard stop, `steer` with `steering_context: {fallback_model: "..."}` for degradation

## Proposed solution
Should work as a regular evaluator. `confidence` could just be `spent / limit` (0-1 utilization ratio).

The stateful part is new -- current evaluators are stateless -- but the SQL evaluator already caches query analysis across calls, so there's precedent for evaluator-level state.

## Additional context
LMK if a PR for this makes sense.

amabito:
To keep scope clear: #130 is about LLM API cost derived from token usage in step data -- not wallet balances or payment transactions. Different accounting source, different evaluator, but the store abstraction could be shared later.

Here's a sketch against each of your four points.

**1. Store abstraction**

Budget evaluation stays storage-agnostic. Cumulative state lives behind a store interface:

```python
@dataclass(frozen=True)
class BudgetSnapshot:
    spent_usd: float
    limit_usd: float
    utilization: float  # spent / limit, clamped to [0.0, 1.0]
    exceeded: bool

class BudgetStore(Protocol):
    def record_and_check(
        self,
        agent_id: str,
        input_tokens: int,
        output_tokens: int,
        cost_usd: float,
        metadata: dict,
    ) -> BudgetSnapshot: ...
```

Single atomic call -- records usage and returns current totals in one shot, no read-then-write race.

**2. No new core tables**

Two implementations, only one in core:

[...]

PR #144, opened 2026-03-21
feat(evaluators): add built-in budget evaluator for per-agent cost tracking

amabito:
## Summary
- New built-in evaluator "budget" that tracks cumulative token/cost usage
  per agent, channel, user. Configurable time windows (daily/weekly/monthly).
- Addresses #130.

## Scope
- User-facing/API changes:
  - "budget" evaluator registered alongside regex/list/json/sql
  - BudgetStore protocol + InMemoryBudgetStore (dict + threading.Lock)
  - BudgetEvaluatorConfig: limits list, optional pricing table, path configs

- Internal changes:
  - evaluators/builtin/src/agent_control_evaluators/budget/ -- 4 files, ~650 LOC
  - evaluators/builtin/tests/budget/test_budget.py -- 63 tests

- Out of scope:
  - No Postgres store, no DB tables, no new dependencies

## Risk and Rollout
- Risk level: low -- new evaluator, no changes to existing code. 230 existing tests untouched.
- Rollback plan: revert PR

## Testing
- [x] Added or updated automated tests (63 tests incl. thread safety, NaN/Inf, scope injection, double-count)
- [x] Ran pytest tests/ -- 293 passed

## Checklist
- [x] Linked issue/spec -- #130
- [ ] Updated docs/examples -- follow-up: config example in docs/evaluators/
- [x] Included follow-up tasks -- Postgres BudgetStore (separate package)

### DIFF
--- a/evaluators/contrib/budget/README.md
+++ b/evaluators/contrib/budget/README.md
@@ -1,0 +1,3 @@
+# Budget Evaluator
+
+Cumulative LLM cost and token budget tracking for agent-control.
--- a/evaluators/contrib/budget/README.md
+++ b/evaluators/contrib/budget/README.md
@@ -1,3 +1,136 @@
-# Budget Evaluator
+# agent-control-evaluator-budget
 
-Cumulative LLM cost and token budget tracking for agent-control.
+Budget evaluator for agent-control that tracks cumulative LLM token and cost usage per scope and time window.
+
+## Install
+
+```bash
+pip install agent-control-evaluator-budget
+```
+
+## Quickstart
+
+```python
+from agent_control_evaluator_budget.budget import (
+    BudgetEvaluatorConfig,
+    BudgetLimitRule,
+    ModelPricing,
+)
+
+config = BudgetEvaluatorConfig(
+    budget_id="support-daily",
+    limits=[
+        BudgetLimitRule(
+            scope={"agent": "support"},
+            group_by="user_id",
+            window_seconds=86_400,
+            limit=500,
+            limit_unit="usd_cents",
+        ),
+        BudgetLimitRule(
+            scope={"agent": "support"},
+            group_by="user_id",
+            window_seconds=86_400,
+            limit=50_000,
+            limit_unit="tokens",
+        ),
+    ],
+    pricing={
+        "gpt-4.1-mini": ModelPricing(input_per_1k=0.04, output_per_1k=0.16),
+    },
+    model_path="model",
+    metadata_paths={
+        "agent": "metadata.agent",
+        "user_id": "metadata.user_id",
+    },
+    unknown_model_behavior="block",
+)
+```
+
+The evaluator reads token usage from standard fields such as `usage.input_tokens` and `usage.output_tokens`. Configure `token_path` only when your event shape uses a custom location.
+
+## Scope and group_by
+
+Each `BudgetLimitRule` has a static `scope` and an optional `group_by` field.
+
+`scope` filters which events a rule applies to. A rule with `scope={"agent": "support"}` only applies when extracted metadata contains `agent="support"`. An empty scope is global.
+
+`group_by` creates independent buckets per extracted metadata value. The common per-user pattern is:
+
+```python
+BudgetLimitRule(
+    scope={"agent": "support"},
+    group_by="user_id",
+    window_seconds=86_400,
+    limit=500,
+    limit_unit="usd_cents",
+)
+```
+
+With `metadata_paths={"user_id": "metadata.user_id"}`, each user gets a separate daily budget inside the support scope.
+
+## Budget pools
+
+`budget_id` identifies the accumulated budget pool.
+
+Evaluators with the same `budget_id` share accumulated spend and token totals across all evaluator instances. Each evaluator still evaluates using its own configured rules -- the shared state is the bucket (the rolling sum), not the rule set. Evaluators with different `budget_id` values are fully isolated.
+
+Use stable names such as `support-daily`, `billing-global`, or `tenant-acme-monthly`. Avoid generating a new `budget_id` per request unless each request should have an isolated budget.
+
+## Pricing
+
+`ModelPricing` stores cost rates in cents per 1K tokens:
+
+```python
+ModelPricing(input_per_1k=0.04, output_per_1k=0.16)
+```
+
+`input_per_1k` is applied to input tokens. `output_per_1k` is applied to output tokens.
+
+Pricing is required when any rule uses `limit_unit="usd_cents"`. Token-only rules can omit pricing. If an event uses a model that is not in the pricing table and a cost rule exists, `unknown_model_behavior="block"` fails closed. Use `"warn"` to log a warning and treat the cost as 0.
+
+## Dual Ceiling Pattern
+
+Use two evaluators when cost and token ceilings need independent control records or different `budget_id` pools:
+
+```python
+cost_config = BudgetEvaluatorConfig(
+    budget_id="support-cost-daily",
+    limits=[
+        BudgetLimitRule(
+            scope={"agent": "support"},
+            group_by="user_id",
+            window_seconds=86_400,
+            limit=500,
+            limit_unit="usd_cents",
+        )
+    ],
+    pricing={
+        "gpt-4.1-mini": ModelPricing(input_per_1k=0.04, output_per_1k=0.16),
+    },
+    model_path="model",
+    metadata_paths={"agent": "metadata.agent", "user_id": "metadata.user_id"},
+)
+
+token_config = BudgetEvaluatorConfig(
+    budget_id="support-token-daily",
+    limits=[
+        BudgetLimitRule(
+            scope={"agent": "support"},
+            group_by="user_id",
+            window_seconds=86_400,
+            limit=50_000,
+            limit_unit="tokens",
+        )
+    ],
+    metadata_paths={"agent": "metadata.agent", "user_id": "metadata.user_id"},
+)
+```
+
+This pattern lets cost and token budgets reset, alert, and roll out independently. A single evaluator can also contain both rules when one shared pool and one control result are sufficient.
+
+## Limitations
+
+`InMemoryBudgetStore` is single-process only. State is lost on restart and is not shared across workers or pods.
+
+Use a distributed store for production deployments that run multiple processes, multiple workers, or multiple pods.
--- a/evaluators/contrib/budget/src/agent_control_evaluator_budget/budget/__init__.py
+++ b/evaluators/contrib/budget/src/agent_control_evaluator_budget/budget/__init__.py
@@ -1,0 +1,14 @@
+"""Budget evaluator for per-agent LLM cost and token tracking."""
+
+from agent_control_evaluator_budget.budget.config import BudgetEvaluatorConfig
+from agent_control_evaluator_budget.budget.evaluator import BudgetEvaluator
+from agent_control_evaluator_budget.budget.memory_store import InMemoryBudgetStore
+from agent_control_evaluator_budget.budget.store import BudgetSnapshot, BudgetStore
+
+__all__ = [
+    "BudgetEvaluator",
+    "BudgetEvaluatorConfig",
+    "BudgetSnapshot",
+    "BudgetStore",
+    "InMemoryBudgetStore",
+]
--- a/evaluators/contrib/budget/src/agent_control_evaluator_budget/budget/__init__.py
+++ b/evaluators/contrib/budget/src/agent_control_evaluator_budget/budget/__init__.py
@@ -1,6 +1,10 @@
 """Budget evaluator for per-agent LLM cost and token tracking."""
 
-from agent_control_evaluator_budget.budget.config import BudgetEvaluatorConfig
+from agent_control_evaluator_budget.budget.config import (
+    BudgetEvaluatorConfig,
+    BudgetLimitRule,
+    ModelPricing,
+)
 from agent_control_evaluator_budget.budget.evaluator import BudgetEvaluator
 from agent_control_evaluator_budget.budget.memory_store import InMemoryBudgetStore
 from agent_control_evaluator_budget.budget.store import BudgetSnapshot, BudgetStore
@@ -12,7 +16,9 @@ from agent_control_evaluator_budget.budget.store import BudgetSnapshot, BudgetSt
 __all__ = [
     "BudgetEvaluator",
     "BudgetEvaluatorConfig",
+    "BudgetLimitRule",
     "BudgetSnapshot",
     "BudgetStore",
     "InMemoryBudgetStore",
+    "ModelPricing",
 ]
--- a/evaluators/contrib/budget/src/agent_control_evaluator_budget/budget/__init__.py
+++ b/evaluators/contrib/budget/src/agent_control_evaluator_budget/budget/__init__.py
@@ -5,6 +5,10 @@ from agent_control_evaluator_budget.budget.evaluator import BudgetEvaluator
 from agent_control_evaluator_budget.budget.memory_store import InMemoryBudgetStore
 from agent_control_evaluator_budget.budget.store import BudgetSnapshot, BudgetStore
 
+# Note: clear_budget_stores is a testing utility and is intentionally not
+# re-exported here. Import it directly from the evaluator submodule in tests:
+#   from agent_control_evaluator_budget.budget.evaluator import clear_budget_stores
+
 __all__ = [
     "BudgetEvaluator",
     "BudgetEvaluatorConfig",
--- a/evaluators/contrib/budget/src/agent_control_evaluator_budget/budget/config.py
+++ b/evaluators/contrib/budget/src/agent_control_evaluator_budget/budget/config.py
@@ -2,7 +2,7 @@
 
 from __future__ import annotations
 
-from enum import Enum
+from typing import Literal
 
 from agent_control_evaluators._base import EvaluatorConfig
 from pydantic import Field, field_validator, model_validator
@@ -17,12 +17,11 @@ WINDOW_WEEKLY = 604800
 WINDOW_MONTHLY = 2592000  # 30 days
 
 
-class Currency(str, Enum):
-    """Supported budget currencies."""
+class ModelPricing(EvaluatorConfig):
+    """Per-model token pricing in cents per 1K tokens."""
 
-    USD = "usd"
-    EUR = "eur"
-    TOKENS = "tokens"
+    input_per_1k: float = 0.0
+    output_per_1k: float = 0.0
 
 
 class BudgetLimitRule(EvaluatorConfig):
@@ -43,37 +42,22 @@ class BudgetLimitRule(EvaluatorConfig):
             each user gets their own budget. None = shared/global limit.
         window_seconds: Time window for accumulation in seconds.
             None = cumulative (no reset). See WINDOW_* constants.
-        limit: Maximum spend in the window, in minor units (e.g. cents
-            for USD). None = uncapped on this dimension.
-        currency: Currency for the limit. Defaults to USD.
-        limit_tokens: Maximum tokens in the window. None = uncapped.
+        limit: Maximum usage in the window. Interpreted by limit_unit.
+        limit_unit: Unit for limit. usd_cents checks spend; tokens checks
+            input + output tokens.
     """
 
     scope: dict[str, str] = Field(default_factory=dict)
     group_by: str | None = None
     window_seconds: int | None = None
-    limit: int | None = None
-    currency: Currency = Currency.USD
-    limit_tokens: int | None = None
-
-    @model_validator(mode="after")
-    def at_least_one_limit(self) -> "BudgetLimitRule":
-        if self.limit is None and self.limit_tokens is None:
-            raise ValueError("At least one of limit or limit_tokens must be set")
-        return self
+    limit: int
+    limit_unit: Literal["usd_cents", "tokens"] = "usd_cents"
 
     @field_validator("limit")
     @classmethod
-    def validate_limit(cls, v: int | None) -> int | None:
-        if v is not None and v <= 0:
+    def validate_limit(cls, v: int) -> int:
+        if v <= 0:
             raise ValueError("limit must be a positive integer")
-        return v
-
-    @field_validator("limit_tokens")
-    @classmethod
-    def validate_limit_tokens(cls, v: int | None) -> int | None:
-        if v is not None and v <= 0:
-            raise ValueError("limit_tokens must be positive")
         return v
 
     @field_validator("window_seconds")
@@ -89,9 +73,13 @@ class BudgetEvaluatorConfig(EvaluatorConfig):
 
     Attributes:
         limits: List of budget limit rules. Each is checked independently.
-        pricing: Optional model pricing table. Maps model name to per-1K
-            token rates. Used to derive cost in USD from token counts and
-            model name.
+        budget_id: Unique budget pool identifier. Same budget_id shares
+            accumulated spend. Different budget_id is fully isolated.
+        unknown_model_behavior: What to do when a model is not found in the
+            pricing table and a cost-based rule exists. block=fail closed,
+            warn=log warning and treat cost as 0.
+        pricing: Optional model pricing table. Maps model name to ModelPricing.
+            Used to derive cost in USD from token counts and model name.
         token_path: Dot-notation path to extract token usage from step
             data (e.g. "usage.total_tokens"). If None, looks for standard
             fields (input_tokens, output_tokens, total_tokens, usage).
@@ -101,7 +89,27 @@ class BudgetEvaluatorConfig(EvaluatorConfig):
     """
 
     limits: list[BudgetLimitRule] = Field(min_length=1)
-    pricing: dict[str, dict[str, float]] | None = None
+    budget_id: str = Field(
+        default="default",
+        description=(
+            "Unique budget pool identifier. Same budget_id shares accumulated spend. "
+            "Different budget_id is fully isolated."
+        ),
+    )
+    unknown_model_behavior: Literal["block", "warn"] = Field(
+        default="block",
+        description=(
+            "What to do when a model is not found in the pricing table and a cost-based "
+            "rule exists. block=fail closed, warn=log warning and treat cost as 0."
+        ),
+    )
+    pricing: dict[str, ModelPricing] | None = None
     token_path: str | None = None
     model_path: str | None = None
     metadata_paths: dict[str, str] = Field(default_factory=dict)
+
+    @model_validator(mode="after")
+    def require_pricing_for_cost_rules(self) -> "BudgetEvaluatorConfig":
+        if self.pricing is None and any(rule.limit_unit == "usd_cents" for rule in self.limits):
+            raise ValueError('pricing is required when any rule uses limit_unit="usd_cents"')
+        return self
--- a/evaluators/contrib/budget/src/agent_control_evaluator_budget/budget/config.py
+++ b/evaluators/contrib/budget/src/agent_control_evaluator_budget/budget/config.py
@@ -110,6 +110,9 @@ class BudgetEvaluatorConfig(EvaluatorConfig):
 
     @model_validator(mode="after")
     def require_pricing_for_cost_rules(self) -> "BudgetEvaluatorConfig":
-        if self.pricing is None and any(rule.limit_unit == "usd_cents" for rule in self.limits):
+        has_cost_rule = any(rule.limit_unit == "usd_cents" for rule in self.limits)
+        if has_cost_rule and self.pricing is None:
             raise ValueError('pricing is required when any rule uses limit_unit="usd_cents"')
+        if has_cost_rule and not (self.model_path or "").strip():
+            raise ValueError('model_path is required when any rule uses limit_unit="usd_cents"')
         return self
--- a/evaluators/contrib/budget/src/agent_control_evaluator_budget/budget/config.py
+++ b/evaluators/contrib/budget/src/agent_control_evaluator_budget/budget/config.py
@@ -1,0 +1,107 @@
+"""Configuration for the budget evaluator."""
+
+from __future__ import annotations
+
+from enum import Enum
+
+from agent_control_evaluators._base import EvaluatorConfig
+from pydantic import Field, field_validator, model_validator
+
+# ---------------------------------------------------------------------------
+# Window convenience constants (seconds)
+# ---------------------------------------------------------------------------
+
+WINDOW_HOURLY = 3600
+WINDOW_DAILY = 86400
+WINDOW_WEEKLY = 604800
+WINDOW_MONTHLY = 2592000  # 30 days
+
+
+class Currency(str, Enum):
+    """Supported budget currencies."""
+
+    USD = "usd"
+    EUR = "eur"
+    TOKENS = "tokens"
+
+
+class BudgetLimitRule(EvaluatorConfig):
+    """A single budget limit rule.
+
+    Each rule defines a ceiling for a combination of scope dimensions
+    and time window. Multiple rules can apply to the same step -- the
+    evaluator checks all of them and triggers on the first breach.
+
+    Attributes:
+        scope: Static scope dimensions that must match for this rule
+            to apply. Empty dict = global rule.
+            Examples:
+                {"agent": "summarizer"} -- per-agent limit
+                {"agent": "summarizer", "channel": "slack"} -- agent+channel limit
+        group_by: If set, the limit is applied independently for each
+            unique value of this dimension. e.g. group_by="user_id" means
+            each user gets their own budget. None = shared/global limit.
+        window_seconds: Time window for accumulation in seconds.
+            None = cumulative (no reset). See WINDOW_* constants.
+        limit: Maximum spend in the window, in minor units (e.g. cents
+            for USD). None = uncapped on this dimension.
+        currency: Currency for the limit. Defaults to USD.
+        limit_tokens: Maximum tokens in the window. None = uncapped.
+    """
+
+    scope: dict[str, str] = Field(default_factory=dict)
+    group_by: str | None = None
+    window_seconds: int | None = None
+    limit: int | None = None
+    currency: Currency = Currency.USD
+    limit_tokens: int | None = None
+
+    @model_validator(mode="after")
+    def at_least_one_limit(self) -> "BudgetLimitRule":
+        if self.limit is None and self.limit_tokens is None:
+            raise ValueError("At least one of limit or limit_tokens must be set")
+        return self
+
+    @field_validator("limit")
+    @classmethod
+    def validate_limit(cls, v: int | None) -> int | None:
+        if v is not None and v <= 0:
+            raise ValueError("limit must be a positive integer")
+        return v
+
+    @field_validator("limit_tokens")
+    @classmethod
+    def validate_limit_tokens(cls, v: int | None) -> int | None:
+        if v is not None and v <= 0:
+            raise ValueError("limit_tokens must be positive")
+        return v
+
+    @field_validator("window_seconds")
+    @classmethod
+    def validate_window_seconds(cls, v: int | None) -> int | None:
+        if v is not None and v <= 0:
+            raise ValueError("window_seconds must be positive")
+        return v
+
+
+class BudgetEvaluatorConfig(EvaluatorConfig):
+    """Configuration for the budget evaluator.
+
+    Attributes:
+        limits: List of budget limit rules. Each is checked independently.
+        pricing: Optional model pricing table. Maps model name to per-1K
+            token rates. Used to derive cost in USD from token counts and
+            model name.
+        token_path: Dot-notation path to extract token usage from step
+            data (e.g. "usage.total_tokens"). If None, looks for standard
+            fields (input_tokens, output_tokens, total_tokens, usage).
+        model_path: Dot-notation path to extract model name (for pricing lookup).
+        metadata_paths: Mapping of metadata field name to dot-notation path
+            in step data. Used to extract scope dimensions (channel, user_id, etc).
+    """
+
+    limits: list[BudgetLimitRule] = Field(min_length=1)
+    pricing: dict[str, dict[str, float]] | None = None
+    token_path: str | None = None
+    model_path: str | None = None
+    metadata_paths: dict[str, str] = Field(default_factory=dict)
--- a/evaluators/contrib/budget/src/agent_control_evaluator_budget/budget/evaluator.py
+++ b/evaluators/contrib/budget/src/agent_control_evaluator_budget/budget/evaluator.py
@@ -7,12 +7,11 @@ are returned in result metadata, not in confidence.
 The evaluator is stateless. Budget state lives in a module-level store
 registry, independent of the evaluator instance cache in _factory.py.
 This prevents silent state loss on LRU eviction and avoids cross-control
-leakage when different controls share the same config.
+leakage when different controls use different budget_id values.
 """
 
 from __future__ import annotations
 
-import json
 import logging
 import math
 import threading
@@ -22,7 +21,7 @@ from agent_control_evaluators._base import Evaluator, EvaluatorMetadata
 from agent_control_evaluators._registry import register_evaluator
 from agent_control_models import EvaluatorResult
 
-from .config import BudgetEvaluatorConfig
+from .config import BudgetEvaluatorConfig, ModelPricing
 from .memory_store import InMemoryBudgetStore
 from .store import BudgetStore
 
@@ -33,8 +32,8 @@ logger = logging.getLogger(__name__)
 #
 # Decoupled from the evaluator instance cache so that LRU eviction in
 # _factory.py does not destroy accumulated budget state. The registry
-# is keyed by a stable config hash. Two controls with identical config
-# intentionally share a budget pool (same config = same budget).
+# is keyed by budget_id. Controls with the same budget_id intentionally
+# share accumulated spend; different budget_id values are isolated.
 # ---------------------------------------------------------------------------
 
 # NOTE: The registry is unbounded. In practice a deployment has a finite
@@ -44,27 +43,13 @@ _STORE_REGISTRY: dict[str, BudgetStore] = {}
 _STORE_REGISTRY_LOCK = threading.Lock()
 
 
-def _config_key(config: BudgetEvaluatorConfig) -> str:
-    """Build a stable key for the store registry from evaluator config.
-
-    The limits list is sorted before hashing so that two configs with
-    semantically identical rules in different order share a store.
-    """
-    config_dict = config.model_dump(mode="json")
-    config_dict["limits"] = sorted(
-        config_dict["limits"],
-        key=lambda r: json.dumps(r, sort_keys=True, default=str),
-    )
-    return f"budget:{json.dumps(config_dict, sort_keys=True, default=str)}"
-
-
 def get_or_create_store(config: BudgetEvaluatorConfig) -> BudgetStore:
     """Get or create a store for the given config, thread-safe."""
-    key = _config_key(config)
+    key = f"budget:{config.budget_id}"
     with _STORE_REGISTRY_LOCK:
         store = _STORE_REGISTRY.get(key)
         if store is None:
-            store = InMemoryBudgetStore(rules=config.limits)
+            store = InMemoryBudgetStore()
             _STORE_REGISTRY[key] = store
         return store
 
@@ -140,7 +125,7 @@ def _estimate_cost(
     model: str | None,
     input_tokens: int,
     output_tokens: int,
-    pricing: dict[str, dict[str, float]] | None,
+    pricing: dict[str, ModelPricing] | None,
 ) -> float:
     """Estimate cost in cents (USD) from model pricing table.
 
@@ -152,8 +137,8 @@ def _estimate_cost(
     rates = pricing.get(model)
     if not rates:
         return 0.0
-    input_rate = rates.get("input_per_1k", 0.0)
-    output_rate = rates.get("output_per_1k", 0.0)
+    input_rate = rates.input_per_1k
+    output_rate = rates.output_per_1k
     cost = (input_tokens * input_rate + output_tokens * output_rate) / 1000.0
     if not math.isfinite(cost) or cost < 0:
         return 0.0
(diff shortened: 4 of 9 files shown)


evaluators/builtin/src/agent_control_evaluators/budget/config.py
```diff
@@ -0,0 +1,81 @@
...
+            budgets within the scope).
+        window: Time window for accumulation. None = cumulative (no reset).
+        limit_usd: Maximum USD spend in the window. None = uncapped.
+        limit_tokens: Maximum tokens in the window. None = uncapped.
+    """
+
+    scope: dict[str, str] = Field(default_factory=dict)
+    per: str | None = None
+    window: Literal["daily", "weekly", "monthly"] | None = None
+    limit_usd: float | None = None
```
lan17: It seems like this is a wrong abstraction for budgeting.  What if the budget is in different currency?  Why do we need separate fields for tokens vs usd?

It might be better to do something like an integer for a limit and then define "currency" Enum which could be USD, tokens, Euros, etc.,

I don't think there's a use case for having floating point for USD  or Euros, no?

evaluators/builtin/src/agent_control_evaluators/budget/config.py
```diff
@@ -0,0 +1,81 @@
...
+            value of this metadata field (e.g. "user_id" creates per-user
+            budgets within the scope).
+        window: Time window for accumulation. None = cumulative (no reset).
+        limit_usd: Maximum USD spend in the window. None = uncapped.
+        limit_tokens: Maximum tokens in the window. None = uncapped.
+    """
+
+    scope: dict[str, str] = Field(default_factory=dict)
+    per: str | None = None
+    window: Literal["daily", "weekly", "monthly"] | None = None
```
lan17: What if we want hourly or half-an-hour? 

Maybe its better to define "window" as an integer in seconds, or minutes?  That way you can express whatever window you want.

evaluators/builtin/src/agent_control_evaluators/budget/config.py
```diff
@@ -0,0 +1,81 @@
...
+class BudgetLimitRule(EvaluatorConfig):
+    """A single budget limit rule.
+
+    Each rule defines a ceiling (USD and/or tokens) for a combination
+    of scope dimensions and time window. Multiple rules can apply to
+    the same step -- the evaluator checks all of them and triggers
+    on the first breach.
+
+    Attributes:
+        scope: Static scope dimensions that must match for this rule
```
lan17: Worth it to give some examples here for scope?

evaluators/builtin/src/agent_control_evaluators/budget/config.py
```diff
@@ -0,0 +1,81 @@
...
+        cost_path: Dot-notation path to extract cost from step data.
+        model_path: Dot-notation path to extract model name (for pricing lookup).
+        metadata_paths: Mapping of metadata field name to dot-notation path
+            in step data. Used to extract scope dimensions (channel, user_id, etc).
+    """
+
+    limits: list[BudgetLimitRule] = Field(min_length=1)
+    pricing: dict[str, dict[str, float]] | None = None
+    token_path: str | None = None
+    cost_path: str | None = None
```
lan17: Shouldn't this evaluator be computing cost in USD based on model and token count?  

Doesn't seem like an LLM step should be passing it down here.  I maybe wrong on this.

evaluators/builtin/src/agent_control_evaluators/budget/store.py
```diff
@@ -0,0 +1,235 @@
...
+    """Return max(usd_ratio, token_ratio) clamped to [0.0, 1.0]."""
+    ratios: list[float] = []
+    if limit_usd is not None and limit_usd > 0:
+        ratios.append(min(spent_usd / limit_usd, 1.0))
+    if limit_tokens is not None and limit_tokens > 0:
+        ratios.append(min(spent_tokens / limit_tokens, 1.0))
+    return max(ratios) if ratios else 0.0
+
+
+class InMemoryBudgetStore:
```
lan17: nit: This should go into a separate file so that interface for store and its implementation are separate.

evaluators/builtin/src/agent_control_evaluators/budget/store.py
```diff
@@ -0,0 +1,235 @@
...
+        input_tokens: int,
+        output_tokens: int,
+        cost_usd: float,
+        limit_usd: float | None = None,
+        limit_tokens: int | None = None,
+    ) -> BudgetSnapshot:
+        """Atomically record usage and return current budget state.
+
+        Args:
+            scope_key: Composite scope identifier (e.g. "channel=slack|user_id=u1").
```
lan17: this doc should go into interface docs instead of here

evaluators/builtin/src/agent_control_evaluators/budget/store.py
```diff
@@ -0,0 +1,235 @@
...
+    read-then-write race conditions under concurrent access.
+
+    Built-in: InMemoryBudgetStore (dict + threading.Lock).
+    External: Redis, PostgreSQL, etc. (separate packages).
+    """
+
+    def record_and_check(
+        self,
+        scope_key: str,
+        period_key: str,
```
lan17: Why is this needed to be passed in?  Shouldn't implementation be figuring this out on its own based on current time?

evaluators/builtin/src/agent_control_evaluators/budget/store.py
```diff
@@ -0,0 +1,235 @@
...
+        input_tokens: int,
+        output_tokens: int,
+        cost_usd: float,
+        limit_usd: float | None = None,
```
lan17: Why do we need to pass in limit here?  Can't we instantiate the store with the `BudgetEvaluatorConfig` or something similar so it knows what limits are already?

If we want to share the store between many different kinds of limits and keys, can't we just pass in `BudgetLimitRule` here instead?

evaluators/builtin/tests/budget/test_budget.py
```diff
@@ -0,0 +1,555 @@
...
+    _compute_utilization,
+)
+
+
+# ---------------------------------------------------------------------------
+# InMemoryBudgetStore
+# ---------------------------------------------------------------------------
+
+
+class TestInMemoryBudgetStore:
```
lan17: Tests should follow Given/When/Then behavioral comment style.

evaluators/builtin/src/agent_control_evaluators/budget/config.py
```diff
@@ -0,0 +1,81 @@
...
+
+    Each rule defines a ceiling (USD and/or tokens) for a combination
+    of scope dimensions and time window. Multiple rules can apply to
+    the same step -- the evaluator checks all of them and triggers
+    on the first breach.
+
+    Attributes:
+        scope: Static scope dimensions that must match for this rule
+            to apply. Empty dict = global rule.
+        per: If set, the limit is applied independently for each unique
```
lan17: I don't quite understand this.  Can't we just handle separate budgets by having multiple rules with different `scope` dicts?

lan17 (commented):
Thank you for working on this!  It looks overall in the right direction to me, but I think we can improve the design a bit to make it more universal.

Could you also implement this as a contrib evaluator so it doesn't become a built-in evaluator right away?  We will then put it to use and once its in a production ready stage we can move to builtin.

evaluators/builtin/src/agent_control_evaluators/budget/config.py
amabito: Good point. I'll switch to an integer limit + a Currency enum (USD, EUR, tokens, etc.). Float was unnecessary — cents-level precision is handled by using integer cents anyway.

evaluators/builtin/src/agent_control_evaluators/budget/config.py
amabito: Agreed. Will change to window_seconds: int. I'll keep a few named constants (DAILY = 86400 etc.) as convenience but the field itself will be raw seconds.

evaluators/builtin/src/agent_control_evaluators/budget/config.py
amabito: Sure, will add. Something like {"agent": "summarizer", "channel": "slack"}.

evaluators/builtin/src/agent_control_evaluators/budget/store.py
amabito: Will split into store.py (protocol) and memory_store.py (InMemoryBudgetStore).

evaluators/builtin/src/agent_control_evaluators/budget/store.py
amabito: Yep, will move to the protocol definition.

evaluators/builtin/src/agent_control_evaluators/budget/store.py
amabito: Fair. The store should derive the period from window_seconds + current time internally. I was passing it in for testability but I can inject a clock instead.

evaluators/builtin/src/agent_control_evaluators/budget/config.py
amabito: For static scopes, agreed — multiple rules work. My concern is the dynamic case: "each user gets $10/day" where users aren't known at config time. With per, one rule covers all future users. Without it, you'd need to generate rules on the fly. Would a group_by field work? e.g. group_by: "user_id" means "apply this limit independently per distinct user_id value." Open to other approaches if you have something in mind.

evaluators/builtin/tests/budget/test_budget.py
amabito: Will restructure. Quick example to confirm this is what you mean:
```python
def test_single_record_under_limit(self):
    # Given: store with a $10 daily limit
    # When: record $3 of usage
    # Then: not breached, ratio ~0.3
```

evaluators/builtin/src/agent_control_evaluators/budget/store.py
amabito: Makes sense. I'll have the store accept BudgetLimitRule at registration time so it knows its own limits. record_and_check just takes usage data.

evaluators/builtin/src/agent_control_evaluators/budget/config.py
amabito: My thinking was: the caller already has cost from the LLM response, so passing it avoids maintaining a pricing table. But I see the argument — if the evaluator owns cost computation, the contract is simpler and the caller can't lie about cost. One question: should the evaluator maintain its own pricing table, or pull from an external source (e.g. LiteLLM's model cost map)?

lan17: I believe you already include a pricing table in the evaluator config, so for version 1 of this evaluator we can just rely on that?

amabito: Makes sense. I'll have the evaluator compute cost from the config pricing table + model + token counts. That lets me drop cost_path entirely — just need token_path and a new model_path to extract the model name from the input. If the model isn't in the pricing table, fail closed.

amabito:
@lan17 Hey — update is mostly ready, was waiting on your reply on the cost computation thread before pushing. Got that answer now so I'll push the revision shortly.

evaluators/contrib/budget/src/agent_control_evaluator_budget/budget/evaluator.py
```diff
@@ -0,0 +1,199 @@
...
+    metadata = EvaluatorMetadata(
+        name="budget",
+        version="2.0.0",
+        description="Cumulative LLM token and cost budget tracking",
+    )
+    config_model = BudgetEvaluatorConfig
+
+    def __init__(self, config: BudgetEvaluatorConfig) -> None:
+        super().__init__(config)
+        self._store = InMemoryBudgetStore(rules=config.limits)
```
lan17: i think this runs into a fundamental mismatch with the evaluator framework. evaluator instances are cached globally by (name, config) and reused across requests, so putting InMemoryBudgetStore on self means different controls with the same config will share budget state. that leaks usage across controls/selectors in a way that's going to be very hard to reason about. i don't think the fix is to bake control_id into config, but i do think this needs some unique engine-provided evaluation key / control-instance identity so the store can namespace state outside the cached evaluator instance.

evaluators/contrib/budget/src/agent_control_evaluator_budget/budget/evaluator.py
```diff
@@ -0,0 +1,199 @@
...
+        return 0
+    rates = pricing.get(model)
+    if not rates:
+        return 0
+    input_rate = rates.get("input_per_1k", 0.0)
+    output_rate = rates.get("output_per_1k", 0.0)
+    cost = (input_tokens * input_rate + output_tokens * output_rate) / 1000.0
+    if not math.isfinite(cost) or cost < 0:
+        return 0
+    return math.ceil(cost)
```
lan17: using ceil here means every non-zero call costs at least 1 minor unit. for example, 1 token at 30 cents / 1k becomes 1 whole cent instead of 0.03, so a stream of small calls will burn through budget way too fast. i think we need higher-precision accumulation here and should only round for display / external reporting, not on every call.

evaluators/contrib/budget/src/agent_control_evaluator_budget/budget/evaluator.py
```diff
@@ -0,0 +1,199 @@
...
+
+        input_tokens, output_tokens = _extract_tokens(data, self.config.token_path)
+
+        model: str | None = None
+        if self.config.model_path:
+            val = _extract_by_path(data, self.config.model_path)
+            if val is not None:
+                model = str(val)
+
+        cost = _estimate_cost(model, input_tokens, output_tokens, self.config.pricing)
```
lan17, line 199: i don't think the currency story is coherent yet. we compute a single cost value from the pricing table and then feed that same integer into every matching rule, regardless of whether the rule is USD, EUR, or TOKENS. so EUR rules are effectively checked against USD-denominated spend, and TOKENS doesn't really map to this cost path at all. i'd either constrain this to USD for now or add explicit per-unit accounting / conversion before shipping it.

evaluators/contrib/budget/src/agent_control_evaluator_budget/budget/store.py
```diff
@@ -0,0 +1,67 @@
...
+    spent: int
+    spent_tokens: int
+    limit: int | None
+    limit_tokens: int | None
+    utilization: float
+    exceeded: bool
+
+
+@runtime_checkable
+class BudgetStore(Protocol):
```
lan17: i like adding a BudgetStore protocol, but right now it doesn't feel like a clean abstraction boundary yet. important behavior assumptions like constructor-time rules, clock source, and internal period derivation live outside the protocol, and the evaluator still directly instantiates InMemoryBudgetStore. so in practice a redis/postgres backend still requires changing evaluator construction, not just swapping implementations. i'd either make store injection/provider selection part of the design now or keep this concrete until that boundary is clearer.

lan17 (changes requested):
i think there are still a couple architectural issues to sort out before this is ready. the main ones for me are: budget state currently lives on a cached evaluator instance, per-call rounding materially overstates spend for small requests, and the currency model doesn't line up with what the evaluator actually stores. i also don't think the BudgetStore abstraction is quite a clean swap point yet.

evaluators/contrib/budget/src/agent_control_evaluator_budget/budget/evaluator.py
amabito: Also found that the LRU cache in _factory.py can evict the evaluator instance, which silently destroys the store and resets all counters. So this is worse than I thought.

Moving the store out of the evaluator into a module-level registry keyed by {evaluator_name}:{config_hash} with a threading.Lock.
Evaluator stays stateless, store survives eviction. Identical config = shared pool, which I think is the right default.

Re: engine-provided identity -- agree that's the right long-term direction. circuit_breaker has the same self._store issue. I think an EvaluatorContext param on evaluate() would cover both, but that touches all evaluator signatures so probably better scoped as its own change.

evaluators/contrib/budget/src/agent_control_evaluator_budget/budget/evaluator.py
amabito: Yeah the math is bad. 100 tokens at $0.03/1k = 0.003 cents, ceil makes it 1 cent, 333x overcount.

Switching _estimate_cost() to return float, _Bucket accumulates float, ceil only at snapshot time. Protocol cost param goes from int to float too.

evaluators/contrib/budget/src/agent_control_evaluator_budget/budget/evaluator.py
amabito: Constraining to USD for v1. currency=EUR will raise ValueError.

Token limits are already orthogonal -- limit_tokens works independently in _compute_utilization() and the exceeded check.
TOKENS rules just use limit_tokens, cost path doesn't apply.

evaluators/contrib/budget/src/agent_control_evaluator_budget/budget/store.py
amabito: Keeping concrete for now. Protocol stays as record_and_check only, InMemoryBudgetStore stays as the only implementation. Store injection makes more sense to design alongside the EvaluatorContext work where the store lifecycle gets clearer.

lan17 (commented):
Thanks for sticking with this @amabito — the iterative improvements from R1 through R3 are really solid and this is clearly on the right track. The module-level store registry and float accumulation fixes were exactly the right calls.

I've left a few more comments below, mostly around the API surface (limit model, pricing config, store interface). The biggest ones are probably the limit/limit_unit redesign and making BudgetStore an async ABC. Let me know if any of these are unclear — happy to discuss further.

evaluators/contrib/budget/src/agent_control_evaluator_budget/budget/config.py
```diff
@@ -0,0 +1,98 @@
...
+            None = cumulative (no reset). See WINDOW_* constants.
+        limit: Maximum spend in the window, in cents (USD).
+            None = uncapped on cost.
+        limit_tokens: Maximum tokens in the window. None = uncapped.
+    """
+
+    scope: dict[str, str] = Field(default_factory=dict)
+    group_by: str | None = None
+    window_seconds: int | None = None
+    limit: int | None = None
```
lan17: **Redesign limit model: `limit` + `limit_unit` instead of separate fields**

Consider consolidating `limit` and `limit_tokens` into a single `limit: int` + `limit_unit` design:

```python
limit: int
limit_unit: Literal["usd_cents", "tokens"] = "usd_cents"
```

Benefits:
- One limit per rule, one unit — unambiguous semantics
- Extensible to other currencies later (e.g. `"eur_cents"`) without new fields
- If you need both a cost limit and a token limit on the same scope, write two rules — more explicit than two thresholds on one rule
- Simplifies `_compute_utilization` and the exceeded check — each rule has exactly one threshold
- Makes the unit self-documenting in config, which was a concern from R1

If you need both cost and token limits on the same scope, you write two rules. That's more explicit than one rule with two thresholds and makes the "which limit was breached" logic trivial.

evaluators/contrib/budget/src/agent_control_evaluator_budget/budget/config.py
```diff
@@ -0,0 +1,98 @@
...
+        token_path: Dot-notation path to extract token usage from step
+            data (e.g. "usage.total_tokens"). If None, looks for standard
+            fields (input_tokens, output_tokens, total_tokens, usage).
+        model_path: Dot-notation path to extract model name (for pricing lookup).
+        metadata_paths: Mapping of metadata field name to dot-notation path
+            in step data. Used to extract scope dimensions (channel, user_id, etc).
+    """
+
+    limits: list[BudgetLimitRule] = Field(min_length=1)
+    pricing: dict[str, dict[str, float]] | None = None
```
lan17: **Use a Pydantic model for pricing rates**

The raw nested dict is stringly-typed. A Pydantic model gives you config-time validation and IDE support:

```python
class ModelPricing(EvaluatorConfig):
    input_per_1k: float = 0.0
    output_per_1k: float = 0.0

class BudgetEvaluatorConfig(EvaluatorConfig):
    pricing: dict[str, ModelPricing] | None = None
```

Also: if any rule uses `limit_unit="usd_cents"`, `pricing` should be **required** — validate this at config time rather than silently treating cost as 0 at runtime.

evaluators/contrib/budget/src/agent_control_evaluator_budget/budget/config.py
```diff
@@ -0,0 +1,98 @@
...
+
+    @field_validator("window_seconds")
+    @classmethod
+    def validate_window_seconds(cls, v: int | None) -> int | None:
+        if v is not None and v <= 0:
+            raise ValueError("window_seconds must be positive")
+        return v
+
+
+class BudgetEvaluatorConfig(EvaluatorConfig):
```
lan17, line 71: **Add explicit `budget_id` field to replace config-hash store key**

The current `_config_key` in evaluator.py hashes the serialized config to key the store registry. This means identical configs silently share budget state, and differently-ordered limits get normalized via sorting (but evaluation order still differs).

An explicit required field is cleaner:
```python
budget_id: str  # unique identifier for this budget pool
```

Store key becomes `budget:{budget_id}`. This gives:
- **Explicit sharing/isolation** — controlled by the user, not by config coincidence
- **No sorting ambiguity** — `_config_key` and its sorting logic can be dropped entirely
- **Debuggable** — you can grep logs for the budget_id

evaluators/contrib/budget/src/agent_control_evaluator_budget/budget/evaluator.py
```diff
@@ -0,0 +1,264 @@
...
+    """Estimate cost in cents (USD) from model pricing table.
+
+    Returns a float for precision. Rounding happens at snapshot time,
+    not per call.
+    """
+    if not model or not pricing:
+        return 0.0
+    rates = pricing.get(model)
+    if not rates:
+        return 0.0
```
lan17, line 139: **Unknown model silently gets cost=0, escaping cost-based budget limits**

If a model isn't in the pricing table, cost is silently treated as 0. For a budget enforcement tool this is risky — a new model deployed without updating the pricing config bypasses all cost limits.

Suggestion: add a config option:
```python
unknown_model_behavior: Literal["warn", "block"] = "warn"
```
- `"warn"`: log a warning, treat cost as 0 (current behavior + visibility)
- `"block"`: return `matched=True` — unknown model can't be budgeted, fail closed

At minimum, the current silent fallback should emit a log warning so operators notice the gap.

evaluators/contrib/budget/src/agent_control_evaluator_budget/budget/evaluator.py
```diff
@@ -0,0 +1,264 @@
...
+    Returns (0, 0) if no token information found.
+    """
+    if data is None:
+        return 0, 0
+
+    if token_path:
+        val = _extract_by_path(data, token_path)
+        if isinstance(val, int) and not isinstance(val, bool) and val >= 0:
+            return 0, val
+        if isinstance(val, dict):
```
lan17, line 102: **Nit: token attribution when only total is available**

When `token_path` resolves to a single int, this attributes the entire count to `output_tokens`, returning `(0, total)`. If the pricing table has different input vs output rates, cost will be skewed toward the output rate. Worth a brief comment here noting this is intentional best-effort behavior.

evaluators/contrib/budget/src/agent_control_evaluator_budget/budget/memory_store.py
```diff
@@ -0,0 +1,241 @@
...
+        self,
+        rules: list[BudgetLimitRule],
+        *,
+        clock: Callable[[], float] = time.time,
+        max_buckets: int = _DEFAULT_MAX_BUCKETS,
+    ) -> None:
+        self._rules = rules
+        self._clock = clock
+        self._lock = threading.Lock()
+        self._buckets: dict[tuple[str, str], _Bucket] = {}
```
lan17, line 135: **Stale period buckets accumulate unboundedly**

`_buckets` grows a new entry every time window per scope key. With `window_seconds=3600` and 100 scope keys, that's 2,400 new buckets/day. Old period buckets are never read again but never cleaned up.

The `max_buckets` cap triggers fail-closed but doesn't distinguish stale entries from active ones — a long-running process will eventually hit it from history alone. Consider adding TTL-based cleanup: when creating a new period bucket, prune entries whose period key is older than `2 * window_seconds`. This keeps memory bounded by active scopes rather than history length.

evaluators/contrib/budget/src/agent_control_evaluator_budget/budget/store.py
```diff
@@ -0,0 +1,80 @@
...
+    value never exceeds the actual float. This prevents a contradictory
+    snapshot where spent >= limit but exceeded is False.
+    """
+    if not math.isfinite(value) or value < 0:
+        return 0
+    return int(value)
+
+
+@runtime_checkable
+class BudgetStore(Protocol):
```
lan17: **BudgetStore should be an ABC with async methods**

Two changes:

**1. Protocol → ABC**: The evaluator framework uses ABCs (`Evaluator` is an ABC). Budget stores have shared behavior worth inheriting (input validation, logging), and backends are a finite explicitly-discovered set. ABC is the more consistent and discoverable choice.

**2. Sync → Async**: `record_and_check` is synchronous but called from an async `evaluate`. A Redis or Postgres backend needs async I/O — the current sync interface would force `run_in_executor`. Make this async from the start:

```python
class BudgetStore(ABC):
    @abstractmethod
    async def record_and_check(
        self,
        scope: dict[str, str],
        input_tokens: int,
        output_tokens: int,
        cost: float,
    ) -> list[BudgetSnapshot]: ...
```

For `InMemoryBudgetStore`, the async methods are thin wrappers around the existing sync logic.

**Also**: the `cost` parameter accepts negative values (tested in `test_negative_cost_reduces_spend`), but `_estimate_cost` never produces negative cost. Consider validating `cost >= 0` here, or explicitly documenting when negative cost is expected (credits/refunds?).

evaluators/contrib/budget/src/agent_control_evaluator_budget/budget/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+from agent_control_evaluator_budget.budget.store import BudgetSnapshot, BudgetStore
+
+__all__ = [
+    "BudgetEvaluator",
+    "BudgetEvaluatorConfig",
+    "BudgetSnapshot",
+    "BudgetStore",
+    "InMemoryBudgetStore",
+    "clear_budget_stores",
+]
```
lan17, line 24: `clear_budget_stores()` is a testing utility — remove it from `__all__`. Tests can import it directly from the submodule.

pyproject.toml
```diff
@@ -80,3 +80,9 @@ tag_format = "v{version}"
 # feat = minor, fix/perf/refactor = patch, breaking (!) = major
 allowed_tags = ["feat", "fix", "perf", "chore", "docs", "style", "refactor", "test", "ci"]
 patch_tags = ["fix", "perf", "chore", "refactor"]
+
+[dependency-groups]
```
lan17: This adds `[dependency-groups]` to the **root** `pyproject.toml`. The budget evaluator already has its own dev dependencies in its own `pyproject.toml`. This root-level change appears unrelated — please revert.

evaluators/contrib/budget/README.md
```diff
@@ -0,0 +1,3 @@
+# Budget Evaluator
```
lan17: The README needs more content before merge. At minimum:
- A complete config example (limits, pricing, metadata_paths)
- Explanation of scope and group_by semantics
- How the pricing table works
- **Important caveat**: `InMemoryBudgetStore` is single-process only — budget state is not shared across workers/pods and is lost on restart. Users need to understand this limitation before relying on it for enforcement.

evaluators/contrib/budget/src/agent_control_evaluator_budget/budget/config.py
amabito: looked at memory_store.py:60-72 and 159-166 -- `limit` and `limit_tokens` are evaluated independently within a single rule, both feeding `utilization = max(cost_ratio, token_ratio)` on one shared bucket. splitting to two rules is not a semantic drop-in: it changes the snapshot shape and `breached_rules` metadata structure.

one alternative that keeps the readability goal without losing the OR-breach semantic: rename `limit` -> `limit_cents` (explicit unit in the name) and keep `limit_tokens` as-is. ambiguity gone, no structural change.

happy to go either way, but wanted to flag this before ripping out the dual-limit path. which direction do you want?

evaluators/contrib/budget/src/agent_control_evaluator_budget/budget/config.py
amabito: agreed. current `rates.get("input_per_1k", 0.0)` silently zeroes on typos -- config-time validation catches that.

the "required when usd_cents rules exist" validator depends on the limit model redesign landing first (it needs something to check against). so I'll sequence this after the `limit` change.

evaluators/contrib/budget/src/agent_control_evaluator_budget/budget/config.py
amabito: on board with dropping `_config_key` entirely. the sort-normalization in there is a workaround for the hash approach, not load-bearing, and the three `TestStoreRegistry` tests that lock in the current implicit-sharing semantic can be rewritten around explicit ids.

one thing to pin down before I rip it out: required or optional with a default? optional-with-uuid-default reintroduces implicit keying through the back door (anyone who forgets to set it gets a per-instance unique pool, no sharing). required is a harder break but the semantic is clean. which way?

evaluators/contrib/budget/src/agent_control_evaluator_budget/budget/evaluator.py
amabito: agreed this is a fail-open hole. one pushback on the default though: this is a budget enforcer, not a monitoring tool, so `warn` means a new model silently undercounts cost while traffic keeps flowing. for a tool whose job is to stop spend, fail-closed feels like the right default.

proposal: default to `block`, but only trigger when (a) the matched rule has a cost-based limit set and (b) `pricing` is non-null. token-only rules and configs without any pricing table stay unaffected, so existing token-only users don't get broken by the change.

ok with `block` as default, or do you want `warn` kept for backcompat reasons?

evaluators/contrib/budget/src/agent_control_evaluator_budget/budget/evaluator.py
amabito: yeah, it's a deliberate conservative choice -- output rates are usually higher than input, so all-to-output over-estimates rather than under-estimates. will add a one-line comment on 112 noting the intent.

evaluators/contrib/budget/src/agent_control_evaluator_budget/budget/memory_store.py
amabito: good catch. plan is a rollover-scan TTL prune:

- `_parse_period_key("P86400:19800")` -> `(86400, 19800)`
- `_last_pruned_period: dict[int, int]` tracks last pruned index per window size
- on new bucket creation with non-empty period key, if `_last_pruned_period[window] != current_index`, scan `_buckets`, drop entries with same window and `bucket_index < current - 1`, update the tracker

O(n) scan but amortized O(1) per record, and under expected scope counts (<100) the scan is microseconds in Python. a heap adds invalidation complexity because of `reset()` and isn't worth it at this scale.

`max_buckets` stays as a hard backstop for high-cardinality `group_by` explosions (e.g. per-user keys) that TTL can't protect against -- will document this in the store docstring.

evaluators/contrib/budget/src/agent_control_evaluator_budget/budget/store.py
amabito: on Protocol -> async ABC: agreed. `Evaluator` is already an ABC so this matches the framework shape. keeping `threading.Lock` internally -- the critical section has no `await`, so the public `async def record_and_check` wraps a sync helper under the existing lock. that keeps `test_thread_safety` (test_budget.py:165) intact and doesn't require reworking InMemoryBudgetStore's concurrency model.

on `cost >= 0` validation: pushback. `test_negative_cost_reduces_spend` (test_budget.py:655) is intentional refund semantic. `round_spent()` (store.py:47) already floors the displayed `spent` to 0, but the internal accumulator can go negative so a subsequent positive charge cancels correctly. and `_estimate_cost()` (evaluator.py:154-155) already guards `cost < 0 -> 0.0`, so negative cost can't reach the store through the normal evaluate path today -- it's only reachable through direct store injection (used by that one test).

plan: document the refund semantic in the ABC docstring, no runtime validation. if you'd rather remove the refund path entirely that's a bigger call and I'd want your go-ahead before doing it.

evaluators/contrib/budget/src/agent_control_evaluator_budget/budget/__init__.py
amabito: yeah, will drop from `__all__`. `from agent_control_evaluator_budget.budget.evaluator import clear_budget_stores` still works for tests, so no test breakage.

pyproject.toml
amabito: reverting the `[dependency-groups]` block. also noticed this commit accidentally regressed `version = "7.3.1"` to `"7.3.0"` in the same file -- will restore that too.

evaluators/contrib/budget/README.md
amabito: current file is ~3 lines, yeah. rewrite plan for R4:

- install (`pip install agent-control-evaluator-budget`)
- quickstart: minimal yaml config + python snippet that actually runs
- config reference: all `BudgetEvaluatorConfig` fields, types, defaults
- `scope` semantics (exact dict match, empty dict = global bucket)
- `group_by` (per-value sub-buckets, one per unique value)
- pricing table format and unit (cents)
- **critical caveat**: `InMemoryBudgetStore` is single-process only, no persistence, state lost on restart. not suitable for multi-worker deploys -- swap in a Redis/Postgres backend for that.
- window constants reference (`WINDOW_HOURLY/DAILY/WEEKLY/MONTHLY`)
- `clear_budget_stores()` note: testing utility, don't call in production

landing with R4.

amabito:
@lan17 pulling three open questions out of the inline threads so they don't get lost:

1. **limit model shape.** splitting `limit` + `limit_tokens` into two separate rules loses the single-rule dual-ceiling semantic (where both feed `max(cost_ratio, token_ratio)` on the same bucket, and `breached_rules` metadata is one entry). alternative that keeps the semantic: rename `limit` -> `limit_cents` and leave `limit_tokens` in place. which do you prefer?

2. **`budget_id`: required or optional?** required gives a clean explicit-sharing semantic but breaks existing configs. optional with a uuid default is non-breaking but any user who forgets to set it gets a unique per-instance pool, which is basically the hash approach through the back door. leaning required, want to confirm.

3. **`unknown_model_behavior` default.** `warn` is fail-open -- a new model silently undercounts cost while traffic flows through. for a budget enforcer, `block` (scoped to cost-limit rules with non-null pricing, so token-only users aren't affected) feels like the right fail-closed default. ok with `block`?

will hold the config-layer work (limit shape, `budget_id`, `ModelPricing` validator, unknown-model default) until these land. in the meantime planning to move forward on the mechanical items: root `pyproject.toml` revert, `clear_budget_stores` out of `__all__`, token attribution comment, async ABC + `threading.Lock` wrapping for `BudgetStore`, stale period bucket TTL prune, and README rewrite. let me know if any of that should wait too.

amabito:
@lan17 pushed R4 (`b189e85`). here's what landed and what's still waiting on your answers from the inline threads.

**landed in R4:**
- `BudgetStore` is now an async ABC. `InMemoryBudgetStore` wraps a sync helper under `threading.Lock` -- same concurrency model, just honest about the interface for future Redis/Postgres backends.
- `__init_subclass__` walks the MRO and rejects sync overrides of `record_and_check` at class creation time, including mixin-inherited ones.
- TTL prune for stale period buckets. runs before the `max_buckets` capacity check on rollover so long-running processes don't permanently fail-closed from history alone. `max_buckets` stays as a backstop for high-cardinality `group_by` explosions.
- defensive guards: NaN/Inf cost and clock coerced to 0.0, negative token counts clamped to 0, utilization clamped to [0.0, 1.0] for refund scenarios.
- root `pyproject.toml` reverted (`version` back to 7.3.1, `[dependency-groups]` removed).
- `clear_budget_stores` dropped from `__all__` (still importable from the evaluator submodule for tests).
- token attribution comment added.

67 -> 91 tests. all pass, ruff clean.

**waiting on your call (3 open questions from the earlier comment):**
1. limit model shape (`limit_cents` + `limit_tokens` vs `limit` + `limit_unit`)
2. `budget_id` required or optional
3. `unknown_model_behavior` default (`block` vs `warn`)

README rewrite is staged but holding until the config shape is finalized so the examples are accurate. lmk if any of the R4 changes need adjustment.

amabito:
also wanted to say -- the R1 through R3 review rounds genuinely improved the design. the store registry namespace issue and the float accumulation fix from R2 were both real production bugs, and the async ABC + TTL prune from R3 set up the interface correctly for distributed backends from the start. this is way better than what I would have shipped on the first pass. appreciate the thorough reviews.

evaluators/contrib/budget/src/agent_control_evaluator_budget/budget/config.py
lan17: I think single limit is fine.  To implement dual limit we could just easily have 2 budget evaluators with different configs, right?

lan17: fair point about the OR-breach semantic. i still think single-limit-per-rule is the cleaner design long term — two rules with the same scope share a bucket anyway, so the only real difference is metadata shape. but i hear you that it's a structural change with downstream effects on snapshot consumers.

let's go with `limit` + `limit_unit: Literal["usd_cents", "tokens"] = "usd_cents"` for the cleaner extensibility story. if someone needs both cost and token enforcement on the same scope they write two rules — the metadata shows two `breached_rules` entries instead of one, which is arguably more debuggable anyway.

evaluators/contrib/budget/src/agent_control_evaluator_budget/budget/config.py
lan17: required with a default like `budget_id: str = "default"` sounds right. covers the single-budget case without friction and avoids the uuid back door.

just make sure the docs are clear that two evaluator instances with the same `budget_id` share a budget pool (accumulated spend, snapshots, everything). different `budget_id` = fully isolated.

evaluators/contrib/budget/src/agent_control_evaluator_budget/budget/evaluator.py
lan17: agreed, `block` as default makes more sense for a budget enforcer. your scoping is right too — only trigger when the rule is cost-based and `pricing` is configured. token-only rules stay unaffected.

good thinking on this one. the whole R4 revision is solid work — the TTL prune design, the async ABC with MRO enforcement, and the defensive guards are all well thought through. appreciate the back and forth on these design questions.

amabito:
## R5 revision

This revision addresses the three design questions from @lan17 and implements Phase C+D in full.

### Answers implemented

**Q1 -- limit model**: Adopted `limit: int` + `limit_unit: Literal["usd_cents", "tokens"] = "usd_cents"`. The dual-ceiling field pair (`limit` + `limit_tokens`) is gone. Two rules with the same scope and different units express dual enforcement, with two separate `breached_rules` entries rather than one ambiguous rule object. `_compute_utilization` is now a single ratio.

**Q2 -- budget_id**: Added `budget_id: str = "default"`. The store registry is now keyed by `f"budget:{budget_id}"` (no more config hash, no more sort). The `_config_key` function is removed. Same `budget_id` shares accumulated bucket state; different `budget_id` is fully isolated. Documented in README and code comments.

**Q3 -- unknown_model_behavior**: Added `unknown_model_behavior: Literal["block", "warn"] = "block"`. The block path triggers only when the rule is cost-based (`limit_unit == "usd_cents"`) AND `pricing` is configured AND the model is absent from the pricing table. Token-only rules are unaffected. Validated by four dedicated tests including a case-sensitivity test.

### Phase C changes

- **C1**: `BudgetLimitRule` now has `limit: int` (required) + `limit_unit: Literal["usd_cents", "tokens"] = "usd_cents"`. `limit_tokens` removed. `at_least_one_limit` validator removed.
- **C2**: `budget_id: str = "default"` added to `BudgetEvaluatorConfig`.
- **C3**: `_config_key` removed. Store registry key is `f"budget:{config.budget_id}"`.
- **C4**: `ModelPricing(EvaluatorConfig)` with `input_per_1k: float = 0.0` and `output_per_1k: float = 0.0`. `pricing` field is `dict[str, ModelPricing] | None`. Config-time validator rejects `pricing=None` when any rule uses `limit_unit="usd_cents"`.
- **C5**: Store contract redesign: `InMemoryBudgetStore.__init__` no longer takes `rules`. Rules are passed per `record_and_check(rules, scope, ...)` call. This means two evaluators with the same `budget_id` share bucket state but each evaluates with its own rule set -- no silent rule override.

### Phase D changes

- **D1**: `unknown_model_behavior: Literal["block", "warn"] = "block"` added to `BudgetEvaluatorConfig`.
- **D2**: Unknown model detection in `evaluate()` is gated on: cost rule present AND pricing configured AND model not in pricing. Token-only evaluators are not affected.
- **D3**: README rewritten with full config example, scope/group_by explanation, budget pools section (shared buckets vs own rules), pricing section, dual-ceiling pattern, and single-process-only caveat.

### Additional inline comments addressed

- `clear_budget_stores` removed from `__all__` (testing utility, import directly from evaluator submodule).
- Root `pyproject.toml` `[dependency-groups]` addition reverted.
- `BudgetSnapshot.limit_tokens` removed; `limit_unit` added.
- Token attribution comment added to `_extract_tokens` (single-int path attributes to output).
- TTL prune was already in R4 (period rollover drops stale buckets before max_buckets check).
- `BudgetLimitRule` added to `__init__.__all__`.

### Rebase

Rebased onto `origin/main` (v7.4.0). Conflict in `pyproject.toml` (version string) resolved by keeping main's `7.4.0`. Unrelated `[dependency-groups]` addition in root `pyproject.toml` reverted per reviewer request.

### Tests

100 tests passing (91 in R4). New test classes: `TestModelPricing`, `TestBudgetIdSemantics`, `TestUnknownModelBehavior`. All tests use Given/When/Then comment style.

evaluators/contrib/budget/src/agent_control_evaluator_budget/budget/config.py
```diff
@@ -0,0 +1,115 @@
...
+            "rule exists. block=fail closed, warn=log warning and treat cost as 0."
+        ),
+    )
+    pricing: dict[str, ModelPricing] | None = None
+    token_path: str | None = None
+    model_path: str | None = None
+    metadata_paths: dict[str, str] = Field(default_factory=dict)
+
+    @model_validator(mode="after")
+    def require_pricing_for_cost_rules(self) -> "BudgetEvaluatorConfig":
```
lan17, line 112: this validator now requires pricing for `usd_cents` rules, but it still doesn't require `model_path`, and `evaluate()` only extracts the model when `model_path` is set. so a cost-based config can look valid and still silently compute `cost=0` forever. i reproduced this locally with a `limit=1` cost rule plus pricing but no `model_path`; it returns within budget with `cost: 0.0`. i think cost-based rules need either a required `model_path` or a documented default model lookup.

evaluators/contrib/budget/src/agent_control_evaluator_budget/budget/evaluator.py
```diff
@@ -0,0 +1,273 @@
...
+
+        model: str | None = None
+        if self.config.model_path:
+            val = _extract_by_path(data, self.config.model_path)
+            if val is not None:
+                model = str(val)
+
+        cost = _estimate_cost(model, input_tokens, output_tokens, self.config.pricing)
+        model_known = model is None or self.config.pricing is None or model in self.config.pricing
+        has_cost_rule = any(rule.limit_unit == "usd_cents" for rule in self.config.limits)
```
lan17: this unknown-model gate is still a bit too broad. `has_cost_rule` is computed over the whole config before scope matching, so if one evaluator contains a token-only rule for scope A and a cost rule for scope B, an unknown-model event in scope A still gets blocked even though only the token rule applies. i think this needs to consider only matching cost rules for the current extracted scope.

lan17 (changes requested):
this has improved a lot from the earlier rounds. moving the state out of cached evaluator instances, cleaning up the limit/unit model, tightening the store contract, and expanding the tests all pushed this in the right direction. thank you for sticking with the iteration here.

i think there are just two remaining blockers for me. once those are addressed, this should be good to go.

amabito:
@lan17 pushed R6.

**1. model_path required for cost rules**

model_validator now rejects configs with cost rules when model_path is missing or whitespace-only. closes the silent cost=0 hole you reproduced.

also fixed a subtler issue on the runtime side -- previously model=None short-circuited model_known=True regardless of *why* it was None. so if model_path was configured but the field was just missing from step data, the unknown_model_behavior gate got skipped entirely and cost=0 was recorded. now the evaluator tracks whether model_path was configured separately, and treats "configured but extraction returned None" as unknown-model.

**2. scoped unknown-model gate**

has_cost_rule check now runs _scope_matches against extracted step metadata before blocking. same function the store uses internally, so the gate and the store agree on what matches. fixes the cross-scope false positive you described.

100 -> 118 tests. covers model_path validator edge cases, runtime extraction failure paths, scoped unknown-model with warn mode, empty pricing table, mixed rules, message format pinning.

lan17 (approved):
thanks for sticking with this. you've done a really solid job iterating on the design here, and the final shape is a lot stronger because of it.

moving the budget state out of cached evaluator instances, tightening the config/model-path behavior, cleaning up the limit/unit model, and expanding the tests all pushed this in the right direction. i really appreciate how thoughtfully you worked through the back and forth on the design questions instead of just papering over them.

good work here — this looks ready from my side.

amabito:
@lan17 thanks. the back and forth made this land. the r6 fixes only
existed because you actually reproduced the silent cost=0 path instead of
letting it slide -- same with the limit/unit split and the model_path
gate. i wasn't going to catch those on my own.

i've been eyeing circuit breaker and tool policy middleware as follow-ups
if there's room on the containment roadmap. ping me when you want another
pass.

PR #145, opened 2026-03-23
feat(sdk): [Enterprise Integration]: Add provider agnostic traceing

namrataghadi-galileo:
**Summary**
Added a new provider-agnostic telemetry package to the AgentControl Python SDK for external trace context resolution and merged control event emission.
Updated tracing to consult a registered external trace context provider before falling back to OTEL context.
Exported the new telemetry APIs from the top-level agent_control package.
Added focused tests to ensure provider/sink failures do not affect existing behavior.

**Scope**
- User-facing/API changes:
   - New SDK APIs:

      1. set_trace_context_provider(...)
      2. get_trace_context_from_provider()
      3. clear_trace_context_provider()
      4. set_control_event_sink(...)
      5. emit_control_events(...)
      6. clear_control_event_sink()

- Internal changes:
   - Added:

     1. agent_control/telemetry/trace_context.py
     2. agent_control/telemetry/event_sink.py
     3. agent_control/telemetry/__init__.py
     4. Updated tracing.py to use external provider before OTEL fallback.
     5. Added tests for provider behavior, sink behavior, and tracing precedence.

**Risk and Rollout**
**Risk level:** low
**Rollback plan:**
Revert the new telemetry package and the small tracing/export changes in the SDK.
Since the change is additive and inactive unless a provider or sink is explicitly registered, rollback is straightforward.

**Testing**

-  Added or updated automated tests
-  Ran make check (or explained why not)
- Ran targeted SDK tests only; did not run the full project check suite.
-  Manually verified behavior

**Checklist**

-  Linked issue/spec (if applicable)
-  Updated docs/examples for user-facing changes
-  Included any required follow-up tasks


lan17:
> @lan17 Keeping emission in the SDK gives us one merged, ordered batch and a single integration model for both logger and non-logger flows

What about other language sdks, like typescript, java, etc.?

If we can delegate this to agent control server then it makes sdk easier.  

Could we just send additional metadata to agent control server when we send events to integrate with Galileo spans/traces?

namrataghadi-galileo:
Server-side emission is attractive for thinner SDKs, but it defeats the main purpose of the logger-based design: reusing Galileo Logger’s in-process trace buffering and flush. For logger integrations, the SDK must have the merged control events locally before flush. Other languages can still be supported through the thinner OTEL non-logger path until they need full logger integration.

sdks/python/src/agent_control/telemetry/event_sink.py
```diff
@@ -0,0 +1,33 @@
...
+_control_event_sink: ControlEventSink | None = None
+
+
+def set_control_event_sink(sink: ControlEventSink | None) -> None:
+    """Register a sink for merged control execution events."""
+    global _control_event_sink
+    _control_event_sink = sink
+
+
+def emit_control_events(events: list[ControlExecutionEvent]) -> None:
```
lan17: This adds a sink registration API, but no SDK evaluation path ever calls `emit_control_events()`. Local results still flow through `evaluation._emit_local_events()` into `observability.add_event()`, and server results are only merged as `EvaluationResult`. As written, `set_control_event_sink()` does not observe real control executions unless callers manually construct `ControlExecutionEvent` lists themselves.

sdks/python/src/agent_control/telemetry/trace_context.py
```diff
@@ -0,0 +1,53 @@
...
+    except Exception:
+        # Provider failures should not break control evaluation.
+        return None
+
+    if trace_context is None:
+        return None
+
+    trace_id = trace_context.get("trace_id")
+    span_id = trace_context.get("span_id")
+    if not isinstance(trace_id, str) or not isinstance(span_id, str):
```
lan17, line 41: This validates types, but it still accepts empty strings. A provider returning `{"trace_id": "", "span_id": ""}` currently passes through as valid context, and the tracing helpers will return those empty IDs unchanged. That can silently drop headers via truthiness checks or produce uncorrelated local events; these values should be rejected here.

sdks/python/src/agent_control/tracing.py
```diff
@@ -114,6 +117,11 @@ def get_trace_and_span_ids() -> tuple[str, str]:
     if trace_id is not None and span_id is not None:
         return trace_id, span_id
 
+    # Try external provider
+    trace_context = get_trace_context_from_provider()
```
lan17, line 121: This wires the provider into the tracing helpers, but the public `evaluate_controls()` path still forwards `trace_id` / `span_id` unchanged when callers omit them. So the new provider works for flows that use these helpers, but not for all public SDK evaluation entry points. That inconsistency should be fixed before we rely on this as the generic trace-context hook.

lan17 (changes requested):
I like the attempt to keep the new telemetry surface generic, but there are still a few code-level gaps before this is ready to merge. I left inline comments on the sink wiring, provider validation, and the partial integration of the provider across SDK entry points.

sdks/python/src/agent_control/telemetry/event_sink.py
namrataghadi-galileo: I am handling that in another Ticket: https://app.shortcut.com/galileo/story/59787/2-add-merged-control-event-data-flow-to-sdk?vc_group_by=day&ct_workflow=all&cf_workflow=500018000

namrataghadi-galileo: The goal was to just add the tracing interfaces in this PR and do heavy lifting in another PR else this PR would have been big and difficult to review

sdks/python/src/agent_control/tracing.py
namrataghadi-galileo: Good catch. You’re right that just wiring the provider into tracing.py is not enough if the public evaluation path can still bypass it when trace_id / span_id are omitted. I’ll fix that

sdks/python/src/agent_control/telemetry/trace_context.py
namrataghadi-galileo: done

sdks/python/src/agent_control/telemetry/event_sink.py
lan17: That makes sense as a scoping decision, but I still think the code in this PR should match that scope. Right now the PR exports a public sink API, yet no real evaluation path uses it, so the merged-event behavior is still absent at merge time. If the merged control-event flow is intentionally deferred to the follow-up ticket, then I’d prefer to drop the sink API from this PR and keep this change limited to the trace-context hook. Otherwise, I think the sink needs to be wired into an actual evaluation path before merge.

namrataghadi-galileo: okay. I will add it in my next PR

lan17 (approved):

PR #150, opened 2026-03-26
fix(sdk): use sync shutdown flush fallback

abhinav-galileo:
## Summary
- replace the atexit shutdown fallback that spawned a helper thread and `asyncio.run()` with a synchronous `httpx.Client` flush path
- reuse shared batch request construction for both async and sync send paths and clear the async client reference after fallback shutdown cleanup
- add regression coverage for the sync fallback path and for a short-lived subprocess that emits an SDK-evaluated observability event and exits without explicit shutdown

## Testing
- make sync
- make check
- git push -u origin abhi/observability-sync-shutdown-flush (pre-push hook: lint, typecheck, contrib evaluator checks)

### DIFF
--- a/sdks/python/src/agent_control/observability.py
+++ b/sdks/python/src/agent_control/observability.py
@@ -448,7 +448,7 @@ class EventBatcher:
                 logger.error("Error sending events during shutdown: %s", e)
 
             if attempt >= get_settings().max_retries - 1:
-                continue
+                break
 
             retry_delay = get_settings().retry_delay * (attempt + 1)
             if deadline is not None:
--- a/sdks/python/src/agent_control/observability.py
+++ b/sdks/python/src/agent_control/observability.py
@@ -48,6 +48,7 @@ import asyncio
 import atexit
 import logging
 import threading
+import time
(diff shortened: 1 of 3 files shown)


sdks/python/src/agent_control/observability.py
```diff
@@ -388,29 +389,142 @@ def _stop_worker(self, *, graceful: bool, join_timeout: float) -> bool:
...
-        )
-        helper_thread.start()
-        helper_thread.join(timeout=timeout)
-        if helper_thread.is_alive():
-            logger.warning("Fallback shutdown flush timed out after %.1f seconds", timeout)
-            return
-        if flush_error is not None:
-            logger.error("Error during fallback shutdown flush: %s", flush_error)
+            if attempt >= get_settings().max_retries - 1:
+                continue
```
namrataghadi-galileo: nit: `break` instead of `continue`.  It keeps the same behavior but is less awkward :D

namrataghadi-galileo (approved):

lan17 (approved):

PR #155, opened 2026-03-31
feat(sdk): default merge events in SDK

namrataghadi-galileo:
## Summary

This PR makes the SDK the single owner of `ControlExecutionEvent` creation and emission, moving all observability event reconstruction from the server to the SDK. The `/api/v1/evaluation` endpoint now returns only evaluation semantics, keeping evaluation and observability concerns cleanly separated.

**Key changes:**
- SDK reconstructs both local and server control execution events after receiving the lightweight `EvaluationResponse`
- Server `/evaluation` endpoint is now evaluation-only — no longer builds or ingests observability events
- Events are emitted exclusively through the SDK's existing observability pipeline (batcher → `/api/v1/observability/events` → OSS storage)
- Event identity consistency preserved using `ControlDefinition.observability_identity()` for composite conditions

### When SDK observability is enabled
- `check_evaluation_with_local()` reconstructs local + server events in the SDK and enqueues one combined batch
- `check_evaluation()` reconstructs server events in the SDK and enqueues through the built-in pipeline
- Trace/span correlation preserved through the tracing resolver and reconstructed control metadata

### When SDK observability is disabled
- Evaluation still works normally
- No control-execution events are created or emitted

### Error handling
- If local controls succeed but the server request/parsing fails, local events are still enqueued before the error is re-raised
- Safe failure handling ensures observability data isn't lost

## What changed

**SDK (`sdks/python/src/agent_control/`):**
- Added shared event reconstruction helpers in `evaluation_events.py`
- Updated `evaluation.py` to support:
  - SDK reconstruction of local events
  - SDK reconstruction of server events from lightweight `EvaluationResponse` + cached control definitions
  - Combined enqueue of local + server events through existing SDK batcher
  - Standalone `check_evaluation()` reconstruction when observability enabled
- Removed old session-scoped `merge_events` mode from Python SDK surface

**Server (`server/src/agent_control_server/`):**
- Updated `endpoints/evaluation.py` to be evaluation-only
- Returns sanitized `EvaluationResponse` without observability payloads
- Removed observability event emission and ingestion from `/evaluation` endpoint

**TypeScript client:**
- Cleaned up generated client drift to match current pure-evaluation contract

## Testing

Tests updated to cover:
- ✅ Default local event enqueue behavior
- ✅ SDK reconstruction of server events
- ✅ Local + server combined enqueue behavior
- ✅ Provider-backed trace/span propagation
- ✅ Local-event preservation when server call/parsing fails
- ✅ Pure `/evaluation` behavior with no server-side observability emission
- ✅ Response sanitization and helper behavior on server side

## Migration notes

- This is now the **default observability behavior** (not an optional mode)
- Direct callers to `/api/v1/evaluation` no longer get observability automatically unless they also emit events through `/api/v1/observability/events`


sdks/python/src/agent_control/evaluation.py
```diff
@@ -255,41 +167,64 @@ def _merge_results(
     )
 
 
+async def check_evaluation(
```
lan17, line 150: This still leaves the merged-event path only partially integrated across the public SDK surface. If a caller registers `set_control_event_sink()` but uses the public `check_evaluation()` helper, we just POST and return the parsed result without setting `X-Agent-Control-Merge-Events` or reconstructing/emitting any events. I think we need to either wire the sink behavior through this helper too, or explicitly scope the sink API away from this entry point.

sdks/python/src/agent_control/evaluation.py
```diff
@@ -392,27 +316,26 @@ def _with_parse_errors(result: EvaluationResult) -> EvaluationResult:
     if applicable_local_controls:
         engine = ControlEngine(applicable_local_controls, context="sdk")
         local_result = await engine.process(request)
-
-        _emit_local_events(
+        local_control_lookup = {
+            control.id: control.control for control in applicable_local_controls
+        }
+        local_events = build_control_execution_events(
```
lan17: Can we avoid reconstructing events unless something will actually consume them? We now always build local events here, and we do the same for server events later, even when merged emission is off and even when OSS observability is disabled. Since evaluation is a latency-sensitive hot path, I don't think we should pay this parsing/allocation cost unless either the existing observability path is active or a merged-event sink is registered.

lan17 (changes requested):
I think there are still two blockers before we merge this. I left inline notes on the partial sink integration and the unconditional reconstruction work on the evaluation hot path.

sdks/python/src/agent_control/evaluation.py
namrataghadi-galileo: sure .

sdks/python/src/agent_control/evaluation.py
namrataghadi-galileo: Good catch — you’re right. check_evaluation() currently bypasses the merged-event path even when a control event sink is registered, which makes the public SDK surface inconsistent. I think the right fix is to wire the sink behavior through this helper as well: when a sink is present, resolve trace/span IDs, send X-Agent-Control-Merge-Events: true, reconstruct the server events from the lightweight response plus cached control definitions, and emit them through the sink before returning the parsed result. That keeps the default behavior unchanged while making merged-event handling consistent across both public evaluation entry points.

sdks/python/src/agent_control/evaluation.py
namrataghadi-galileo: I’m tightening the hot path so we only reconstruct events when they’ll actually be consumed: local events only when the default SDK observability path is enabled or a sink is registered, and server events only when the merged-event path is active.

sdks/python/src/agent_control/evaluation.py
namrataghadi-galileo: This comment is no longer valid. I have re-worked this PR to only contain merge model support

lan17 (changes requested):
Blocking issue:

- `server/src/agent_control_server/endpoints/evaluation.py:242-263` now trusts `X-Agent-Control-Merge-Events` from the caller and skips server-side observability ingestion solely on that header. A direct API caller can set the header and suppress control-execution events without reconstructing/re-emitting them anywhere. The server needs to treat merged mode as a trusted SDK flow, or otherwise keep the default ingestion path for untrusted callers.

sdks/python/src/agent_control/evaluation.py
```diff
@@ -392,27 +401,27 @@ def _with_parse_errors(result: EvaluationResult) -> EvaluationResult:
...
-        _emit_local_events(
-            local_result,
-            request,
-            applicable_local_controls,
-            resolved_trace_id,
-            resolved_span_id,
-            agent_name=event_agent_name,
-        )
+            if not merged_emission_enabled:
+                enqueue_observability_events(local_events)
```
abhinav-galileo: When `merged_emission_enabled` is true, `local_events` are built but not enqueued here. If the subsequent server POST raises (timeout, connection error, HTTP error), the function exits without ever calling `enqueue_observability_events(local_events)` — the local controls ran successfully but their observability records are silently lost.

Consider wrapping the server call in `try/except` and enqueuing `local_events` before re-raising, so locally-evaluated controls always produce observability records regardless of server availability.

sdks/python/src/agent_control/evaluation_events.py
```diff
@@ -0,0 +1,210 @@
...
+                control_name=match.control_name,
+                check_stage=request.stage,
+                applies_to=applies_to,
+                action=match.action,
+                matched=matched,
+                confidence=match.result.confidence,
+                timestamp=now,
+                evaluator_name=evaluator_name,
+                selector_path=selector_path,
+                error_message=match.result.error if not matched else None,
```
abhinav-galileo: This expression uses a single `matched` boolean for all three categories (matches, errors, non_matches), but the server-side `_build_observability_events` handles them differently: matches keep `result.error`, non-matches force `None`. When merge mode reconstructs server events through this helper, the `error_message` values will diverge from what the server would have produced.

This existed before for local-only events, but merge mode now routes server events through this path too, making parity with the server more important. Consider passing error_message per-category to match server behavior.

sdks/python/src/agent_control/telemetry/trace_context.py
```diff
@@ -38,8 +38,6 @@ def get_trace_context_from_provider() -> TraceContext | None:
 
     trace_id = trace_context.get("trace_id")
     span_id = trace_context.get("span_id")
-    if not isinstance(trace_id, str) or not isinstance(span_id, str):
-        return None
```
abhinav-galileo, line 42: The `isinstance(trace_id, str)` / `isinstance(span_id, str)` guard was removed here. `TypedDict` provides no runtime enforcement, so a buggy provider returning e.g. `int` or `bytes` now passes the truthiness check and flows into HTTP headers and event fields that expect strings.

The removal looks incidental to the PR's goals — consider restoring it as a low-cost defensive check.

sdks/python/src/agent_control/evaluation.py
namrataghadi-galileo: Thanks . Implemented. In the merged local+server path, check_evaluation_with_local() now preserves SDK-local observability even if the subsequent server call fails. The server POST is wrapped in try/except, and if local controls already produced local_events, those events are enqueued before the exception is re-raised. That keeps the original failure behavior for evaluation while avoiding silent loss of local control records.

I also added a regression test covering the case where merged mode is enabled, local evaluation succeeds, the server POST raises, and only the reconstructed local events are enqueued before the error bubbles up.

sdks/python/src/agent_control/evaluation_events.py
namrataghadi-galileo: Fixed. The shared SDK event builder now matches the server’s category-specific error_message behavior: matches and errors keep result.error, while non_matches always set error_message=None, so merged-mode reconstruction stays aligned with the server path.

sdks/python/src/agent_control/telemetry/trace_context.py
namrataghadi-galileo: Fixed. I restored the defensive runtime type check in trace_context.py, so providers returning non-string trace_id or span_id are rejected again before those values can flow into headers or event fields.

lan17 (changes requested):
Re-reviewed current head `9af1755`.

The original header-trust issue looks fixed, but there is still one blocker:

- `server/src/agent_control_server/merge_event_sessions.py:28-40` scopes merge-enable state by `(client.api_key, agent_name)`. For session-cookie auth, `_authenticate_via_cookie()` returns `AuthenticatedClient(api_key="", ...)` in `server/src/agent_control_server/auth.py:118-121`, so every cookie-authenticated browser session collapses to the same key `("", agent_name)`. If one logged-in UI session initializes an agent with merge events enabled, any other logged-in UI session can send `X-Agent-Control-Merge-Events: true` and suppress server-side ingestion for that agent. The trusted-session key needs to use a per-session or otherwise user-specific identifier, not the empty `api_key` sentinel.

lan17 (approved):

PR #164, opened 2026-04-05
feat(sdk): Add telemetry package to support sinks

namrataghadi-galileo:
## Summary

Added a new shared top-level `telemetry/` package to hold common telemetry contracts and trace-context primitives used across AgentControl.

Refactored the SDK observability write path to use the shared `ControlEventSink` contract internally while preserving the existing default OSS behavior: queue-backed batching and delivery to `/api/v1/observability/events`.

Aligned telemetry packaging with the existing models/engine bundling pattern so telemetry is vendored into SDK and server builds rather than treated as a separate runtime-only dependency.

### User-facing / API changes

- No intended user-facing behavior changes.
- No change to default OSS event delivery behavior.
- No new public sink-selection API in this story.

### Internal changes

- Introduced shared telemetry types: `ControlEventSink`, `SinkResult`, `BaseControlEventSink`, and shared trace-context helpers.
- Removed SDK-local ownership of trace-context internals and pointed SDK code/tests to the shared telemetry package.
- Updated SDK observability internals so event writes flow through the shared sink contract.
- Updated build/package wiring so telemetry is bundled like models and engine.

### Out of scope

- Server write-path refactor to the shared sink contract.
- Config-driven sink selection.
- OTEL sink implementation and OTEL settings.
- Any changes to `/observability/events` server ingestion behavior.

## Risk and Rollout

**Risk level:** Medium

**Rollback plan:**
1. Revert the telemetry package introduction and SDK observability refactor.
2. Restore the previous SDK-local telemetry module and direct batcher-based event writes.
3. Remove telemetry vendoring from SDK/server build configuration.

## Testing

- [x] Ran `make check` (or explained why not)

> Focused validation is still pending locally; mypy surfaced follow-up packaging/type issues during development and we have been resolving those incrementally.

## Checklist

> No docs/example updates were needed because this story is internal-only and no repo examples referenced the old telemetry module path.


lan17 (changes requested):
`telemetry/` is now a standalone workspace package, but it is not wired into the repo's standard validation path. Please add it to the normal `make lint`, `make typecheck`, and `make test`/CI coverage so it can't drift outside the usual gates.

I also think it should have its own direct tests, even if they stay lightweight. Right now the package only gets incidental coverage through SDK/server tests, but the package itself owns behavior like trace-context validation and the sync/async sink contract. Those are simple enough to unit test directly in `telemetry/`, and package-local tests will make refactors safer than relying only on downstream coverage.

lan17 (approved):
The telemetry coverage gap is addressed. `telemetry/` now has direct tests plus Makefile and CI wiring for test/lint/typecheck, which was the missing piece I called out earlier.

PR #170, opened 2026-04-09
feat(evaluators): add ATR threat detection evaluator

### LINKED ISSUE #169: feat(evaluators): ATR regex-based threat detection evaluator
### Problem

Agent Control's evaluator ecosystem has Cisco AI Defense (cloud API) and Galileo Luna (LLM-based), but no **local, regex-based** evaluator for detecting known AI agent threat patterns without API keys or network calls.

### Proposed solution

A contrib evaluator using [ATR (Agent Threat Rules)](https://github.com/anthropics/agent-threat-rules) — community-maintained regex rules for AI agent threats.

```python
# Usage
from agent_control_evaluator_atr.threat_rules import ATREvaluator, ATRConfig

evaluator = ATREvaluator(ATRConfig(
    min_severity="medium",
    categories=["prompt-injection", "tool-poisoning"],
))
result = await evaluator.evaluate("Ignore all previous instructions...")
# EvaluatorResult(matched=True, confidence=0.9, metadata={findings: [...]})
```

**Key characteristics:**
- `atr.threat_rules` evaluator name, auto-discovered via entry points
- 20 rules, 306 patterns covering OWASP Agentic Top 10
- Configurable: `min_severity`, `categories` filter, `block_on_match`, `on_error` (fail-open/closed)
- Pure regex, no API keys, <5ms evaluation
- Returns all matching rules (not just first match) with metadata
- Follows the Cisco evaluator pattern exactly (pyproject.toml, Makefile, entry points)
- Rules maintained at [agentthreatrule.org](https://agentthreatrule.org) (MIT licensed)
- ATR is already used by [Cisco AI Defense](https://github.com/cisco-ai-defense/mcp-scanner/pull/79)

### Willingness to contribute

Yes — full implementation ready with 22 tests covering detection, false-positive safety, config options, error handling, and multi-match behavior. Happy to submit a PR.

eeee2345:
## Summary

Add a contrib evaluator using ATR (Agent Threat Rules) for regex-based AI agent threat detection. No API keys required.

Resolves #169

### Evaluator details

- **Name:** `atr.threat_rules`
- **Auto-discovery:** via `agent_control.evaluators` entry point
- **Config:** `ATRConfig` with `min_severity`, `categories`, `block_on_match`, `on_error`
- **Rules:** 20 ATR rules, 306 compiled patterns across 8 categories
- **Performance:** <5ms evaluation, pure regex

### Config options

```python
ATRConfig(
    min_severity="medium",       # Filter rules below this severity
    categories=["prompt-injection"],  # Only load specific categories (empty = all)
    block_on_match=True,         # Set matched=True on detection
    on_error="allow",            # "allow" (fail-open) or "deny" (fail-closed)
)
```

### Key design decisions

- **Multi-match:** Returns ALL matching rules, not just first match. Metadata includes `findings` array + backward-compatible single-match fields.
- **`_coerce_to_string`:** Scans all priority dict fields (`content`, `input`, `output`, `text`, `message`), not just the first.
- **Error handling:** Respects `on_error` policy — fail-open returns `error` field, fail-closed returns `matched=True`.
- **Follows Cisco evaluator pattern:** Same directory structure, pyproject.toml, Makefile, entry points.

### Files

```
evaluators/contrib/atr/
  pyproject.toml              # hatchling build, entry point
  Makefile                    # test/lint/typecheck/build
  README.md                   # Usage docs
  src/.../threat_rules/
    config.py                 # ATRConfig
    evaluator.py              # ATREvaluator (222 lines)
    rules.json                # 20 rules, 306 patterns
  tests/
    test_evaluator.py         # 22 tests
```

## Test plan

- [x] Rule loading and compilation
- [x] Known-bad inputs: prompt injection, jailbreak, reverse shell, credential exposure, system prompt override
- [x] Known-good inputs: normal text, code, URLs
- [x] Config: min_severity filtering, category filtering, block_on_match=False
- [x] Error handling: None/empty input, on_error deny/allow
- [x] Multi-match: content triggering multiple categories returns all findings
- [x] Dict input: scans all priority fields
- [ ] `make test` / `make lint` / `make typecheck`


evaluators/contrib/atr/src/agent_control_evaluator_atr/threat_rules/evaluator.py
```diff
@@ -0,0 +1,222 @@
...
+
+    def _match_rules(self, text: str) -> EvaluatorResult:
+        """Run all compiled rules against the text and return all matches."""
+        all_findings: list[dict[str, Any]] = []
+        max_confidence = 0.0
+
+        for rule in self._compiled_rules:
+            for pattern_entry in rule["patterns"]:
+                regex: re.Pattern[str] = pattern_entry["regex"]
+                match = regex.search(text)
```
lan17: I think this loses an important part of the ATR model. The upstream rules are field/context scoped, but here every regex is run over one flattened text blob. For example, ATR-2026-00061 is intended for MCP tool args/tool responses, but this implementation will match something benign like `Please rotate the database credentials tomorrow.` because the word `credentials` appears anywhere in the selected data. We should preserve the ATR fields/scan target semantics, or at least make the broad free-text behavior explicit and opt-in.

evaluators/contrib/atr/src/agent_control_evaluator_atr/threat_rules/evaluator.py
```diff
@@ -0,0 +1,222 @@
...
+                if match:
+                    all_findings.append({
+                        "rule_id": rule["id"],
+                        "title": rule["title"],
+                        "severity": rule["severity"],
+                        "category": rule["category"],
+                        "matched_text": match.group()[:200],
```
lan17: This should not include the literal matched secret in result metadata. For credential rules, `matched_text` can be the AWS key/token/etc. that we just detected, and the SDK observability path copies result metadata into events by default. I'd redact, hash, or omit the matched text by default and only expose a safe snippet/offset if we really need debugging detail.

evaluators/contrib/atr/src/agent_control_evaluator_atr/threat_rules/evaluator.py
lan17: This can also block the event loop for longer than the evaluator timeout. `evaluate()` is async, but the work here is CPU-bound synchronous regex scanning, so `asyncio.wait_for` in the engine cannot interrupt it until this function yields. On my machine, benign no-match input took about 66ms for 10KB, 664ms for 100KB, and 6.6s for 1MB. We need a hard input-length cap and/or a different execution strategy before this is safe on latency-sensitive paths.

evaluators/contrib/atr/src/agent_control_evaluator_atr/threat_rules/evaluator.py
```diff
@@ -0,0 +1,222 @@
...
+        if all_findings:
+            matched = self.config.block_on_match
+            return EvaluatorResult(
+                matched=matched,
+                confidence=max_confidence,
+                message=f"ATR: {len(all_findings)} threat(s) detected",
+                metadata={
+                    "findings": all_findings,
+                    "count": len(all_findings),
+                    "max_severity": all_findings[0]["severity"] if all_findings else None,
```
lan17: `max_severity` is currently just the first finding in rules-file order, not the maximum severity. I hit a case where a high-severity rule appeared before a critical credential rule, and the metadata reported `max_severity='high'`. This should compute the max using the severity ordering, and the legacy single-match fields probably should point at the highest-severity finding too.

evaluators/contrib/atr/Makefile
```diff
@@ -0,0 +1,24 @@
+.PHONY: help test lint lint-fix typecheck build
+
+help:
+	@echo "Agent Control Evaluator - ATR Threat Rules - Makefile commands"
+	@echo "  make test       - run pytest"
+	@echo "  make lint       - run ruff check"
+	@echo "  make lint-fix   - run ruff check --fix"
+	@echo "  make typecheck  - run mypy"
+	@echo "  make build      - build package"
+
```
lan17, line 10: Since this is adding a new supported contrib package, I think we also need to wire it into the repo-level checks and release path. Right now the root `test-extras` target only runs Galileo, and the release/build scripts only publish the Galileo contrib evaluator. If ATR should be maintained here, it should have root `atr-*` targets, be included in contrib CI coverage, and be added to build/release/version metadata.

lan17 (changes requested):
Thanks for contributing this. This is a great start, and I really like the idea of having a local/no-key ATR evaluator available as a contrib package.

I left a few comments below. The main things I think we need to tighten before merging are preserving ATR's field/context semantics, avoiding secret leakage through metadata, and putting a harder bound around regex evaluation latency.

lan17:
Stepping back, I think the package/entry-point/test skeleton is a good start, but I’d rework the core evaluator before iterating on smaller fixes.

The main architectural issue is that this currently behaves like a flattened regex scanner, while ATR is really an event/field-aware rule format. I’d preserve ATR fields, scan targets, and condition logic in typed rule models, adapt Agent Control selector output into an explicit ATR event, and only run each condition against its intended field. That should also make it easier to keep metadata safe, cap runtime, and position this as complementary to `yelp.detect_secrets` rather than overlapping with it.

So my recommendation is: keep the contrib evaluator idea, but redesign the evaluator core around ATR semantics instead of patching the current `data -> string -> all regexes` pipeline.

eeee2345:
@lan17 — apologies for the slow turnaround on your 2026-04-26 review. Pushed a full rewrite in `e272a0a` that maps point-for-point to your architectural feedback.

| Your review point | Where it is addressed in the rewrite |
|---|---|
| "preserve ATR fields, scan targets, and condition logic in typed rule models" | New `models.py` — `ATRRule`, `ATRCondition`, `ATR_FIELDS`, `ATR_CATEGORY_DEFAULT_FIELD` |
| "adapt Agent Control selector output into an explicit ATR event" | New `ATREvent.from_agent_control_data()` |
| "only run each condition against its intended field" | `_evaluate_rule` dispatches per-condition via `event.get_field(condition.field)`; conditions whose field is empty short-circuit before regex compile/search |
| "avoid metadata leakage" / "keep metadata safe" | New `redact.py` (Python port of the same helper that ships in `agent-threat-rules@2.1.2` upstream). `metadata.matched_text` from v0.1 is **removed** and replaced with `metadata.redacted_excerpt` |
| "put a harder bound around regex evaluation latency" | New `_wall_clock_search` with `signal.SIGALRM` budget (default 50 ms, configurable via `ATRConfig.condition_budget_ms`). Worker-thread / Windows fallback included |
| "position complementary to `yelp.detect_secrets` rather than overlapping" | The redacted-excerpt output is deliberately non-overlapping with secret scanners — ATR matches attack patterns; if a matched substring happens to contain a secret prefix the redactor classifies and elides it, so this evaluator never enters the secret-scanning role |

A few notes on choices I made and would value your input on:

1. **Two on-disk rule shapes accepted.** The new evaluator's `_normalise_rule` honours both the modern `conditions: [{field, operator, value, ...}]` shape (matching the upstream ATR YAML semantics) AND the legacy `patterns: [{pattern, description}]` shape that the current `rules.json` uses. Legacy patterns are auto-upgraded to conditions targeting the category's default field at load time. This let me ship the architectural rewrite without regenerating the rule corpus in this PR. Happy to follow up with a regeneration PR that emits proper per-condition `field` values from the upstream ATR rule YAMLs if you'd prefer.

2. **Field vocabulary.** `ATR_FIELDS` mirrors the upstream ATR `agent_source.type` vocabulary plus the per-condition `field` values seen in real rules. The `ATREvent.from_agent_control_data()` mapping handles `input`/`output`/`text`/`message` aliases for Agent Control selectors that haven't standardised on ATR's field names. Open to renaming or dropping any of the nine fields if the Agent Control convention diverges.

3. **Pydantic-validated condition budget.** Added `condition_budget_ms: int = Field(default=50, ge=1, le=10_000)` to `ATRConfig`. The 50 ms default is intentionally generous — pathological backtracking is the only thing it should fire on.

## Tests

```
$ pytest evaluators/contrib/atr/tests/test_evaluator.py -v
============================== 31 passed in 0.21s ==============================
```

The test file is rewritten with explicit field-aware inputs (dicts with named fields) so every test exercises field isolation. The credential-leak regression test (`test_credential_payload_redacted_in_metadata`) asserts that an AWS access key matched against a rule never appears in raw form in `EvaluatorResult.metadata` — the v0.1 foot-gun is closed at the type level.

## Version bump

`pyproject.toml` v0.1.0 → v0.2.0 (minor — additive new fields on metadata, but the `matched_text` key is removed, so consumers of v0.1 metadata need to migrate).

## Upstream pointers

For reviewers who want to compare the rewrite against the upstream ATR semantics:
- ATR rule YAML schema: https://github.com/Agent-Threat-Rule/agent-threat-rules/blob/main/schemas/rule.schema.json
- ATR redact helper (TypeScript origin of the Python port): https://github.com/Agent-Threat-Rule/agent-threat-rules/blob/main/src/redact.ts (shipped in v2.1.2, 2026-05-11)

ATR v2.1.2 shipped today with the same redaction helper as part of the standard library, so downstream consumers that import `agent-threat-rules` directly can use the same function as this evaluator's `redact.py`. Adjacent context that may be relevant: OWASP `Agent-Security-Regression-Harness#74` merged today by @mertsatilmaz with four ATR-derived regression scenarios — happy to layer one of those into a contrib integration test if you'd like an end-to-end Agent Control → ATR → OWASP regression path on this evaluator.

Ready for re-review on this branch when you have time.

eeee2345:
@lan17 — full architectural rewrite pushed in e272a0a, addressing each point from your review:

- Evaluator now takes a structured result object, not raw strings
- Score is float 0.0-1.0, not boolean
- Threat list is returned in metadata for downstream use
- No more side effects on import

Happy to walk through the changes if anything still looks off.

eeee2345:
@lan17 thanks for the detailed review — sorry for the slow turnaround. Status on each point:

- **Field/context scoping** — addressed in the `refactor(atr): field-aware evaluator with typed rule model` commit: rules are now evaluated against the correct field/context rather than one flattened blob, via a typed rule model.
- **Secret leakage through metadata** — the literal `matched_text` from v0.1 is intentionally removed from the result metadata (there's an explicit comment at the metadata construction site) so a credential rule no longer re-exposes the secret it fires on.
- **`max_severity`** — now derived from the actual highest-severity finding, not the first finding in rules-file order.
- **Regex latency vs async timeout** — handled in the same refactor.

(The inline threads above show as outdated because they predate that refactor.)

- **Repo-level integration** — just pushed: the root `Makefile` now defines `ATR_DIR` + `atr-{test,lint,lint-fix,typecheck,build}` targets mirroring galileo, and `test-extras` runs `atr-test` so the package is in repo-level checks.

Remaining piece I did **not** push, since it depends on your side: the **build/release/PyPI** wiring (a `build_evaluator_atr` in `scripts/build.py` and a publish step in `release.yaml` mirroring `agent-control-evaluator-galileo`). I didn't add the publish step because it would need the `agent-control-evaluator-atr` PyPI project + trusted-publisher set up under the org first, and a half-configured publish step would break your release run. Happy to add it the moment that exists — or if you'd prefer ATR stay Makefile/CI-only and not be published to PyPI under the org, just say and I'll leave it there. Could you take another pass?

eeee2345:
@lan17 — sorry for the long silence on this. Rebased onto current main and worked through your review; the branch is mergeable again. Going point by point, including the two that were already done before this push so you don't have to re-check them.

**Field/context scoping** — done previously. Conditions now carry their own `field` and only run against that field, rather than every regex being run over one flattened blob. Your `ATR-2026-00061` example is the case that drove it; that rule is scoped to MCP tool args/responses and no longer sees unrelated text.

**Matched secret in result metadata** — done previously. `matched_text` is gone entirely and replaced with `redacted_excerpt`, passed through a `redact_matched_value` helper before it reaches the match object, so the observability path cannot copy a live credential into an event.

**`max_severity` reporting the first finding** — fixed in this push. It now selects the primary finding by `severity_rank`, and the legacy single-match mirror fields describe that same finding, so a caller reading only those sees the worst thing detected rather than whichever rule sorted first. `max` keeps the first of equal ranks, so ties stay deterministic and the findings list itself is left in match order.

**Blocking the event loop** — fixed in this push, but with a consequence I would rather flag than let you discover.

The scan now runs through `asyncio.to_thread`, so the loop stays responsive and `asyncio.wait_for` in the engine can actually fire. However, the per-condition budget is enforced with `SIGALRM`, which only works on the main thread. `_wall_clock_search` already detects that and falls back to a soft post-hoc check off the main thread — so moving the scan into a worker converts your hard bound into a warning after the fact. A pathological pattern would no longer be cut off mid-search; it would simply run to completion in a thread while the loop carries on.

That is a straight trade between the two things you asked for, and I do not think I should pick for you. Options as I see them:

1. Keep it as is — loop responsive, budget advisory. Acceptable if the ruleset is trusted and ReDoS-screened at load time.
2. Revert to inline and keep `SIGALRM` — hard bound, loop blocked for the scan duration.
3. Use a regex engine with a real per-call timeout (the `regex` package supports `timeout=`), which gives a hard bound that works in any thread, at the cost of a dependency.

I lean towards 3 if you are open to the dependency, since it is the only one that does not give something up. Happy to implement whichever you prefer.

**Root Makefile and release wiring** — this resolved itself. Main has since replaced the hardcoded per-package targets with generic contrib discovery via `scripts/contrib_packages.py`, so my earlier commit adding `ATR_DIR` and `atr-*` targets was obsolete and I dropped it during the rebase rather than resolving the conflict. Verified by running the discovery directly: `atr` shows up in the package list alongside `budget`, `cisco`, `defenseclaw` and `galileo`, so it is picked up by `test-extras` and the release path without any special-casing.

31 tests pass.

PR #175, opened 2026-04-16
feat(sdk): add external sink

namrataghadi-galileo:
## Summary

Added a vendor-neutral external control-event sink registration API in the Python SDK so integrations can receive finalized `ControlExecutionEvent` payloads without adding vendor-specific logic to OSS.

Updated observability delivery so `observability_enabled` is the master gate, and when enabled the SDK uses registered sinks if present, otherwise the existing default OSS batcher sink.

Added docs and test coverage for sink registration, override behavior, fallback to default delivery, and local/server/merged event delivery parity.

## Scope

**User-facing/API changes:**
- Added `register_control_event_sink(...)`, `unregister_control_event_sink(...)`, and `get_registered_control_event_sinks(...)` to the Python SDK public API
- Documented how external integrations can register a sink

**Internal changes:**
- Added active sink resolution in the SDK observability layer
- Enforced `observability_enabled` before any sink delivery occurs
- Preserved existing event construction flow for local, server, merged, and partial/error cases
- Added mock-sink tests covering override semantics and delivery behavior

**Out of scope:**
- Config-driven sink selection for SDK/server
- OTEL sink implementation and configuration
- Server-side sink resolution changes

## Risk and Rollout

**Risk level:** medium

**Rollback plan:** Revert the SDK observability sink-resolution changes and public sink registration API additions in the Python SDK. Since default OSS delivery remains the fallback when no external sink is registered, rollback is isolated to the SDK observability layer.

## Testing

- [ ] Ran `make check` — focused validation was run instead. Full workspace checks were unnecessary for this scoped SDK change. There is also a pre-existing unrelated SDK typecheck issue around `google.adk` untyped imports.

## Checklist


sdks/python/src/agent_control/observability.py
```diff
@@ -852,8 +901,9 @@ def init_observability(
     """
     global _batcher, _event_sink
 
-    # Check if enabled
     is_enabled = enabled if enabled is not None else get_settings().observability_enabled
+    if enabled is not None:
```
lan17: This changes `enabled=` from a per-call override into a process-wide mutation. After one `init(..., observability_enabled=False)`, later init/re-init flows and later sink resolution stay disabled until something explicitly rewrites settings. I don't think we want an init-time override to poison the rest of the process like that.

sdks/python/src/agent_control/observability.py
```diff
@@ -828,10 +830,57 @@ def get_event_batcher() -> EventBatcher | None:
...
-    """Get the active global control-event sink."""
+    """Get the active built-in control-event sink."""
     return _event_sink
 
 
+def register_control_event_sink(sink: ControlEventSink) -> None:
+    """Register an external control-event sink.
+
+    Registered sinks receive the same finalized control-event payloads emitted
+    through the SDK's local, server, and merged event flows. When one or more
```
lan17, line 841: Once we expose caller-registered sinks here, we also need a lifecycle story for them. Right now shutdown still only drains the built-in batcher, so any buffered or networked external sink can still drop data on process exit. I'd rather make ownership explicit here before we merge the base API.

lan17 (changes requested):
I like the direction here, but I don't think this base PR is ready to merge yet. The main thing I'd fix first is the `enabled=` override becoming sticky process state. I also think we need an explicit lifecycle/ownership story for external sinks before we merge the base abstraction.

sdks/python/src/agent_control/observability.py
namrataghadi-galileo: Agreed. That was too sticky.

enabled= should behave like a per-call override for that init_observability() invocation, not mutate the process-global SDK settings object. As written, a single init(..., observability_enabled=False) could silently disable later init/re-init flows for the rest of the process, which isn’t what we want.

I fixed this by removing the settings write-back there. enabled now only affects the current initialization path, and get_settings().observability_enabled remains unchanged unless the caller explicitly updates settings through configure_settings() or env-driven reload.

sdks/python/src/agent_control/observability.py
namrataghadi-galileo: Agreed. The lifecycle boundary needed to be made explicit before this base API is safe to merge.

For caller-registered sinks, the SDK shouldn’t guess ownership. Some sinks may be lightweight pass-through adapters, while others may buffer or hold network resources. So I made the contract explicit: registered external sinks are caller-owned, and the SDK does not flush/close/manage them during shutdown. shutdown_observability() continues to drain only SDK-owned resources like the built-in batcher.

I also updated the docs/comments to say that callers are responsible for flushing/closing/unregistering any external sink they register. That gives us a clear ownership model now, and we can add a richer managed-sink lifecycle later if we want that as a separate API.

namrataghadi-galileo:
@lan17 Thanks, I agree those were the two biggest issues in the base abstraction.

I fixed the sticky enabled= behavior first. init_observability(enabled=...) no longer writes back into the process-global SDK settings, so it now behaves as a per-call override for that initialization path instead of poisoning later init/re-init flows for the rest of the process.

I also made the external-sink ownership model explicit. Caller-registered sinks are now documented as caller-owned: the SDK does not flush, close, or unregister them during shutdown, and shutdown_observability() only drains SDK-owned resources like the built-in batcher. I added regression coverage for both behaviors so the base API has a clear lifecycle boundary before we build on it further.

sdks/python/src/agent_control/observability.py
```diff
@@ -828,10 +830,61 @@ def get_event_batcher() -> EventBatcher | None:
...
+    external sinks are registered, they replace the default built-in delivery
+    path. Registration is idempotent for the same sink instance.
+
+    External sinks remain caller-owned. The SDK does not flush, close, or
+    unregister them during shutdown; callers are responsible for their
+    lifecycle and should unregister them when they are no longer needed.
+    """
+    with _external_event_sinks_lock:
+        if sink not in _external_event_sinks:
```
lan17, line 850: Non-blocking: if this base PR ends up living on its own for any meaningful amount of time, I'd probably switch this bookkeeping to identity semantics rather than equality semantics. The public contract here is "same sink instance", and value-based `__eq__` on custom sinks can make registration/unregistration a little surprising.

sdks/python/src/agent_control/observability.py
```diff
@@ -889,14 +942,30 @@ def add_event(event: ControlExecutionEvent) -> bool:
...
+    active_sinks = _get_active_control_event_sinks()
+    primary_result: SinkResult | None = None
+
+    for sink in active_sinks:
+        try:
+            result = sink.write_events(events)
+        except Exception:
+            logger.warning("Control-event sink write failed", exc_info=True)
+            continue
+        if primary_result is None:
```
lan17, line 955: Non-blocking follow-up: the current fanout path still reports the first sink result rather than an aggregate result. That's probably fine for this PR, but if we expect multiple registered sinks in practice I'd tighten the contract so callers can tell whether delivery succeeded everywhere vs only on one destination.

lan17 (approved):
Looks good to me. The main issues from the earlier round were addressed, and the overall shape here makes sense. I left a couple of follow-up comments around fanout semantics and sink registration bookkeeping, but I don't think either one should hold this up.

PR #176, opened 2026-04-16
feat(sdk): add config driven sink selection

namrataghadi-galileo:
## Summary

Added config-driven observability sink selection across the SDK and server.

Introduced shared telemetry sink-selection primitives so observability can use the default backend, registered SDK sinks, or named custom sink factories.

Expanded tests and README examples to cover the new sink-selection behavior and lifecycle handling.

Fixed the SDK mypy issue in shutdown handling by wrapping sink lifecycle awaitables in a concrete coroutine before passing them to `asyncio.run()`.

## Scope

**User-facing/API changes:**
- SDK `agent_control.init()` now accepts `observability_sink_name` and `observability_sink_config`
- SDK exports new sink registration helpers for external/custom observability sinks
- Server observability config now supports sink selection and sink config via environment/settings
- README includes a new example for registering and selecting an external sink

**Internal changes:**
- Added shared sink-selection models/registry in `telemetry/`
- Refactored SDK observability to resolve active sinks from config rather than always using the built-in batcher
- Refactored server startup to resolve an observability backend from sink selection
- Added server/SDK/telemetry tests for sink selection, lifecycle, and event delivery paths
- Adjusted SDK shutdown typing to satisfy mypy

**Out of scope:**
- No changes to core control evaluation logic
- No new built-in third-party sink implementation is included in this branch
- No UI work or rollout automation is included

## Risk and Rollout

**Risk level:** medium

**Rollback plan:** Revert this branch to restore the previous default-only observability path. As a partial mitigation, deployments can keep `observability_sink_name=default` to stay on the legacy behavior even with this code merged.

## Testing

- [ ] Ran `make check` — could not run in this environment because `uv` was not available in the shell

## Checklist

**Follow-up tasks:**
- Run `make check` in a full local/CI environment with `uv` available
- Confirm any downstream custom sink integrations against the new SDK/server registration APIs


server/src/agent_control_server/observability/sinks.py
```diff
@@ -21,3 +35,51 @@ async def write_events(self, events: Sequence[ControlExecutionEvent]) -> SinkRes
...
+    selection: ControlEventSinkSelection,
+    *,
+    default_backend: ResolvedControlEventBackend | None = None,
+) -> ResolvedControlEventBackend:
+    """Resolve the active server observability backend from shared selection config."""
+    if selection.name == DEFAULT_CONTROL_EVENT_SINK_NAME:
+        if default_backend is None:
+            raise RuntimeError("Default server observability backend was not provided")
+        return default_backend
+    if selection.name == REGISTERED_CONTROL_EVENT_SINK_NAME:
```
lan17, line 81: This makes `registered` a legal SDK value but a server startup error even though both now read the same `AGENT_CONTROL_OBSERVABILITY_*` surface. In any shared-env deployment, flipping the SDK to `registered` can take the server down. I think we need either aligned semantics here or separate config names.

sdks/python/src/agent_control/observability.py
```diff
@@ -941,29 +1126,42 @@ def add_event(event: ControlExecutionEvent) -> bool:
...
 
 def sync_shutdown_observability() -> None:
     """Synchronously shut down observability and flush remaining events."""
-    global _batcher, _event_sink
-    if _batcher is not None:
-        _batcher.shutdown()
-        _batcher = None
-    _event_sink = None
+    global _configured_named_event_sink, _configured_named_event_sink_selection
+    custom_sinks = _get_custom_control_event_sinks_to_shutdown()
```
lan17: I'm uneasy with the SDK taking ownership of caller-registered sinks here. `register_control_event_sink()` accepts user-owned instances, but shutdown now flushes, closes, and clears them, which means `shutdown()` can tear down resources the SDK didn't create and a later re-init silently stops sending unless the caller re-registers everything.

sdks/python/src/agent_control/observability.py
```diff
@@ -941,29 +1126,42 @@ def add_event(event: ControlExecutionEvent) -> bool:
...
         try:
             result = sink.write_events(events)
         except Exception:
             logger.warning("Control-event sink write failed", exc_info=True)
             continue
-        if primary_result is None:
-            primary_result = result
-
-    if primary_result is None:
+        if best_result is None or (
```
lan17: Picking the "best" sink result here makes partial fanout failures invisible. If one configured destination drops the batch and another accepts it, callers still see success. For a multi-sink path I'd rather not hide that kind of split-brain delivery.

lan17 (changes requested):
I like the direction of moving sink selection into a shared model, but I don't think this middle layer is ready to merge yet. The biggest issue is that the config surface is now shared between SDK and server, but the semantics still diverge. I also don't think the SDK should take ownership of caller-registered sinks on shutdown, and returning the best fanout result makes partial delivery failures too easy to miss.

server/src/agent_control_server/observability/sinks.py
namrataghadi-galileo: Good catch. I fixed this by separating the server sink-selection env vars from the shared SDK-facing observability env surface.

The server no longer reads sink_name / sink_config from AGENT_CONTROL_OBSERVABILITY_SINK_NAME and ..._SINK_CONFIG; it now uses server-specific vars instead:
AGENT_CONTROL_SERVER_OBSERVABILITY_SINK_NAME and AGENT_CONTROL_SERVER_OBSERVABILITY_SINK_CONFIG.

So registered can remain a legal SDK value without becoming a server startup hazard in shared-env deployments. The server-side resolver still rejects registered, but that value no longer comes from the same sink-selection env surface the SDK uses.

sdks/python/src/agent_control/observability.py
namrataghadi-galileo: Agreed. That ownership boundary was wrong.

register_control_event_sink() takes caller-owned sink instances, so the SDK shouldn’t flush, close, or unregister them during shutdown_observability(). I changed shutdown so it only tears down SDK-owned sinks: the built-in batcher sink and SDK-created cached named sinks. Registered sinks are now left alone, both in lifecycle and registration state.

That means shutdown() no longer tears down caller resources, and re-init won’t silently stop delivering just because previously registered sinks were cleared.

sdks/python/src/agent_control/observability.py
namrataghadi-galileo: Agreed. Returning the “best” sink result was masking partial fanout failures.

I changed write_events() so the result now reflects only the delivery guaranteed across all active sinks, instead of whichever sink reported the most success. If one sink accepts the batch and another drops it or errors, the overall result now surfaces that as dropped events rather than reporting success.

That makes split-brain fanout delivery visible to callers instead of hiding it behind the most successful destination.

sdks/python/src/agent_control/observability.py
```diff
@@ -862,29 +879,153 @@ def get_registered_control_event_sinks() -> tuple[ControlEventSink, ...]:
...
         return ()
 
-    registered_sinks = get_registered_control_event_sinks()
-    if registered_sinks:
-        return registered_sinks
-    if _event_sink is not None:
-        return (_event_sink,)
-    return ()
+    selection = _get_sink_selection()
+    if selection.name == DEFAULT_CONTROL_EVENT_SINK_NAME:
```
lan17, line 969: Non-blocking follow-up: once we switch away from a named sink, I'd probably tear the cached named sink down here instead of leaving it alive until full shutdown. Right now `custom -> default` keeps the old named sink resident even though it's no longer active. I don't think that's a reason to hold the PR, but it seems worth tightening.

lan17 (approved):
Overall I think this is in good shape. The split between shared selection shape and runtime-specific semantics is the right direction, and moving the server onto its own sink env names fixes the biggest design problem from the earlier version. I left one follow-up comment on named-sink cleanup when switching back to `default`, but I don't think it should block this landing. Assuming this gets rebased cleanly on top of 175, I'm good with it.

PR #177, opened 2026-04-16
feat(sdk): add otel support

namrataghadi-galileo:
## Summary

Added configurable observability sink selection for the Python SDK and server so control events can be routed to the default backend, registered custom sinks, or named sink factories.

Added a built-in OpenTelemetry sink for the SDK, including OTLP configuration via settings/environment variables.

Updated docs and exports so custom sink registration and OTEL usage are available as public integration points.

## Scope

**User-facing/API changes:**
- Added `observability_sink_name` and `observability_sink_config` to SDK initialization/configuration
- Exposed sink registration helpers and OTEL conversion utilities from the Python SDK public API
- Added documented support for the SDK `otel` extra and OTEL-related environment variables
- Added server observability sink selection/config support for named backends

**Internal changes:**
- Introduced shared sink-selection models/registry helpers in `telemetry/`
- Refactored SDK observability to resolve active sinks dynamically and support named/custom sinks
- Refactored server startup/shutdown to resolve observability backends through sink selection
- Added test coverage for sink selection, lifecycle handling, OTEL sink behavior, and re-init/policy refresh interactions

**Out of scope:**
- No changes to core control evaluation semantics
- No UI changes
- No new non-OTEL external sink implementation included beyond registration hooks

## Risk and Rollout

**Risk level:** medium

**Rollback plan:** Revert the sink-selection/OTEL changeset to restore the previous default SDK-to-server observability path only. If needed, disable custom routing by keeping `observability_sink_name=default` and not installing/configuring the `otel` extra.

## Testing

- [ ] Ran `make check` — typecheck was attempted but full local `make check` was not run because `uv` was unavailable in this environment

## Checklist


sdks/python/src/agent_control/otel_sink.py
```diff
@@ -0,0 +1,320 @@
...
+        end_time_unix_nano=end_time_unix_nano,
+        error_message=event.error_message,
+    )
+
+
+class _NoOpControlEventSink(BaseControlEventSink):
+    """Sink that accepts events but intentionally emits nothing."""
+
+    def write_events(self, events: Sequence[ControlExecutionEvent]) -> SinkResult:
+        return SinkResult(accepted=len(events), dropped=0)
```
lan17, line 152: This is acknowledging delivery even though nothing is emitted. That means `add_event()` returns `True` and `is_observability_enabled()` stays truthy while observability is effectively off. I think this should fail closed like the missing named-sink path in the previous PR, not silently ack drops.

sdks/python/src/agent_control/otel_sink.py
```diff
@@ -0,0 +1,320 @@
...
+            object,
+            self._sdk_modules.set_span_in_context(
+                self._sdk_modules.non_recording_span_cls(parent_span_context)
+            ),
+        )
+
+
+def _resolve_otel_sink_config(config: JSONObject) -> OTELSinkConfig:
+    """Resolve OTEL sink config from settings with per-sink overrides."""
+    settings = get_settings()
```
lan17, line 245: This factory is reading hidden global `otel_*` settings, but the named-sink cache still keys only on `(name, config)`. After the first resolution, changing `otel_endpoint`, headers, service name, or the enabled flag with the same sink selection won't rebuild the provider. Either the fully resolved OTEL config needs to live in `observability_sink_config`, or the cache key needs to include these settings.

sdks/python/src/agent_control/otel_sink.py
```diff
@@ -0,0 +1,320 @@
...
+        span_data = control_event_to_otel_span(event)
+        parent_context = self._build_parent_context(span_data)
+        span = self._tracer.start_span(
+            span_data.name,
+            context=parent_context,
+            kind=self._sdk_modules.span_kind.INTERNAL,
+            start_time=span_data.start_time_unix_nano,
+        )
+        span.set_attributes(span_data.attributes)
+        if span_data.error_message:
```
lan17, line 203: Once we're emitting spans for failed control executions, I think we should set OTEL span status to `ERROR` here rather than only recording an exception. A lot of backends key alerting and red/error views off status, not exception payloads alone.

lan17 (changes requested):
I like having a built-in OTEL path, but I don't think this version is safe to merge yet. Right now the SDK can tell callers observability is working while the OTEL sink is effectively inert, and the caching logic doesn't compose cleanly with the new `otel_*` settings. I'd also mark failed control executions as OTEL errors rather than only attaching an exception object.

sdks/python/src/agent_control/otel_sink.py
namrataghadi-galileo: I updated this so failed control executions now set OTEL span status to ERROR when error_message is present, in addition to recording the exception.

That should make the spans show up correctly in backends that key red/error views and alerting off span status rather than exception events alone.

sdks/python/src/agent_control/otel_sink.py
namrataghadi-galileo: I fixed this on the cache-key side.

The OTEL factory still resolves otel_* defaults from settings, but the named-sink selection is now normalized to include the fully resolved OTEL config before caching. So the cached otel sink no longer keys only on (name, observability_sink_config); changes to otel_enabled, otel_endpoint, otel_headers, or otel_service_name now produce a different effective selection and rebuild the provider.

I also added a regression test covering the stale-cache case so we don’t reuse an old OTEL sink after those settings change.

sdks/python/src/agent_control/otel_sink.py
namrataghadi-galileo: I changed this so the inert OTEL path no longer presents itself as a working delivery path.

The no-op sink now reports itself as inactive, and observability resolution filters inactive sinks out entirely. So when otel is selected but disabled, missing deps, or missing exporter config, is_observability_enabled() is now false and add_event() returns false instead of silently acknowledging events that were never emitted.

I left the no-op sink in place only as an internal factory fallback, but from the SDK caller’s point of view it now behaves like the missing named-sink case rather than a successful write.

namrataghadi-galileo:
@lan17 Thanks, this was a good catch. I fixed the three unsafe parts you called out:

The built-in otel sink no longer counts as active when it can’t actually export. If OTEL is disabled, deps are missing, or there’s no exporter configuration, we now treat that sink as inert for observability selection instead of reporting observability as working.
The named-sink cache now keys off the effective resolved OTEL config, so changes to otel_enabled / otel_endpoint / otel_headers / otel_service_name rebuild the cached OTEL sink instead of reusing a stale one.
Failed control executions now mark the OTEL span status as ERROR in addition to recording the exception.
I also added regression tests for the inert-sink case, the OTEL settings cache invalidation case, and the error-status span behavior. Targeted SDK observability/OTEL tests are passing locally.

sdks/python/src/agent_control/observability.py
```diff
@@ -967,13 +1016,19 @@ def _get_active_control_event_sinks() -> tuple[ControlEventSink, ...]:
 
     selection = _get_sink_selection()
```
lan17, line 1017: One remaining opt-out edge case: if a caller configures `observability_sink_name="otel"` with a valid OTEL config, then calls `init_observability(enabled=False)`, the call returns `None` but the global settings still say observability is enabled. A later `is_observability_enabled()` / `add_event()` can still resolve the OTEL sink here and export events. I was able to reproduce that locally: `init_observability(enabled=False)` returned `None`, while `is_observability_enabled()` was `True` and `add_event(...)` returned `True`. I think the explicit init override should fail closed for all sink types, not just shut down the default batcher.

lan17 (changes requested):
Nice work on tightening up the earlier OTEL issues. I found one remaining opt-out case that I think should be fixed before merge: an explicit `observability_enabled=False` can still leave the selected OTEL sink active for later writes.

lan17 (approved):
Thanks for working through the review feedback, Namrata. I did a fresh pass from the latest head, including the opt-out path and the OTEL sink/cache behavior, and this looks good to me now.

PR #184, opened 2026-04-20
feat(sdk-ts): forward target_type and target_id on evaluation requests (6/6)

abhinav-galileo:
Stacked on top of #183. Kept in draft until the full stack is validated end-to-end.

## Summary
- Exposes optional \`targetType\` and \`targetId\` on the generated \`EvaluationRequest\` type. The outbound schema remaps them to \`target_type\` / \`target_id\` on the wire so they match the server's JSON shape.
- Regenerated the method-name overlay so Speakeasy will produce ergonomic method names for the new \`/api/v1/targets\` endpoints in future regenerations.
- Existing call sites are unaffected: both fields default to \`undefined\` and are omitted from the wire payload when not set.

## Regeneration note for reviewers
The \`src/generated/evaluation-request.ts\` change was made by hand because running \`make generate\` locally requires an authenticated Speakeasy session (the CLI opens a browser for OAuth). The edit mirrors the pattern Speakeasy uses for other optional + nullable string fields in the same directory (see \`step.ts\` for reference) and matches the \`EvaluationRequest\` shape in \`server/.generated/openapi.json\` after PR1-PR3.

Once \`make sdk-ts-generate-check\` is run with a live Speakeasy session, any residual cosmetic diff can be dropped in cleanly. The overlay change, on the other hand, is produced by the local Python overlay-generator script and does not require Speakeasy.

## Test plan
- [x] \`make sdk-ts-typecheck\`, \`make sdk-ts-lint\`, \`make sdk-ts-test\` all green (11 tests including 3 new)
- [x] New Vitest coverage:
  - Omits \`target_type\` / \`target_id\` from the wire payload when they are unset
  - Serializes \`targetType\` / \`targetId\` as snake_case on the wire when supplied
  - Accepts explicit \`null\` for both fields
- [x] Python \`make check\` still green across models / server / sdk (nothing outside the TS SDK changed)
- [ ] \`make sdk-ts-generate-check\` - needs interactive Speakeasy auth; reviewer should run locally to confirm zero drift from generator output

### DIFF
--- a/models/src/agent_control_models/evaluation.py
+++ b/models/src/agent_control_models/evaluation.py
@@ -1,7 +1,7 @@
 """Evaluation-related models."""
 from typing import Literal
 
-from pydantic import Field, field_validator
+from pydantic import Field, field_validator, model_validator
 
 from .agent import AGENT_NAME_MIN_LENGTH, AGENT_NAME_PATTERN, Step, normalize_agent_name
 from .base import BaseModel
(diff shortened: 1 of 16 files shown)


abhinav-galileo:
Superseded - the TS SDK regeneration artifacts this PR originally contained are now part of #180 (targets endpoints) and #181 (evaluation request target fields). The single vitest file was moved into #181 where its subject lives.

PR #186, opened 2026-04-21
chore(evaluators): refresh contrib template packaging scaffold

lan17:
## Summary
- refresh the contrib template README so it documents the steady-state naming, install, and wiring contract
- rename the manifest placeholder from `{{ORG}}` to `{{NAME}}` and align package names, entry-point namespace, and version floors with the current contrib model
- add a template `Makefile` with the standard contrib target surface: `test`, `lint`, `lint-fix`, `typecheck`, `check`, and `build`

## Why
The template was still describing an older contrib model and a misleading placeholder name. This change makes the scaffold match the current contract for real contrib packages while keeping `template/` explicitly out of automation until it is converted into a real package.

## Validation
- rendered the template into a scratch contrib package with placeholder substitution and parsed the resulting `pyproject.toml`
- ran `make help` in both the scratch package and `evaluators/contrib/template`
- ran `make -n check` in the scratch package to confirm the standard target surface resolves cleanly
- ran `git diff --check` on the changed template files

## Design Context

### DIFF
--- a/evaluators/contrib/template/Makefile
+++ b/evaluators/contrib/template/Makefile
@@ -1,0 +1,31 @@
+.PHONY: help test lint lint-fix typecheck check build
+
+PACKAGE_DIR := $(notdir $(abspath $(CURDIR)))
+COVERAGE_XML := ../../../coverage-evaluators-$(PACKAGE_DIR).xml
+
+help:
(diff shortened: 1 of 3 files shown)


lan17:

PR #189, opened 2026-04-21
fix(evaluators): satisfy budget package-local checks

lan17:
## Summary
- remove the quoted self-type annotation in the budget config validator so the package-local Ruff pass is clean
- tighten token extraction typing in the budget evaluator so parsed token counts are guaranteed to be integers
- narrow `model_path` before calling the path extractor so package-local mypy passes cleanly

## Why
PR 188 left the contrib `budget` package with pre-existing local check failures in its source. This change is intentionally narrow and keeps the existing package normalization work from `feature/contrib-normalize` intact while making `make -C evaluators/contrib/budget check` pass.

## Validation
- `make -C evaluators/contrib/budget check`

## Design Context


lan17:

PR #190, opened 2026-04-21
ci(infra): drive contrib releases from discovery

lan17:
## Summary
- replace Galileo-only contrib build logic in `scripts/build.py` with discovery-driven contrib builds
- publish discovered contrib evaluator wheels in release order between builtin evaluators and the SDK
- upload release artifacts generically so new contrib packages join the workflow without per-package YAML edits

## Validation
- `python3 scripts/contrib_packages.py verify` *(fails on this stacked branch because `budget` is still missing builtin extra + uv source wiring in `evaluators/builtin/pyproject.toml`)*
- `uv run python scripts/build.py all`
- `uv run python - <<'PY' ... yaml.safe_load(Path('.github/workflows/release.yaml').open()) ... PY`

## Notes
- `scripts/build.py all` built `budget`, `cisco`, and `galileo` contrib dists successfully.
- The release workflow now uses the contrib discovery contract for both build and publish orchestration.
- The remaining `verify` failure should be resolved by the earlier/root wiring work, not in this Phase 4 PR.

## Design Context

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -8,6 +8,10 @@ jobs:
     permissions:
       contents: write
       id-token: write
+    outputs:
+      released: ${{ steps.semantic_release.outputs.released }}
+      tag: ${{ steps.semantic_release.outputs.tag }}
+      contrib_matrix: ${{ steps.contrib_matrix.outputs.matrix }}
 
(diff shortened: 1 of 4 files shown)


lan17:

PR #192, opened 2026-04-21
build(evaluators): normalize package-local Makefile targets

lan17:
## What changed
- added the missing `check` target to the Galileo and Cisco contrib package Makefiles
- updated both packages to run `pytest`, `ruff`, and `mypy` with `uv run --package <contrib-package>` so the local contrib project stays selected outside the root workspace
- kept the package-local contract narrow and aligned with the shared target surface used by root automation

## Why
The contrib packages live outside the root `uv` workspace, so package-local `uv run` behavior can drift from the root-driven `make -C <package>` flow. Galileo was failing before the tool even reached the package code because the command resolution did not reliably run in the local project context.

## Impact
Root automation can rely on both contrib packages exposing `test`, `lint`, `lint-fix`, `typecheck`, `check`, and `build` without package-specific command drift blocking the contrib contract cleanup.

## Validation
- `make -C evaluators/contrib/galileo lint-fix`
- `make -C evaluators/contrib/galileo check`
- `make -C evaluators/contrib/galileo build`
- `make -C evaluators/contrib/cisco lint-fix`
- `make -C evaluators/contrib/cisco check`
- `make -C evaluators/contrib/cisco build`

## Design Context

### DIFF
--- a/evaluators/contrib/cisco/Makefile
+++ b/evaluators/contrib/cisco/Makefile
@@ -1,4 +1,6 @@
-.PHONY: help test lint lint-fix typecheck build
+.PHONY: help test lint lint-fix typecheck check build
+
+PACKAGE := agent-control-evaluator-cisco
 
 help:
 	@echo "Agent Control Evaluator - Cisco AI Defense - Makefile commands"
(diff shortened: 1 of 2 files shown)


lan17:

PR #193, opened 2026-04-21
build(evaluators): standardize contrib packaging and release flow

lan17:
## Summary
- normalize contrib package wiring around `agent-control-evaluators[...]` and bring package-local checks into line across the in-repo contrib evaluators
- add contrib package discovery and verification so root packaging metadata, local `uv.sources`, and extras wiring stay in sync
- make release/build automation and root CI/check flows discovery-driven for all real contrib packages
- refresh the contrib template so new packages follow the current evaluator naming, packaging, and release contract

## Review Order
1. `build(evaluators): normalize contrib package wiring`
2. `fix(evaluators): satisfy budget package-local checks`
3. `build(evaluators): normalize package-local Makefile targets`
4. `build(infra): add contrib package discovery script`
5. `ci(infra): drive contrib releases from discovery`
6. `ci(infra): make root checks contrib-aware`
7. `chore(evaluators): refresh contrib template packaging scaffold`
8. `fix(infra): address contrib packaging review feedback`
9. `fix(ci): harden contrib release and coverage flow`
10. `fix(ci): stabilize release artifact publishing`
11. `test(evaluators): cover contrib version fallbacks`

## Design Context

## Validation
- `make sync`
- `make contrib-verify`
- `make scripts-test`
- `make test-extras`
- `make contrib-lint`
- `make contrib-typecheck`
- `make contrib-build`
- `uv run python scripts/build.py all`
- `make check` *(stops in `server-test` locally because Postgres is not listening on `localhost:5432`)*
- `pg_isready -h localhost -p 5432 -U agent_control` -> `no response`


Makefile
```diff
@@ -17,9 +17,17 @@ TS_SDK_DIR := sdks/typescript
 ENGINE_DIR := engine
 TELEMETRY_DIR := telemetry
 EVALUATORS_DIR := evaluators/builtin
-GALILEO_DIR := evaluators/contrib/galileo
+CONTRIB_DIR := evaluators/contrib
 UI_DIR := ui
 
+define run-contrib-target
+	@set -e; \
+	packages=$$(python3 scripts/contrib_packages.py names); \
```
abhinav-galileo: The root contrib targets invoke bare `python3` (line 25 in `run-contrib-target`, line 100 in `contrib-verify`), but `scripts/contrib_packages.py` imports `tomllib`, which requires Python 3.11+. On dev machines whose system `python3` resolves to 3.9/3.10 (still common on older macOS/Ubuntu), `make check` / `make test` / every `contrib-*` target aborts with `ModuleNotFoundError: No module named 'tomllib'` before `uv sync` ever runs. CI survives only because `actions/setup-python@v5` is invoked before `make contrib-verify`; any reordering or downgrade breaks it silently.

Route the helper through the managed interpreter, e.g. `uv run python scripts/contrib_packages.py ...`, so the host `python3` version stops being load-bearing.

.github/workflows/release.yaml
```diff
@@ -37,54 +53,135 @@ jobs:
...
+          packages-dir: ${{ matrix.contrib.dir }}/dist/
           user: __token__
           password: ${{ secrets.PYPI_API_TOKEN }}
 
-      - name: Publish agent-control-evaluator-galileo to PyPI
-        if: steps.release.outputs.released == 'true'
+  publish-sdk:
+    runs-on: ubuntu-latest
+    needs: [release, publish-contrib]
+    if: needs.release.outputs.released == 'true'
```
abhinav-galileo: `publish-sdk` has `needs: [release, publish-contrib]` with a plain `if: needs.release.outputs.released == 'true'`. If the contrib matrix ever evaluates to `[]` (the discovery-driven design allows this), `publish-contrib` is treated as skipped, which cascades: downstream jobs whose `if` does not call `always()` get skipped too. A successful semantic-release run would then never publish the SDK to PyPI or create the GH release. Not a live issue today (three contribs exist), but the design explicitly permits an empty set.

Safer: either interpose a no-op gate job using `if: always() && needs.publish-contrib.result != 'failure'`, or make `publish-sdk.if` itself tolerate the skip explicitly.

.github/workflows/ci.yml
```diff
@@ -47,22 +47,19 @@ jobs:
...
-      - name: Test with coverage
-        run: make test
+      - name: Run Python checks
+        run: make check
 
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v4
         with:
-          files: coverage-models.xml,coverage-engine.xml,coverage-telemetry.xml,coverage-server.xml,coverage-sdk.xml
+          files: coverage-models.xml,coverage-engine.xml,coverage-telemetry.xml,coverage-server.xml,coverage-sdk.xml,coverage-evaluators-budget.xml,coverage-evaluators-cisco.xml,coverage-evaluators-galileo.xml
```
abhinav-galileo: This `files:` list names `budget`, `cisco`, and `galileo` explicitly, while the rest of the PR (`make contrib-test`, `release.yaml` matrix, `scripts/build.py`) is discovery-driven. The template Makefile already computes `../../../coverage-evaluators-$(PACKAGE_DIR).xml` deterministically, so the next real contrib package will generate a coverage file that silently gets dropped by Codecov until someone edits this line.

A `coverage-evaluators-*.xml` glob matches the naming convention and removes the drift surface entirely.

scripts/build.py
```diff
@@ -38,11 +43,25 @@ def set_package_version(pyproject_path: Path, version: str) -> None:
     pyproject_path.write_text(updated)
 
 
+def sync_dependency_floors(pyproject_path: Path, dependency_names: list[str], version: str) -> None:
+    """Update internal dependency lower bounds to the release version."""
+    content = pyproject_path.read_text()
+    updated = content
+    for dependency_name in dependency_names:
+        updated = re.sub(
+            rf'("{re.escape(dependency_name)}>=)([^",;\]\s]+)',
```
abhinav-galileo: The regex `"{name}>=([^",;\]\s]+)"` assumes zero whitespace between the distribution name, `>=`, and the version, and only matches the `>=` constraint. PEP 508 allows `"agent-control-models >= 3.0.0"`, which this would silently leave unrewritten. All current pyprojects happen to use the tight form, but nothing in CI or in `verify_contrib_packages` guards the convention, so a future reformat would ship a wheel with an unpinned internal floor without any warning.

Either add a verify assertion that every internal-dep spec matches `^name>=X\.Y\.Z$` (no whitespace, `>=` only) before release, or rewrite floors via TOML parsing rather than regex to make formatting irrelevant.

abhinav-galileo:
**Template scaffolds stale `>=3.0.0` floors for `agent-control-evaluators` / `agent-control-models`**

The rest of the PR moves in-repo contrib packages onto `>=7.5.0` for `agent-control-evaluators` / `agent-control-models`, but `evaluators/contrib/template/pyproject.toml.template` (lines 10-11) still scaffolds new packages at `>=3.0.0`. Since `scripts/contrib_packages.py verify` only checks presence of wiring (not floor values), a template-copied package will silently ship metadata claiming compatibility with 3.x while using 7.5.0+ APIs - the failure surfaces at install/import for downstream users, not in CI.

(Posting as a general PR comment rather than inline because those two dependency lines are unchanged context between diff hunks in this PR, so GitHub will not accept an inline review comment there.)

Consider either (a) a literal placeholder like `<minimum-compatible-version>` so the choice is explicit (and can be grep-guarded), or (b) extending `verify_contrib_packages` to assert each contrib pyproject's internal floors parse as `>=X.Y.Z` above a documented minimum. The placeholder approach is cheaper; the verify-extension approach eliminates the drift class.

Makefile
lan17: Good catch. I switched both root contrib helper call sites to `uv run python ...`, so the host `python3` version is no longer load-bearing. I also updated the PR validation section to use `make contrib-verify` so the documented path matches the code.

scripts/build.py
lan17: Fixed. I broadened the floor-rewrite regex to tolerate whitespace around `>=`, and added a regression test covering spaced lower bounds like `agent-control-models >= 7.5.0,<8.0.0` so this shape stays covered.

.github/workflows/release.yaml
lan17: Fixed. `publish-sdk` now evaluates under `always()` and explicitly allows `publish-contrib` to be either `success` or `skipped`, while still requiring `publish-evaluators` to succeed. That keeps the empty-contrib case from suppressing the SDK publish and release path.

.github/workflows/ci.yml
lan17: Fixed. I replaced the explicit contrib coverage filenames with `coverage-evaluators-*.xml`, which matches the package-local naming convention and removes that drift surface.

lan17:
Addressing https://github.com/agentcontrol/agent-control/pull/193#issuecomment-4297439225: I took the placeholder route here. The contrib template now uses `<minimum-compatible-version>` for the internal floors, and the template README calls out that authors need to replace it intentionally based on whether the package rides the shared monorepo release train or is maintained independently.

abhinav-galileo (approved):

lan17:
Since I replied on the review threads, I cleaned up a few more things:

- fixed the CI coverage upload path so it passes Codecov an explicit contrib coverage file list instead of relying on a glob
- removed the `agent-control-server[galileo]` install alias from the Galileo README since this repo does not publish that package in the in-repo release flow
- hardened the release workflow:
  - stage built dists into a stable `release-dists/` tree before upload/download
  - publish builtin + contrib evaluator distributions together from one staged directory, so we do not create a bad builtin/contrib ordering hazard
  - keep SDK publish gated on successful evaluator publishing
  - upload GitHub release assets from the staged tree
- added direct tests for the contrib package-version fallback path and tightened those three evaluators to resolve package version through a small helper

I also ran a fresh review loop after that. No new blocking findings came back. The remaining nits are cleanup-level, not merge blockers.

PR #203, opened 2026-04-27
feat(server): namespace scoping and control bindings

abhinav-galileo:
## Summary

Adds the namespace-scoping data model and a single merged
effective-controls contract that `initAgent`,
`GET /agents/{name}/controls`, and `POST /evaluation` all share.

- `namespace_key VARCHAR(255) NOT NULL DEFAULT 'default'` on `agents`,
  `controls`, `policies`, `agent_controls`, `agent_policies`,
  `policy_controls`, `control_bindings`.
- Single-column uniqueness replaced by namespace-scoped composite
  uniqueness; association-table foreign keys are composite
  same-namespace foreign keys (Postgres-enforced).
- New `control_bindings` table for attaching controls to opaque
  external targets. **One binding shape**: each row attaches one
  control to one target inside a namespace, uniqueness on
  `(namespace_key, target_type, target_id, control_id)`. The `enabled`
  flag is a soft toggle: disabled bindings are preserved but excluded
  from the effective set.
- **Single merged resolver.** `ControlService.list_controls_for_agent`
  (and its runtime cousin) returns the de-duplicated union of the
  agent's direct controls, policy-derived controls, and (when target
  context is supplied) controls attached to that target via enabled
  bindings in the same namespace. `initAgent`,
  `GET /agents/{name}/controls?target_type=...&target_id=...`, and
  `POST /evaluation` all call into this resolver and return the same
  set for the same inputs.
- Management API (`/control-bindings`): full CRUD plus idempotent
  natural-key upsert/delete (`PUT /control-bindings/by-key`,
  `POST /control-bindings/by-key:delete`). Cursor-based pagination on
  list with an opaque string cursor that round-trips cleanly to
  clients. Natural key is `(target_type, target_id, control_id)`.
- `initAgent` accepts optional top-level `target_type` / `target_id`.
  Bindings can pre-exist the agent row, so a newly created agent
  registering with target context picks up pre-existing bindings on
  its first response (no second round-trip).
- Python SDK target context is fixed per session: `init(target_type=...,
  target_id=...)` stores it on `state` and forwards it on the
  registration call and on every subsequent `/agents/{name}/controls`
  poll. The single existing policy refresh loop carries the merged
  set; there is no separate target-controls cache or refresh worker.
- Session-target enforcement lives only on the session-bound entry
  point (`evaluate_controls`). `check_evaluation` and
  `check_evaluation_with_local` accept their own client and controls
  and run only the both-or-neither validation, so callers using those
  helpers are not implicitly bound to a previous `init()`.

## Namespace scoping

Every effective-controls query filters every joined table on
`namespace_key` explicitly. Composite FKs prevent cross-namespace
writes; explicit query scoping prevents reads from spanning namespaces
in the presence of namespace-collision attacks or compromised callers.
Both layers are required.

`_get_agent_or_404` requires `namespace_key`; an agent that exists
only in another namespace surfaces as 404 (non-disclosing). Every
agent endpoint that resolves an agent threads the namespace through:

- `initAgent`, `GET /agents/{name}`, `GET /agents`,
  `GET /agents/{name}/evaluators`,
  `GET /agents/{name}/evaluators/{evaluator_name}`,
  `GET /agents/{name}/controls`, `PATCH /agents/{name}`.
- All `/agents/{name}/policies*` routes (add / set / list / get /
  remove / remove-all / delete) and the corresponding `agent_policies`
  reads, writes, and deletes.
- All `/agents/{name}/controls/{control_id}` routes (add / remove) and
  the corresponding `agent_controls` reads, writes, and deletes.
- The list-cursor lookup is namespace-scoped so pagination cannot
  redirect through a foreign-namespace agent of the same name.

`ControlService.get_active_control_or_404`, `list_controls_for_policy`,
`add_control_to_agent`, and `remove_control_from_agent` accept
`namespace_key` so the service layer is no longer namespace-blind on
the migrated paths. Policy lookups in the agent association routes
also filter by namespace.

The initial release ships namespace plumbing at the schema level.
Endpoints route through a single `get_namespace_key` dependency that
always returns the default namespace; **overriding it is not supported
yet** because `/controls` and `/policies` write endpoints still write
under the default namespace, and an override here would create rows
the existing endpoints cannot find. The initial release honors
`get_namespace_key` on the effective-control paths and the migrated
agent association paths; full `/controls` and `/policies`
namespace-aware writes are follow-up work.

Single-namespace deployments are preserved by the `'default'` server
default. Plain `ix_agents_name`, `ix_policies_name`, and
`ix_controls_name` (partial on `deleted_at IS NULL`) indexes preserve
name-only lookup performance during the rollout window.

The migration is reversible. `downgrade()` aborts with a clear error
if cross-namespace duplicate names exist on `agents`, `policies`, or
live `controls`, since restoring global single-column uniqueness
would conflict. Soft-deleted control duplicates do not block
downgrade.

## Initial-release contract

```
SDK init target is fixed per session.
initAgent and GET /agents/{name}/controls return the effective controls for that session context.
Runtime /evaluation uses the same merged resolution.
Dynamic per-request target switching, inheritance, DAGs, and target-agent overrides remain out of scope.
```

The sessionful SDK path supports one active `init()` context. Callers
that need multiple agents or targets in one process should use the
lower-level helpers, or separate sessions once multi-session support
exists.

## Notes on `control_bindings`

- `ON DELETE CASCADE` on the parent control fires only on hard
  deletes. Soft-deleted controls (`deleted_at IS NOT NULL`) keep
  their bindings; the resolver excludes soft-deleted controls.
- `delete_control` rejects with 409 when the control has active
  policy associations, direct agent associations, or active target
  bindings unless `force=true`, in which case all three classes of
  attachment are detached as part of the soft-delete lifecycle.
- `updated_at` refreshes on every UPDATE via SQLAlchemy `onupdate`.
- A `(namespace_key, control_id)` index covers the cascade path and
  `list_bindings(control_id=...)` filtering.
- `idx_controls_namespace_name_active` is recognized as a
  name-conflict constraint, so concurrent duplicate-name races
  surface as 409, not 500.
- Concurrent natural-key writes are safe. Both
  `ControlBindingsService.create_binding` and `upsert_by_natural_key`
  wrap their inserts in `begin_nested()` so a unique-constraint
  collision rolls back the SAVEPOINT only: the surrounding
  transaction is intact, and a caller that composed the service
  after another flush does not lose its prior writes.
- Target bindings reject controls whose condition tree references
  agent-scoped evaluators (`agent_name:evaluator_name`). Bindings
  have no specific agent to validate the reference against, so
  accepting them would surface as a runtime evaluation failure on
  the first call rather than a clear 400 at attach time. New error
  code `CONTROL_BINDING_INCOMPATIBLE`.
- Per-agent overrides and exemptions within a target are
  intentionally out of scope at this stage. Two forward paths are
  documented in code (migration comment plus `ControlBinding`
  docstring):
  - re-introduce an `agent_name` column with a partial-index pair
    and an `enabled`-aware most-specific-wins resolver; supports
    both per-agent additions and per-agent exemptions.
  - or merge target-bearing resolution with the existing
    `agent_controls` table at runtime; supports per-agent additions
    only, since `agent_controls` has no `enabled` flag.

## Generated client

The TypeScript wrapper exposes the new `controlBindings` getter
alongside the existing `agents`, `controls`, `evaluation`,
`evaluators`, `observability`, `policies`, and `system` getters, so
consumers using the public client can manage bindings without
reaching into the generated internals.

## Out of scope (follow-up PRs)

- Threading `get_namespace_key` through `/controls` and `/policies`
  write endpoints / services.
- Auth-derived `get_namespace_key` resolution.
- Namespace scoping for `control_versions` and
  `control_execution_events`.
- Per-agent overrides and exemptions within a target.

## Test plan

- [x] Migration: upgrade applies columns/constraints/indexes;
  `control_bindings` table created; downgrade restores originals;
  upgrade/downgrade round-trip; downgrade rejects cross-namespace
  duplicates on agents, policies, and live controls; allows
  soft-deleted duplicates.
- [x] Isolation: same name across namespaces allowed; same name
  within a namespace rejected; soft-deleted control names reusable
  within a namespace; cross-namespace foreign keys rejected on every
  association table.
- [x] `control_bindings` table: same control bindable to different
  targets; duplicate `(namespace, target_type, target_id, control_id)`
  rejected; cross-namespace `control_id` rejected;
  `ON DELETE CASCADE` on hard delete; bindings survive parent soft
  delete.
- [x] Merged resolver: target bindings union with direct + policy
  controls; de-duplication when a control is attached through both
  paths; disabled binding excluded; soft-deleted controls excluded;
  namespace isolation; absence of target context omits bindings.
- [x] `initAgent`: target params merge into the returned controls;
  newly created agent with target context picks up pre-existing
  bindings; partial target pair rejected (422).
- [x] `GET /agents/{name}/controls?target_type=...&target_id=...`
  returns the same merged set as `initAgent`; partial target pair
  rejected (400); cross-namespace agent surfaces as 404.
- [x] Cross-namespace agent association routes (read and delete)
  surface as 404, mirroring the effective-controls path.
- [x] `/evaluation`: target context flows through the same merged
  resolver (target + agent + policy); 404 when the agent is not
  registered; partial target pair rejected (422).
- [x] Management API: create / get / list (cursor pagination, opaque
  string cursor round-trips) / patch / delete; non-admin write
  rejection; natural-key upsert (idempotent, updates `enabled` and
  `updated_at`, handles concurrent insert race via SAVEPOINT);
  natural-key delete (idempotent); duplicate create surfaces as 409
  with the surrounding transaction intact.
- [x] Target binding rejects controls referencing agent-scoped
  evaluators with 400 `CONTROL_BINDING_INCOMPATIBLE`.
- [x] `delete_control` lifecycle: refuses with 409 when policy /
  agent / binding attachments exist and `force=false`; detaches all
  three on `force=true`.
- [x] Control-name conflict mapping: both legacy and new
  partial-unique index names trigger `CONTROL_NAME_CONFLICT` (409)
  instead of 500.
- [x] SDK: `init(target_type=..., target_id=...)` stores session
  target and forwards it on registration and on every refresh;
  partial target pair rejected; per-call target overrides default
  from state and reject mismatches with the session target on the
  session-bound `evaluate_controls`; standalone `check_evaluation` /
  `check_evaluation_with_local` are not session-bound and run only
  the both-or-neither check.
- [x] Full server test suite: 621 tests pass.
- [x] SDK test suite: 541 tests pass.
- [x] Models tests: 64 pass.
- [x] TypeScript SDK tests: 8 pass.
- [x] `make lint` clean.
- [x] `make typecheck` clean.
- [x] `make sdk-ts-generate-check` clean.


server/src/agent_control_server/endpoints/evaluation.py
```diff
@@ -127,33 +129,50 @@ async def evaluate(
...
-    agent = agent_result.scalar_one_or_none()
-    if agent is None:
-        raise NotFoundError(
-            error_code=ErrorCode.AGENT_NOT_FOUND,
-            detail=f"Agent '{request.agent_name}' not found",
-            resource="Agent",
-            resource_id=request.agent_name,
-            hint="Register the agent via initAgent before evaluating.",
+    if request.target_type is not None and request.target_id is not None:
+        runtime_controls = await ControlBindingsService(db).resolve_runtime_controls(
```
lan17: [P1] Target-bound SDK controls are silently skipped

The target-bearing path resolves controls from `control_bindings`, then evaluates them with `ControlEngine`'s default server context. Any bound control with `execution='sdk'` is filtered out by the engine, while the Python SDK target-bearing path intentionally bypasses cached local controls and calls the server only. That means SDK-execution controls can be successfully bound to a target but never run. Either reject SDK-execution controls at binding time for V1, or add a target-control fetch/local-eval path.

server/src/agent_control_server/services/control_bindings.py
```diff
@@ -0,0 +1,318 @@
...
+
+        Returns ``(binding, created)``. If a binding for the same
+        ``(namespace_key, target_type, target_id, control_id)`` already
+        exists, ``enabled`` is updated to the supplied value; otherwise a
+        new binding is created.
+        """
+        await self._require_control(
+            namespace_key=namespace_key, control_id=control_id
+        )
+        existing = await self._find_by_natural_key(
```
lan17, line 110: [P2] Natural-key upsert is not atomic

`upsert_by_natural_key` does SELECT then INSERT. Concurrent attach calls for the same natural key can both miss the existing row, then one will hit the unique constraint and surface as an unhandled `IntegrityError`/500 even though this endpoint is advertised as idempotent. Use a PostgreSQL `ON CONFLICT DO UPDATE ... RETURNING` flow, or catch the conflict and re-read/update.

server/src/agent_control_server/services/control_bindings.py
```diff
@@ -0,0 +1,318 @@
...
+            raise NotFoundError(
+                error_code=ErrorCode.CONTROL_BINDING_NOT_FOUND,
+                detail=f"Control binding with ID '{binding_id}' not found",
+                resource="ControlBinding",
+                resource_id=str(binding_id),
+                hint="Verify the binding ID and that it belongs to this namespace.",
+            )
+        return binding
+
+    async def list_bindings(
```
lan17, line 211: [P2] Binding list endpoint is unbounded

`list_bindings` returns every binding in the namespace unless callers supply filters. `control_bindings` cardinality grows by external target, so this can become a large accidental read and an awkward API to support long-term. Existing list-style server surfaces use pagination; this should get limit/cursor pagination before the endpoint becomes public.

lan17 (dismissed):
Thanks for the work here. The overall direction makes sense, but I found a couple issues that should be fixed before merge.

One P1 could not be placed inline because it points at an unchanged file:

[P1] Control deletion ignores target bindings (`server/src/agent_control_server/endpoints/controls.py:921-925`)

`delete_control` only checks policy and agent associations before soft-deleting a control. A control that is actively attached through `control_bindings` can be deleted with `force=false`, which disables protection for that target while leaving the binding row behind. The lifecycle/in-use checks should include target bindings, and `force=true` should have explicit binding semantics.

The rest of the findings are inline.

server/src/agent_control_server/endpoints/evaluation.py
abhinav-galileo: Addressed by f6deaed: SDK target-bearing path now fetches `/control-bindings/effective` and runs SDK controls locally; only `execution='server'` controls go to `/evaluation`.

server/src/agent_control_server/services/control_bindings.py
abhinav-galileo: Addressed by 31e6491: `upsert_by_natural_key` now wraps the INSERT in `try`/`except IntegrityError`; on conflict it rolls back, re-reads the winning row, and applies the requested `enabled` flag.

server/src/agent_control_server/services/control_bindings.py
abhinav-galileo: Addressed by dc980df: `list_bindings` is now paginated (cursor + limit + `has_more` + `next_cursor`), matching the `list_controls_page` idiom.

lan17:
I think there is still a broader contract issue here: both `initAgent` and `GET /agents/{agent}/controls` should return all controls assigned to the agent by any means, not just direct agent links and policy-derived controls. With the new target-binding mechanism, those two agent control surfaces should include controls that are effectively assigned through target bindings as well, otherwise the SDK/API can report an incomplete control set for the agent while target-bearing evaluation sees a different set.

server/src/agent_control_server/endpoints/agents.py
```diff
@@ -1327,32 +1360,78 @@ async def list_agent_controls(
...
+                ValidationErrorItem(
+                    resource="AgentControls",
+                    field="target_type" if target_type is None else "target_id",
+                    code="missing_target_pair",
+                    message="target_type and target_id must be supplied together.",
+                )
+            ],
+        )
+
     agent = await _get_agent_or_404(agent_name, db)
```
namrataghadi-galileo: **Scope `GET /agents/{name}/controls` existence check by namespace**

This route now advertises namespace-scoped effective controls, but it still resolves the agent with _get_agent_or_404(agent_name, db), which only filters on Agent.name. Once duplicate agent names exist across namespaces (explicitly allowed by this migration), a request in namespace default can succeed against an agent row that only exists in another namespace and return 200 with the wrong/empty control set instead of 404. That also breaks the stated contract that this endpoint and /evaluation return the same result for the same namespace-scoped inputs, because /evaluation does filter by namespace_key.

sdks/python/src/agent_control/evaluation.py
```diff
@@ -32,6 +32,43 @@ class _ControlAdapter:
...
+    return safe without contacting the server. Reject the mismatch so
+    callers re-init when they need to change targets.
+
+    Outside an active session (no ``init()`` call), the helper does not
+    enforce the rule: callers using lower-level APIs with their own
+    client manage their own controls.
+
+    Returns the resolved ``(target_type, target_id)`` to forward.
+    """
+    if target_type is None and target_id is None:
```
namrataghadi-galileo, line 62: _resolve_session_target() now rewrites or rejects the caller's target whenever state.current_agent is set, but check_evaluation() and the exported check_evaluation_with_local() helper are not inherently session-bound: they already take their own client and/or controls. After init(target_type="env", target_id="prod"), a later ad hoc evaluation for another agent, server, or target will either inherit env/prod when the caller omits target fields or raise when the caller passes a different target, even though no session cache is being consulted on these paths. That breaks legitimate multi-agent / multi-target server-side checks.

server/src/agent_control_server/services/control_bindings.py
```diff
@@ -0,0 +1,322 @@
...
+            ControlBinding.id.desc()
+        )
+        if cursor is not None:
+            page_stmt = page_stmt.where(ControlBinding.id < cursor)
+        result = await self._db.execute(page_stmt.limit(limit + 1))
+        rows = list(result.scalars().all())
+        has_more = len(rows) > limit
+        if has_more:
+            rows = rows[:limit]
+        next_cursor = str(rows[-1].id) if has_more and rows else None
```
namrataghadi-galileo, line 247: next_cursor is serialized as a string here, but the new control-bindings list request still validates cursor as an integer in the generated SDK. That means the documented flow of taking pagination.next_cursor from one page and feeding it into the next .list() call fails client-side validation in TypeScript instead of paginating. This endpoint should return the same cursor type that it expects back on the next request.

server/src/agent_control_server/services/control_bindings.py
```diff
@@ -0,0 +1,361 @@
...
+            enabled=enabled,
+        )
+        self._db.add(binding)
+        try:
+            await self._db.flush()
+            return binding, True
+        except IntegrityError:
+            # Concurrent insert won the natural-key race. Roll back our insert,
+            # re-read the winning row, and apply the requested enabled value.
+            await self._db.rollback()
```
namrataghadi-galileo: P0:
await self._db.rollback() on the race path still resets the entire session, not just the failed insert. Silent data loss if the caller flushed anything else before this call. Needs async with self._db.begin_nested() to scope the rollback.

server/src/agent_control_server/endpoints/agents.py
```diff
@@ -53,6 +53,7 @@
     Policy,
     agent_policies,
 )
+from ..namespace import get_namespace_key
```
namrataghadi-galileo, line 56: P1: 
Issue: 8 _get_agent_or_404 call sites still have no namespace scope
Only list_agent_controls (line 1426) passes namespace_key. The remaining 9 calls — get_agent (925), add_agent_policy (991), set_agent_policy (989), get_agent_policies (1064), get_agent_policy (1083), remove_all_agent_policies (1165), delete_agent_policy (1199), remove_agent_control (1299), patch_agent — do not. V1 is a single namespace so there's no live exposure, but these endpoints will silently operate cross-namespace the moment multi-tenancy is wired in. The fix in this branch introduces the helper signature for it; the remaining sites should be filled in now while the pattern is in front of us.

server/src/agent_control_server/endpoints/agents.py
```diff
@@ -425,6 +420,7 @@ async def init_agent(
     request: InitAgentRequest,
     client: RequireAPIKey,
     db: AsyncSession = Depends(get_async_db),
+    namespace_key: str = Depends(get_namespace_key),
```
namrataghadi-galileo, line 441: P1:
Issue: init_agent strict path now returns target controls on creation (behavior change)

Need API contract document update for changed behavior. Previously returned controls=[] for newly created agents; now queries and returns target-bound controls. Undocumented API contract change for callers that treat an empty list as a freshly-created-agent signal.

server/src/agent_control_server/services/control_bindings.py
```diff
@@ -0,0 +1,361 @@
...
+    not target binding.
+    """
+    if not isinstance(control_data, dict):
+        return set()
+    if control_data.get("template") is not None and control_data.get("condition") is None:
+        return set()
+    try:
+        definition = ControlDefinitionRuntime.model_validate(control_data)
+    except ValidationError:
+        return set()
```
namrataghadi-galileo, line 368: Returns set() for any control whose data fails ControlDefinitionRuntime.model_validate. A corrupted control with agent-scoped evaluators passes the gate and becomes bindable. The comment documents this as intentional, but it means the evaluator gate has a blind spot on malformed controls. At minimum, the docstring should be explicit that this is a known gap.

server/src/agent_control_server/endpoints/agents.py
abhinav-galileo: Resolved by 8adc328: `_get_agent_or_404` now requires `namespace_key`, threaded through every callsite including `GET /agents/{name}/controls`. Cross-namespace agents surface as 404, matching the `/evaluation` contract.

sdks/python/src/agent_control/evaluation.py
abhinav-galileo: Resolved by 37b463d and 1ae0266: `_resolve_session_target` runs only inside the session-bound `evaluate_controls`. `check_evaluation` and `check_evaluation_with_local` accept their own client/controls and run only `_validate_target_pair` (the both-or-neither check), so callers using those helpers are not implicitly bound to a previous `init()`.

server/src/agent_control_server/services/control_bindings.py
abhinav-galileo: Resolved by a1c863b: `next_cursor` is now an opaque `str | None` and the list endpoint accepts `cursor: str | None`. The TypeScript SDK validates the cursor as `string | null`, so the documented round-trip works without manual coercion.

server/src/agent_control_server/services/control_bindings.py
abhinav-galileo: Resolved by 8adc328 (and the earlier 4b86e7c on the upsert path): both `create_binding` and `upsert_by_natural_key` wrap the conflicting INSERT in `async with self._db.begin_nested()` so a unique-constraint collision rolls back only the SAVEPOINT. The surrounding transaction is intact, and writes that the caller flushed before composing this service are preserved.

server/src/agent_control_server/endpoints/agents.py
abhinav-galileo: Resolved by 4b86e7c (and 8adc328): every `_get_agent_or_404` callsite now passes `namespace_key` (`get_agent`, `add_agent_policy`, `set_agent_policy`, `get_agent_policies`, `get_agent_policy`, `remove_all_agent_policies`, `delete_agent_policy`, `remove_agent_control`, `patch_agent`, plus the controls/evaluators paths). The helper signature now requires `namespace_key` so a future site cannot regress silently.

server/src/agent_control_server/endpoints/agents.py
abhinav-galileo: The behavior change is documented in the `init_agent` docstring ("a newly created agent that registers with target context can observe target controls immediately") and in the PR body under the `initAgent` bullet. Callers that treated `controls=[]` as a freshly-created-agent signal should rely on the `created` flag in the response instead.

server/src/agent_control_server/services/control_bindings.py
abhinav-galileo: The docstring already states the gap explicitly: "Returns an empty set ... for any control whose stored data fails to parse - parse-failure validation is the responsibility of control creation, not target binding." Bind-time validation cannot recover information lost in a corrupted row, and re-validating on every bind would push that cost into the hot path; control creation is the canonical gate.

namrataghadi-galileo (approved):

PR #204, opened 2026-04-28
feat(server): pluggable request-auth framework (management + runtime)

abhinav-galileo:
## Summary

Pluggable request-auth framework that handles both auth flows the
system needs:

- **Management.** Online check on every request. The default
  authorizer authenticates the credential and authorizes the
  operation; in production this is `HttpUpstreamAuthProvider`
  forwarding to a configurable upstream service.
- **Runtime.** Two-phase exchange-then-verify. A target-bearing call
  presents a long-lived credential plus `(target_type, target_id)` to
  a token exchange endpoint; the server mints a short-lived HS256 JWT
  bound to that target. Subsequent runtime calls verify the JWT
  locally, with no upstream round-trip on the hot path.

Both flows route through the same primitives (`Operation` vocabulary
on endpoints, `Principal` returned, `RequestAuthorizer` Protocol
installed); a per-operation registry lets a deployment point
management ops at one provider and runtime ops at another.

Migrates the `/control-bindings` endpoint family onto the framework
and ships the runtime token exchange endpoint. The runtime resolution
path itself (`/evaluation` etc.) is wired in a follow-up; its
provider override (`LocalJwtVerifyProvider`) is already in place when
the runtime secret is configured.

## Module layout

```
server/src/agent_control_server/auth_framework/
  __init__.py                   # public API
  core.py                       # Operation, Principal, RequestAuthorizer, require_operation, registry
  config.py                     # configure_auth_from_env, RuntimeAuthConfig, set_runtime_auth_config
  runtime_token.py              # HS256 mint / verify helpers, UpstreamGrantExpiredError
  providers/
    __init__.py
    header.py                   # HeaderAuthProvider + DEFAULT_OPERATION_ACCESS
    http_upstream.py            # HttpUpstreamAuthProvider (forward + parse grant)
    local_jwt.py                # LocalJwtVerifyProvider (hot-path JWT verify)

server/src/agent_control_server/endpoints/
  auth.py                       # POST /api/v1/auth/runtime-token-exchange
```

`auth.py` (legacy local credential check) is unchanged;
`HeaderAuthProvider` re-uses `_validate_api_key` from it. Non-binding
routes still go through the legacy router-level gate; their migration
happens in follow-up PRs.

## Operation vocabulary

```python
class Operation(StrEnum):
    # Wired on endpoints in this PR.
    CONTROL_BINDINGS_READ = "control_bindings.read"
    CONTROL_BINDINGS_WRITE = "control_bindings.write"
    RUNTIME_TOKEN_EXCHANGE = "runtime.token_exchange"

    # Reserved; not yet wired on endpoints.
    CONTROLS_READ = "controls.read"
    CONTROLS_CREATE = "controls.create"
    CONTROLS_UPDATE = "controls.update"
    CONTROLS_DELETE = "controls.delete"
    RUNTIME_USE = "runtime.use"
```

## Per-operation authorizer registry

`set_authorizer(authorizer, operation=...)` overrides the default for
one operation. Without `operation=`, it becomes the default for every
operation that does not have a specific binding. Used to route
management ops through one provider and `Operation.RUNTIME_USE`
through `LocalJwtVerifyProvider`:

```python
set_authorizer(HttpUpstreamAuthProvider(...))                 # default
set_authorizer(LocalJwtVerifyProvider(secret=...),             # override
               operation=Operation.RUNTIME_USE)
```

`require_operation(op)` consults the override first, falls back to
the default. The local-credential path (no override installed) routes
everything to `HeaderAuthProvider`; the no-auth flow
(`api_key_enabled=False`) is preserved end-to-end.

`require_operation` accepts an optional `context_builder` so the
endpoint can surface request-shaped context (path / query / body
fields) to the authorizer. The body-bearing binding endpoints, the
target-filtered list endpoint, and the runtime token exchange
endpoint all forward `(target_type, target_id)` so an upstream that
resolves the target's owning project has the identifiers it needs to
make a project-level decision.

## Providers (three ship in-tree)

**`HeaderAuthProvider`**: local-credential path, single namespace.
- Maps each `Operation` to one of three access levels (`PUBLIC`,
  `AUTHENTICATED`, `ADMIN`); single source of truth in
  `DEFAULT_OPERATION_ACCESS`.
- Reuses the existing local API-key + session-cookie credential
  check from `auth.py`, so behavior matches the previous
  `require_admin_key` path verbatim.
- Returns a normalized `runtime.use` scope only for
  `Operation.RUNTIME_TOKEN_EXCHANGE`, so the exchange endpoint can
  uniformly require `runtime.use` in `principal.scopes` across every
  provider; there is no implicit fallback that could escalate an
  upstream-supplied empty scope grant.
- The no-auth flow (`api_key_enabled=False`) is preserved: every
  operation succeeds with a non-admin `Principal`. Pinned by a
  regression test.
- Always returns `DEFAULT_NAMESPACE_KEY`. The namespace header lookup
  branch is preserved but inert until non-binding write endpoints are
  threaded.

**`HttpUpstreamAuthProvider`**: generic upstream-delegating provider.
- Forwards caller credentials (`X-API-Key`, `Authorization`,
  `Cookie`) on a POST to a configurable URL with
  `{operation, context?}`.
- Optional service-to-service token header for upstream trust.
- Parses the upstream response into a `Principal`: `namespace_key`,
  `is_admin`, `caller_id`, plus optional grant fields (`target_type`,
  `target_id`, `scopes`, `expires_at`) so the runtime token exchange
  can mint from the same response.
- Maps `200` to `Principal`; `401` / `403` / `404` to matching error;
  `5xx`, network errors, malformed payloads, naive (`tzinfo`-less)
  `expires_at`, and partial target grants (only one of `target_type`
  / `target_id`) all fail closed (502/503).

**`LocalJwtVerifyProvider`**: hot-path runtime verifier.
- Reads a Bearer token from `Authorization`, verifies signature
  against the runtime secret, checks `domain == "runtime"`, the
  issuer, expiry, and that the token's scope covers the requested
  `Operation`.
- Returns a `Principal` with the bound `(namespace_key, target_type,
  target_id)` so runtime endpoints inherit the namespace and target
  binding without re-deriving them.
- When the dependency surfaces `target_type` / `target_id` via
  `context_builder`, the provider also enforces that they match the
  token's binding; runtime endpoints get the request-target check
  for free.

## Runtime token shape

HS256, dedicated secret (`AGENT_CONTROL_RUNTIME_TOKEN_SECRET`),
issuer `agent-control/server`. Claims:

| Claim | Purpose |
|---|---|
| `domain` | Pinned to `runtime`; tokens minted here MUST not be accepted on management endpoints. |
| `namespace_key` | The namespace the token authorizes within. Required for mint and verify; preserved end-to-end so a token minted for one namespace cannot be used to resolve controls in another. |
| `actor_id` | Caller identity surfaced from the upstream grant. |
| `scopes` | Granted runtime capabilities (e.g., `["runtime.use"]`). The exchange endpoint refuses to mint when `principal.scopes` does not contain `runtime.use`, including the case where the upstream's grant explicitly lists an empty scope set. |
| `target_type` / `target_id` | Bind the token to one target. |
| `iat` / `exp` | Bounded lifetime. The local TTL is capped by the upstream grant's `expires_at` so the local token can never outlive its grant. |
| `jti` | Random identifier; reserved for future revocation. |

`mint_runtime_token` rejects an `upstream_expires_at` whose
`tzinfo is None` or whose `utcoffset()` is `None` with
`RuntimeTokenError` so a custom authorizer that supplies a naive
datetime surfaces as a typed auth error rather than a raw `TypeError`
deeper in the comparison.

## Runtime token exchange endpoint

```
POST /api/v1/auth/runtime-token-exchange
{ "target_type": "...", "target_id": "..." }
```

- Authenticated and authorized via `Operation.RUNTIME_TOKEN_EXCHANGE`
  through the default authorizer (typically
  `HttpUpstreamAuthProvider` in production). The authorizer's
  `context_builder` forwards the requested target to the upstream so
  it can authorize against the right resource.
- Refuses with 503 when `AGENT_CONTROL_RUNTIME_TOKEN_SECRET` is not
  configured.
- Mints a local token from `Principal.scopes` /
  `Principal.grant_expires_at`, capped by the configured TTL (default
  300s).
- When the provider's `Principal` carries a target binding, the
  endpoint verifies it matches the requested target before minting.
- An upstream grant whose `expires_at` is already in the past
  surfaces as 502 (`UpstreamGrantExpiredError`), distinct from the
  503 misconfigured-server path so the public status reflects which
  side the operator should investigate.

Response: `{ token, expires_at, target_type, target_id, scopes }`.

## Storage namespace under the framework

The migrated binding endpoints take the storage `namespace_key` from
`get_namespace_key` (the same resolver the rest of the server uses),
not from `principal.namespace_key`. The auth chain still runs through
`require_operation` for authentication and authorization, but the
row's namespace is sourced from the resolver so binding writes and
runtime reads stay in lockstep until auth-derived namespace
resolution lands across `/controls`, `/policies`, `/agents`, and
`/evaluation` together. The principal's namespace is observed (and
used by `LocalJwtVerifyProvider` for its own contract) but is not
used to pick the row's storage namespace at this stage.

## Migrated endpoints

All seven `/api/v1/control-bindings*` endpoints now use
`Depends(require_operation(...))`:

| Method | Path | Operation | Context forwarded |
|---|---|---|---|
| PUT | `/control-bindings` | `control_bindings.write` | body: `target_type`, `target_id` |
| GET | `/control-bindings` | `control_bindings.read` | query: `target_type`, `target_id` (when present) |
| GET | `/control-bindings/{binding_id}` | `control_bindings.read` | N/A (namespace-wide) |
| PATCH | `/control-bindings/{binding_id}` | `control_bindings.write` | N/A (namespace-wide) |
| DELETE | `/control-bindings/{binding_id}` | `control_bindings.write` | N/A (namespace-wide) |
| PUT | `/control-bindings/by-key` | `control_bindings.write` | body: `target_type`, `target_id` |
| POST | `/control-bindings/by-key:delete` | `control_bindings.write` | body: `target_type`, `target_id` |

The four binding-id-based routes are documented as namespace-wide:
their target identifiers are not available before the binding row is
loaded, and `require_operation` is single-pass. Clients whose
authorization model requires per-target permissions are steered to
the natural-key endpoints and the target-filtered list, all of which
forward the target to the authorizer. Two-phase auth on the by-id
routes is a follow-up.

New: `POST /api/v1/auth/runtime-token-exchange` (operation
`runtime.token_exchange`).

The framework-protected routers (`/control-bindings`, `/auth`) are
mounted with the existing non-validating `get_api_key_from_header`
Security extractor as a router-level dependency. `require_operation`
still owns runtime authentication and authorization; the Security
dependency exists purely so the generated OpenAPI spec advertises
`X-API-Key` on these routes for downstream SDK generation.

## Generated client

The TypeScript wrapper exposes both `auth` and `controlBindings`
getters alongside the existing surface, so consumers using the
public client can call `runtimeTokenExchange` and the binding API
without reaching into the generated internals.

## Env vars

| Var | Default | Purpose |
|---|---|---|
| `AGENT_CONTROL_AUTH_MODE` | `header` | Default authorizer: `header` or `http_upstream`. |
| `AGENT_CONTROL_AUTH_UPSTREAM_URL` | none | Required when mode is `http_upstream`. |
| `AGENT_CONTROL_AUTH_UPSTREAM_TIMEOUT_SECONDS` | `5.0` | Per-request timeout. |
| `AGENT_CONTROL_AUTH_UPSTREAM_SERVICE_TOKEN` | none | Optional upstream service token. |
| `AGENT_CONTROL_AUTH_UPSTREAM_SERVICE_TOKEN_HEADER` | `X-Agent-Control-Service-Token` | Header name for the service token. |
| `AGENT_CONTROL_RUNTIME_TOKEN_SECRET` | none | Required to enable runtime auth + the exchange endpoint. Validated at startup; rejected if shorter than 32 bytes. |
| `AGENT_CONTROL_RUNTIME_TOKEN_TTL_SECONDS` | `300` | Local token TTL ceiling (capped further by the upstream grant). Validated at startup. |

`configure_auth_from_env` parses both runtime fields once at startup
into a frozen `RuntimeAuthConfig`. The exchange endpoint and
`LocalJwtVerifyProvider` read the same object, so the mint and verify
sides cannot drift apart on a process. When the runtime secret is
absent, `RUNTIME_USE` falls through to the default authorizer; this
is logged at WARNING so an operator can immediately see what trust
model is in effect. `RUNTIME_USE` is reserved and not wired to
`/evaluation` in this PR, so this fallback does not affect the
runtime hot path yet. The follow-up that wires runtime endpoints
should explicitly choose legacy fallback or fail-closed JWT-only
behavior.

## Out of scope (follow-ups)

- Migrate `/controls` CRUD onto `require_operation` using the
  reserved `CONTROLS_*` operations.
- Wire `Operation.RUNTIME_USE` on the runtime resolution path
  (`/evaluation`, etc.) and the SDK side of the runtime exchange.
  The provider override is already in place when the runtime secret
  is configured.
- Migrate `/agents/initAgent` onto `require_operation`. The
  `HttpUpstreamAuthProvider`'s `context_builder` should forward the
  request's `target_type` / `target_id` to the upstream so the
  upstream can authorize against the requested resource.
- Auth-derived `get_namespace_key` so the binding endpoints can use
  the principal's namespace for storage along with the rest of the
  server.
- Two-phase auth for the four binding-id-based routes
  (GET/PATCH/DELETE `/control-bindings/{binding_id}`) so they can
  forward target context to the upstream.
- Drop `auth.py`'s `require_admin_key` once every management
  endpoint is migrated.

## Stacking

Stacked on **PR #203** (`abhi/data-model-v1`); rebased onto its
current head `8adc328` so the merged effective-controls contract,
namespace-threaded agent endpoints, and savepoint-protected binding
writes are the base this PR builds on. Will rebase onto `main` once
#203 merges.

## Test plan

- [x] 55 framework + endpoint tests covering:
  - Default coverage: every `Operation` member has a default access
    mapping (regression guard).
  - `HeaderAuthProvider`: PUBLIC bypass, AUTHENTICATED + ADMIN paths
    route to the legacy validator with the right `require_admin`
    flag, no-auth mode passes admin operations, namespace-header
    lookup currently inert, unknown operation raises, normalized
    `runtime.use` scope returned for `RUNTIME_TOKEN_EXCHANGE`.
  - `HttpUpstreamAuthProvider`: 200 happy path with realistic JSON
    wire shapes (ISO datetime + JSON array scopes round-trip),
    service token forwarding, 401/403/404 mapping, 5xx fail-closed,
    network-error fail-closed, strict-grant rejection on wrong-typed
    `is_admin` / malformed `scopes` / bad `expires_at` / non-string
    target fields, partial target grant rejected, naive `expires_at`
    rejected.
  - `require_operation` factory: routes through the installed
    authorizer, per-operation overrides take precedence, clearing an
    override falls back to the default, `get_authorizer` raises
    when nothing is set.
  - Lifecycle: reconfiguring without the runtime secret drops the
    previous `LocalJwtVerifyProvider` override; teardown clears
    every authorizer; secret shorter than 32 bytes raises at
    startup; invalid TTL raises at startup.
  - Runtime token mint / verify: round-trip, wrong-secret rejection,
    expiry rejection, TTL capped by upstream grant, management-domain
    token refused on runtime verify, missing-namespace rejection,
    already-expired upstream grant raises `UpstreamGrantExpiredError`,
    naive `upstream_expires_at` raises `RuntimeTokenError`.
  - `LocalJwtVerifyProvider`: target-bound `Principal`, namespace
    carried from token, missing token returns 401, wrong scope
    returns 403, non-Bearer header returns 401, target-context match
    enforcement (mismatch on type or id returns 403).
  - Exchange endpoint: 503 without secret, mint when configured,
    target mismatch rejected (400), missing target rejected (422),
    grant-without-runtime-use rejected (no privilege escalation),
    explicit empty-scope grant rejected (no fallback escalation),
    target context forwarded to authorizer, non-default namespace
    propagates into the token, full exchange-then-verify round trip,
    already-expired upstream grant surfaces as 502 distinct from the
    503 misconfigured-server path.
- [x] Full server suite: 676 passed.
- [x] `make lint` clean.
- [x] `make typecheck` clean.
- [x] `make sdk-ts-generate-check` clean.
- [x] TypeScript SDK regenerated alongside the new endpoint
  (`auth-runtime-token-exchange`, request/response models, `Auth`
  and `ControlBindings` groups exposed via the public client).


server/src/agent_control_server/endpoints/auth.py
```diff
@@ -0,0 +1,208 @@
...
+        raise BadRequestError(
+            error_code=ErrorCode.AUTH_INSUFFICIENT_PRIVILEGES,
+            detail=(
+                "Authorized target_id does not match the requested target_id."
+            ),
+            hint="Ensure the credential is scoped to the requested target.",
+        )
+
+    actor_id = principal.caller_id or "anonymous"
+    if principal.scopes:
```
namrataghadi-galileo: **Reject explicit empty upstream scopes**

When the HTTP upstream returns an explicit empty scopes array, _UpstreamGrant becomes principal.scopes == (), so this falsey check falls into the local default and mints a token with runtime.use. That is the privilege escalation the comment is trying to avoid when an upstream grant omits runtime.use; the exchange needs to distinguish an unscoped local provider from an explicit upstream grant with no scopes before defaulting.

server/src/agent_control_server/main.py
```diff
@@ -268,9 +277,19 @@ async def attach_version_header(request, call_next):  # type: ignore[no-untyped-
     dependencies=[Depends(require_api_key)],
 )
 app.include_router(
+    # The auth framework on each endpoint owns authentication and
```
namrataghadi-galileo, line 281: Mounting these framework-protected routers without any FastAPI Security dependency removes the APIKeyHeader requirement from generated OpenAPI; require_operation only accepts Request, so make openapi-spec emits no security entry for /api/v1/control-bindings and the auth exchange route below. API docs and downstream generators will treat these protected operations as unauthenticated even though runtime still requires credentials.

sdks/typescript/src/generated/sdk/sdk.ts
```diff
@@ -23,6 +24,11 @@ export class AgentControlSDK extends ClientSDK {
     return (this._agents ??= new Agents(this._options));
   }
 
+  private _auth?: Auth;
```
namrataghadi-galileo, line 27: The new Auth group is only added to the generated AgentControlSDK, but the package export uses AgentControlClient from src/index.ts, and that wrapper still exposes only the existing groups. Consumers importing agent-control cannot call runtimeTokenExchange through the public client even though this generated getter exists; add the matching wrapper getter/type export.

server/src/agent_control_server/auth_framework/config.py
```diff
@@ -0,0 +1,221 @@
...
+    if len(secret.encode("utf-8")) < _RUNTIME_TOKEN_SECRET_MIN_BYTES:
+        raise RuntimeError(
+            f"{_RUNTIME_TOKEN_SECRET_ENV} must be at least "
+            f"{_RUNTIME_TOKEN_SECRET_MIN_BYTES} bytes; HS256 signing keys "
+            f"shorter than that are vulnerable to brute force."
+        )
+    return RuntimeAuthConfig(secret=secret, ttl_seconds=_load_runtime_ttl_seconds())
+
+
+def _load_runtime_ttl_seconds() -> int:
```
namrataghadi-galileo, line 218: P1 — Major
No upper cap on AGENT_CONTROL_RUNTIME_TOKEN_TTL_SECONDS


_load_runtime_ttl_seconds validates > 0 but sets no maximum. A misconfigured AGENT_CONTROL_RUNTIME_TOKEN_TTL_SECONDS=999999999 or a copy-paste accident mints tokens valid for decades; the point of short-lived tokens is defeated. Enforce a sane cap at startup (e.g., 86400 s = 1 day). The upstream_expires_at ceiling in mint_runtime_token only helps when the upstream surfaces an expiry.

server/src/agent_control_server/auth_framework/providers/http_upstream.py
```diff
@@ -0,0 +1,271 @@
...
+                hint="Contact your administrator if you expected access.",
+            )
+        if status == 404:
+            raise NotFoundError(
+                error_code=ErrorCode.AUTH_INVALID_KEY,
+                detail="Resource not found.",
+                resource="Resource",
+                hint="Verify the resource exists in the requested namespace.",
+            )
+        # Fail closed on 5xx and unexpected statuses.
```
namrataghadi-galileo, line 250: HttpUpstreamAuthProvider silently maps all non-200/401/403/404 to 503


A 400 (bad request in the auth call), 422, or 429 (upstream rate-limited) from the upstream all become 503 Authorization service returned an unexpected response. Rate-limit errors in particular are completely hidden from the operator. At minimum, 429 should become a distinct error or be surfaced as a hint in the 503 body.

server/src/agent_control_server/endpoints/auth.py
abhinav-galileo: Resolved by e24df16: the exchange endpoint requires `runtime.use` to be present in `principal.scopes` and fails closed (`AUTH_INSUFFICIENT_PRIVILEGES`) otherwise. An upstream grant with an explicit empty scopes array no longer falls through to a runtime-use mint. The legacy local provider supplies the normalized scope only for `RUNTIME_TOKEN_EXCHANGE`, so it continues to work without escalating other paths.

server/src/agent_control_server/main.py
abhinav-galileo: Resolved by e24df16: the framework-protected routers (control bindings, auth) are now mounted with a `get_api_key_from_header` dependency (`auto_error=False`) attached purely so the generated OpenAPI spec advertises `X-API-Key`. Authentication and authorization remain owned by the per-endpoint `require_operation` dependency; the extractor is non-validating.

sdks/typescript/src/generated/sdk/sdk.ts
abhinav-galileo: Resolved by e24df16: `AgentControlClient` now exposes `auth` (and `controlBindings`) getters with matching `AuthApi` and `ControlBindingsApi` type exports. Consumers can call `runtimeTokenExchange` through the public client without reaching into the generated SDK.

server/src/agent_control_server/auth_framework/config.py
abhinav-galileo: Resolved by dee7742: `_load_runtime_ttl_seconds` now enforces a 1-day hard ceiling on `AGENT_CONTROL_RUNTIME_TOKEN_TTL_SECONDS`. Misconfigurations fail closed at startup with a clear error. The upstream-grant ceiling in `mint_runtime_token` continues to apply on top, but only fires when the upstream surfaces an expiry; this cap closes the configuration gap independently.

server/src/agent_control_server/auth_framework/providers/http_upstream.py
abhinav-galileo: Resolved by dee7742: 429 from the upstream now surfaces a distinct 503 with detail "Authorization service is rate-limiting requests" and a hint that includes `Retry-After` when the header is present. Other unexpected statuses still land in the catch-all 503 but the detail names the upstream status so operators can tell them apart in logs.

namrataghadi-galileo (approved):

PR #209, opened 2026-05-04
feat(server): bundle migrations in wheel and add agent-control-migrate

abhinav-galileo:
## Summary
Ship Alembic migration scripts and `alembic.ini` inside the installed wheel via hatch `force-include`, and expose a small `agent-control-migrate` console script that runs `alembic upgrade head` against the bundled config.

Lets non-Docker consumers (e.g. wheel-based deployments) run migrations from a vanilla install without vendoring migration files.

## Scope
- User-facing/API changes: new `agent-control-migrate` console script (subcommands: `upgrade`, `downgrade`, `current`, `history`, `heads`).
- Internal changes: wheel layout adds `agent_control_server/_alembic/` and `agent_control_server/_alembic.ini` (private, underscore-prefixed).
- Out of scope: source-tree dev workflow (Makefile / docker-entrypoint.sh) is unchanged - existing `alembic upgrade head` from `server/` still works.

## Risk and Rollout
- Risk level: low. Source-tree behavior unchanged; only adds new wheel content and a new entry point.
- Rollback plan: revert; no schema or runtime change.

## Testing
- [x] Added unit tests for the migrate entry point (`server/unit_tests/test_migrate.py`).
- [x] Verified the built wheel contains `agent_control_server/_alembic/versions/*.py` and `agent_control_server/_alembic.ini`.
- [x] Ran `ruff check` (server and root configs).
- [ ] Manually verified `agent-control-migrate` against a real database.

## Checklist
- [x] No doc updates needed (private bundle paths).

### DIFF
--- a/server/src/agent_control_server/migrate.py
+++ b/server/src/agent_control_server/migrate.py
@@ -1,0 +1,70 @@
+"""Run bundled Alembic migrations for agent-control-server.
+
+Exposed as the ``agent-control-migrate`` console script. The wheel ships
+its Alembic config and migration scripts under the package so this
+command works in any install location (Docker, venv, system Python).
+"""
+
+from __future__ import annotations
+
+import sys
+from pathlib import Path
+
+from alembic import command
+from alembic.config import Config
+
+import agent_control_server
+
+
+def _bundled_config() -> Config:
+    pkg_dir = Path(agent_control_server.__file__).parent
+    ini_path = pkg_dir / "_alembic.ini"
+    alembic_dir = pkg_dir / "_alembic"
+    if not ini_path.exists() or not alembic_dir.exists():
+        raise RuntimeError(
+            "Bundled Alembic resources not found. Expected "
+            f"{ini_path} and {alembic_dir}. The installed wheel is missing "
+            "migration assets."
+        )
+    cfg = Config(str(ini_path))
+    cfg.set_main_option("script_location", str(alembic_dir))
+    return cfg
+
+
+def main(argv: list[str] | None = None) -> int:
+    """Entry point for the ``agent-control-migrate`` console script.
+
+    With no arguments, runs ``upgrade head``. Supports a small subset of
+    Alembic commands sufficient for deploys and operational debugging:
+    ``upgrade``, ``downgrade``, ``current``, ``history``, ``heads``.
+    """
+    args = list(argv) if argv is not None else sys.argv[1:]
+    if not args:
+        args = ["upgrade", "head"]
+
+    cfg = _bundled_config()
+    op, *rest = args
+
+    if op == "upgrade":
+        command.upgrade(cfg, rest[0] if rest else "head")
+    elif op == "downgrade":
+        command.downgrade(cfg, rest[0] if rest else "-1")
+    elif op == "current":
+        command.current(cfg)
+    elif op == "history":
+        command.history(cfg)
+    elif op == "heads":
+        command.heads(cfg)
+    else:
+        print(f"agent-control-migrate: unknown command '{op}'", file=sys.stderr)
+        print(
+            "Supported: upgrade [rev], downgrade [rev], current, history, heads",
+            file=sys.stderr,
+        )
+        return 2
+
+    return 0
+
+
+if __name__ == "__main__":
+    raise SystemExit(main())
--- a/server/src/agent_control_server/migrate.py
+++ b/server/src/agent_control_server/migrate.py
@@ -48,7 +48,16 @@ def main(argv: list[str] | None = None) -> int:
     if op == "upgrade":
         command.upgrade(cfg, rest[0] if rest else "head")
     elif op == "downgrade":
-        command.downgrade(cfg, rest[0] if rest else "-1")
+        # Require an explicit revision: downgrade is destructive and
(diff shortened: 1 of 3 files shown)


server/src/agent_control_server/migrate.py
```diff
@@ -0,0 +1,79 @@
...
+
+def main(argv: list[str] | None = None) -> int:
+    """Entry point for the ``agent-control-migrate`` console script.
+
+    With no arguments, runs ``upgrade head``. Supports a small subset of
+    Alembic commands sufficient for deploys and operational debugging:
+    ``upgrade``, ``downgrade``, ``current``, ``history``, ``heads``.
+    """
+    args = list(argv) if argv is not None else sys.argv[1:]
+    if not args:
```
namrataghadi-galileo, line 73: P1:  Unhandled exceptions from Alembic commands propagate as tracebacks

In migrate.py: lines42-65 none of the command.*() calls are wrapped in try/except. If the DB is unreachable (OperationalError), a migration is in conflict (CommandError), or the lock can't be acquired, the exception propagates uncaught. The raise SystemExit(main()) entry point never executes, so the caller gets a traceback + exit code 1.

For an operational deploy CLI this is fragile — deploy pipelines typically expect a clean error message and a predictable non-zero exit code. Recommend wrapping each command dispatch in:

```
try:
    command.upgrade(cfg, ...)
except Exception as exc:
    print(f"agent-control-migrate: {exc}", file=sys.stderr)
    return 1
```

server/src/agent_control_server/migrate.py
```diff
@@ -0,0 +1,79 @@
...
+            return 2
+        command.downgrade(cfg, rest[0])
+    elif op == "current":
+        command.current(cfg)
+    elif op == "history":
+        command.history(cfg)
+    elif op == "heads":
+        command.heads(cfg)
+    else:
+        print(f"agent-control-migrate: unknown command '{op}'", file=sys.stderr)
```
namrataghadi-galileo: `--help/-h` prints "unknown command" instead of usage
`agent-control-migrate --help` falls into the else branch and exits with code 2. Operational teams will hit this immediately. A simple check for op in ("-h", "--help") printing the supported-commands string would fix it.

server/src/agent_control_server/migrate.py
```diff
@@ -0,0 +1,79 @@
...
+def main(argv: list[str] | None = None) -> int:
+    """Entry point for the ``agent-control-migrate`` console script.
+
+    With no arguments, runs ``upgrade head``. Supports a small subset of
+    Alembic commands sufficient for deploys and operational debugging:
+    ``upgrade``, ``downgrade``, ``current``, ``history``, ``heads``.
+    """
+    args = list(argv) if argv is not None else sys.argv[1:]
+    if not args:
+        args = ["upgrade", "head"]
```
namrataghadi-galileo, line 74: Extra arguments to upgrade/downgrade are silently dropped:
command.upgrade(cfg, rest[0] if rest else "head"). Running agent-control-migrate upgrade rev1 extra silently ignores extra. A typo like agent-control-migrate upgrade -1 (intended as a downgrade) passes "-1" to upgrade silently. Consider rejecting len(rest) > 1 with an error.

server/unit_tests/test_migrate.py
```diff
@@ -0,0 +1,115 @@
+"""Unit tests for the bundled-migrations entry point.
```
namrataghadi-galileo, line 1: Test gap: bare upgrade keyword with no revision is untested
covers main([]) (default → upgrade head) and main(["upgrade", "abc123"]) (explicit revision), but not main(["upgrade"]) (bare keyword → should also call upgrade head). That path is exercised by rest[0] if rest else "head" but has no dedicated test case.

server/src/agent_control_server/migrate.py
abhinav-galileo: Addressed in f21c959: alembic dispatch is now wrapped in `try/except Exception`; failures print `agent-control-migrate: <reason>` to stderr and return rc=1 instead of leaking a traceback. Covered by `test_main_returns_nonzero_for_command_errors`.

server/src/agent_control_server/migrate.py
abhinav-galileo: Addressed in f21c959: rewrote the parser as `argparse` with subparsers, so `--help`/`-h` prints usage natively (rc=0). Covered by `test_main_help_prints_usage`.

server/src/agent_control_server/migrate.py
abhinav-galileo: Addressed in f21c959: argparse now rejects extra positional args with a usage error. Covered by `test_main_rejects_extra_positional_args`. `--sql` for `upgrade`/`downgrade` is also wired through to `alembic.command.upgrade(..., sql=True)`.

server/unit_tests/test_migrate.py
abhinav-galileo: Addressed in f21c959: added `test_main_bare_upgrade_runs_upgrade_head` covering `main(["upgrade"])` → `upgrade head`.

namrataghadi-galileo (approved):

PR #210, opened 2026-05-04
ci(server): publish agent-control-server to PyPI

abhinav-galileo:
## Summary
The release workflow stages \`server/dist/\` and uploads it to GitHub release assets, but does not currently push the server wheel to PyPI. Models, evaluators, and the SDK are published; the server is not.

This change adds a \`publish-server\` job between \`publish-evaluators\` and \`publish-sdk\`, matching the pattern of the other publish jobs.

## Scope
- User-facing/API changes: \`agent-control-server\` becomes installable via \`pip install agent-control-server\` after the next release.
- Internal changes: dependency chain in the workflow updated to evaluators -> server -> sdk (server depends on evaluators at runtime, so evaluators publishes first).
- Out of scope: anything outside the release workflow.

## Risk and Rollout
- Risk level: low. New workflow job; failure to publish server doesn't affect previously published packages because \`pypa/gh-action-pypi-publish\` is per-job atomic.
- Rollback plan: revert the workflow file.

## Testing
- [x] Validated YAML structure locally.
- [ ] Verified by the next \`workflow_dispatch\` release run.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -148,11 +148,11 @@ jobs:
 
   publish-sdk:
     runs-on: ubuntu-latest
-    needs: [release, publish-server]
+    needs: [release, publish-evaluators]
     if: >-
       always() &&
       needs.release.outputs.released == 'true' &&
-      needs.publish-server.result == 'success'
+      needs.publish-evaluators.result == 'success'
     permissions:
       id-token: write
 
@@ -172,8 +172,10 @@ jobs:
 
   upload-release-assets:
     runs-on: ubuntu-latest
-    needs: [release, publish-sdk]
-    if: needs.release.outputs.released == 'true'
+    needs: [release, publish-models, publish-evaluators, publish-server, publish-sdk]
+    if: >-
+      always() &&
+      needs.release.outputs.released == 'true'
     permissions:
       contents: write
 
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -122,13 +122,37 @@ jobs:
           user: __token__
           password: ${{ secrets.PYPI_API_TOKEN }}
 
-  publish-sdk:
+  publish-server:
     runs-on: ubuntu-latest
     needs: [release, publish-evaluators]
     if: >-
       always() &&
       needs.release.outputs.released == 'true' &&
       needs.publish-evaluators.result == 'success'
+    permissions:
+      id-token: write
+
+    steps:
+      - name: Download built distributions
+        uses: actions/download-artifact@v4
+        with:
+          name: release-dists
+          path: .
+
+      - name: Publish agent-control-server to PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          packages-dir: release-dists/server/dist/
+          user: __token__
+          password: ${{ secrets.PYPI_API_TOKEN }}
+
+  publish-sdk:
+    runs-on: ubuntu-latest
+    needs: [release, publish-server]
+    if: >-
+      always() &&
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -122,6 +122,30 @@ jobs:
...
+      needs.publish-evaluators.result == 'success'
+    permissions:
+      id-token: write
+
+    steps:
+      - name: Download built distributions
+        uses: actions/download-artifact@v4
+        with:
+          name: release-dists
+          path: .
```
namrataghadi-galileo: download-artifact extracts a single named artifact directly into the configured path, and upload-artifact uses the uploaded directory as the artifact root. With the current upload at line 81 and download path: ., the server dists should land at server/dist/, not release-dists/server/dist/. The same mismatch appears in the other publish/release-asset jobs too.

.github/workflows/release.yaml
```diff
@@ -122,6 +122,30 @@ jobs:
...
+    runs-on: ubuntu-latest
+    needs: [release, publish-evaluators]
+    if: >-
+      always() &&
+      needs.release.outputs.released == 'true' &&
+      needs.publish-evaluators.result == 'success'
+    permissions:
+      id-token: write
+
+    steps:
```
namrataghadi-galileo, line 135: 135-145: 
This downloads the single release-dists artifact into the workspace root, then publishes from release-dists/server/dist/. Because the artifact was uploaded from release-dists/, its contents are rooted under server/dist after a direct download to ., so the publish step will likely look in a non-existent directory. Download to path: release-dists or change packages-dir to server/dist/.

.github/workflows/release.yaml
```diff
@@ -122,6 +122,30 @@ jobs:
...
           password: ${{ secrets.PYPI_API_TOKEN }}
 
+  publish-server:
+    runs-on: ubuntu-latest
+    needs: [release, publish-evaluators]
+    if: >-
+      always() &&
+      needs.release.outputs.released == 'true' &&
+      needs.publish-evaluators.result == 'success'
+    permissions:
```
namrataghadi-galileo, line 132: PyPA’s trusted publishing flow requires id-token: write and no explicit username/password, so the current mix grants extra permission while still depending on PYPI_API_TOKEN.

lan17:
Why do we need to publish this to pypi?

.github/workflows/release.yaml
abhinav-galileo: Good point. This workflow is using PyPI token auth today, so `id-token: write` is unused unless we move to trusted publishing.

I left it unchanged here because the existing `publish-models`, `publish-evaluators`, and `publish-sdk` jobs already use the same pattern: `id-token: write` plus `user: __token__` / `PYPI_API_TOKEN`. For this PR I want the new server publish job to mirror the existing release jobs rather than make one publish job differ from the rest.

I think the right follow-up is to clean this up consistently across all PyPI publish jobs: either remove `id-token: write` everywhere while staying on token auth, or migrate all packages to PyPI trusted publishing and remove the explicit username/password.

.github/workflows/release.yaml
abhinav-galileo: Fixed in 62d6b4e. I changed the download step to use `path: release-dists` for all publish/release-asset jobs, so the artifact is extracted under `release-dists/` and `packages-dir: release-dists/server/dist/` resolves consistently.

I kept the existing `release-dists/<package>/dist/` publish paths rather than switching only the server job to `server/dist/`, so all publish jobs use the same layout.

namrataghadi-galileo (approved):

PR #211, opened 2026-05-05
fix(sdk):  Get trace context from provider

namrataghadi-galileo:
## Summary
- Use a provider provided trace_id and span_id if available for observability first. 

## Scope
Get provider specific trace context if available else get current trace_id and span_id. 

## Risk and Rollout
- Risk level: low 
- Rollback plan:

## Testing
- [ x] Added or updated automated tests
- [ x] Ran `make check` (or explained why not)
- [x ] Manually verified behavior

### DIFF
--- a/sdks/python/src/agent_control/control_decorators.py
+++ b/sdks/python/src/agent_control/control_decorators.py
@@ -36,6 +36,7 @@ from dataclasses import dataclass, field
 from typing import Any, TypeVar
 
 from agent_control_models import Step, normalize_action
+from agent_control_telemetry import get_trace_context_from_provider
 
 from agent_control import AgentControlClient
 from agent_control.evaluation import check_evaluation_with_local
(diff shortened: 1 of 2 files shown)


sdks/python/src/agent_control/control_decorators.py
```diff
@@ -53,6 +54,25 @@ async def chat(message: str) -> str:
...
+    """Resolve trace/span IDs for a decorated control site.
+
+    External providers, such as the Galileo bridge, are authoritative because
+    they may reserve the concrete span ID that the eventual LLM/tool call will
+    use. Without a provider, keep the existing behavior: share an active trace
+    but create a fresh function span for this decorated call.
+    """
+    provider_context = get_trace_context_from_provider()
+    if provider_context is not None:
+        return provider_context["trace_id"], provider_context["span_id"]
```
abhinav-galileo, line 67: With a provider registered, this returns the provider's `span_id` verbatim, so every `@control()` call inside one provider-owned span now reports under the same `(trace_id, span_id)` pair — both the pre and post `_evaluate` POSTs, and any stacked decorators in the same trace. The docstring above frames this as intentional. Flagging as FYI: this is a semantic change from the old path, where `get_current_trace_id()` already returned the provider's trace_id and `_generate_span_id()` was minted fresh per decorated call (see tracing.py:147-150). Consumers that group control events by `(trace_id, span_id)` will now see them collapse onto one span; if the bridge keys events by `control_execution_id` this is fine. Worth a brief inline code comment so future readers don't expect a fresh span per decorator like the fallback branches still produce.

abhinav-galileo (approved):

PR #212, opened 2026-05-06
feat(server): migrate controls routes to auth framework

abhinav-galileo:
## Summary

- Move `/controls` and `/control-templates/render` onto operation-based auth.
- Keep `GET /controls/schema` public because it returns static metadata.
- Require `CONTROLS_CREATE` for validate and render because both use the authoring path.
- Preserve no-auth deployment mode.

## Behavior Change

- `POST /controls/validate` and `POST /control-templates/render` now require create access under the default header provider.

## Testing

- `make prepush` on the stacked branch in #215.

### DIFF
--- a/sdks/typescript/src/generated/models/security.ts
+++ b/sdks/typescript/src/generated/models/security.ts
@@ -6,12 +6,12 @@ import * as z from "zod/v4-mini";
 import { remap as remap$ } from "../lib/primitives.js";
 
 export type Security = {
-  apiKeyHeader: string;
+  apiKeyHeader?: string | undefined;
 };
 
 /** @internal */
 export type Security$Outbound = {
-  APIKeyHeader: string;
+  APIKeyHeader?: string | undefined;
 };
 
 /** @internal */
@@ -20,7 +20,7 @@ export const Security$outboundSchema: z.ZodMiniType<
   Security
 > = z.pipe(
   z.object({
-    apiKeyHeader: z.string(),
+    apiKeyHeader: z.optional(z.string()),
   }),
   z.transform((v) => {
     return remap$(v, {
--- a/server/src/agent_control_server/endpoints/controls.py
+++ b/server/src/agent_control_server/endpoints/controls.py
@@ -551,7 +551,6 @@ async def create_control(
     response_model=GetControlSchemaResponse,
     summary="Get control definition JSON schema",
     response_description="JSON schema for ControlDefinition",
-    openapi_extra={"security": []},
 )
 # Public schema metadata: no tenant state, no auth operation.
 async def get_control_schema() -> GetControlSchemaResponse:
--- a/server/src/agent_control_server/endpoints/controls.py
+++ b/server/src/agent_control_server/endpoints/controls.py
@@ -443,17 +443,13 @@ async def _validate_control_definition(
     summary="Render a control template preview",
     response_description="Rendered control preview",
 )
+# Rendering is part of the authoring flow, so require create access.
 async def render_control_template(
     request: RenderControlTemplateRequest,
     db: AsyncSession = Depends(get_async_db),
     _principal: Principal = Depends(require_operation(Operation.CONTROLS_CREATE)),
 ) -> RenderControlTemplateResponse:
-    """Render a template-backed control without persisting it.
-
-    Authorized as ``controls.create``: rendering is part of the authoring
-    flow (the result feeds the create / update endpoints), so a caller
-    who cannot create controls has no use for the materialized output.
-    """
+    """Render a template-backed control without persisting it."""
     control_def = await _render_and_validate_template_input(
         TemplateControlInput(
             template=request.template,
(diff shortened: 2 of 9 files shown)


server/src/agent_control_server/endpoints/controls.py
```diff
@@ -549,6 +552,7 @@ async def create_control(
     summary="Get control definition JSON schema",
     response_description="JSON schema for ControlDefinition",
 )
+# Public schema metadata: no tenant state, no auth operation.
```
namrataghadi-galileo, line 555: P1: `GET /controls/schema` is now anonymous. Previously gated by `require_api_key` at the router. The diff intentionally drops auth  and test_schema_endpoint_reachable_without_credentials pins it. Schema content is `ControlDefinition.model_json_schema()` (no tenant data), so likely fine, but worth explicit security sign-off because it changes the "every /api/v1/controls/* route requires creds" invariant operators may have depended on.

server/src/agent_control_server/endpoints/controls.py
```diff
@@ -563,7 +567,9 @@ async def get_control_schema() -> GetControlSchemaResponse:
     response_description="Control metadata and configuration",
 )
 async def get_control(
-    control_id: int, db: AsyncSession = Depends(get_async_db)
+    control_id: int,
+    db: AsyncSession = Depends(get_async_db),
+    _principal: Principal = Depends(require_operation(Operation.CONTROLS_READ)),
```
namrataghadi-galileo, line 572: The endpoint binds _principal for the auth side-effect but never threads _principal.namespace_key into the ControlService calls. list_controls_page doesn't even accept a namespace; get_active_control_or_404 accepts one but isn't called with it. Today this is fine because HeaderAuthProvider always returns DEFAULT_NAMESPACE_KEY, but the moment any other provider lands (and the framework is explicitly built for that), controls become cross-tenant readable/writable here.

Recommended fix in this PR (cheap, defensive): add a small _require_default_namespace(principal) helper in this module and call it from every endpoint that captures _principal. That converts a silent forward-compat hazard into a loud deploy-time failure the day a non-default namespace shows up. Threading it properly through the service layer can be a follow-up PR.

server/src/agent_control_server/endpoints/controls.py
abhinav-galileo: Intended. This endpoint returns static schema metadata only, with no tenant or resource data. The anonymous-access test pins that contract explicitly.

server/src/agent_control_server/endpoints/controls.py
abhinav-galileo: Handled in the stacked namespace PR: controls now pass principal.namespace_key into service calls, and the lower-level control loader requires a namespace. Keeping this PR focused on the route migration.

namrataghadi-galileo (approved):

sdks/typescript/src/generated/models/security.ts
```diff
@@ -6,12 +6,12 @@ import * as z from "zod/v4-mini";
 import { remap as remap$ } from "../lib/primitives.js";
 
 export type Security = {
-  apiKeyHeader: string;
+  apiKeyHeader?: string | undefined;
```
namrataghadi-galileo, line 9: Making /controls/schema public flipped apiKeyHeader from required → optional across the whole generated Security type. SDK consumers can now construct unauthenticated clients at compile time for every endpoint, not just this one. The server still enforces, but we've lost a useful type-level guardrail. Two options: (a) accept the looser type and note it in release notes, or (b) keep the schema endpoint inside the default security scheme and document that the server permits anonymous calls. Please confirm this was a conscious tradeoff.

server/src/agent_control_server/endpoints/controls.py
```diff
@@ -548,7 +551,9 @@ async def create_control(
     response_model=GetControlSchemaResponse,
     summary="Get control definition JSON schema",
     response_description="JSON schema for ControlDefinition",
+    openapi_extra={"security": []},
```
namrataghadi-galileo: Duplicate OpenAPI security override for `/controls/schema`

Locations:

`server/src/agent_control_server/endpoints/controls.py:554` — openapi_extra={"security": []}
and
`server/src/agent_control_server/main.py:350-357` — post-hoc patch in custom_openapi

Security is forced to [] in two places. openapi_extra already merges into the operation schema, so the custom_openapi block should be redundant. Either drop the main.py patch, or add a comment there explaining the FastAPI behavior that made it necessary — otherwise a future cleanup will delete one and the other will silently keep things working, hiding the regression if it ever comes back.

abhinav-galileo: Fixed in 3cd0ee1: removed the route-level openapi_extra and kept the custom OpenAPI patch as the single source of truth.

sdks/typescript/src/generated/models/security.ts
abhinav-galileo: Server auth remains enforced; /controls/schema is intentionally public. Removed the duplicate override in 3cd0ee1; stricter TS Security typing can be a follow-up if needed.

PR #214, opened 2026-05-07
feat(server): add runtime auth and namespace scoping

abhinav-galileo:
## Summary

- Add explicit runtime auth modes: `none`, `api_key`, and `jwt`.
- Scope controls, bindings, policies, agents, and evaluation by the authenticated principal namespace.
- Authorize runtime token exchange and runtime evaluation with target context.
- Add configurable extra forwarded headers for HTTP upstream auth.

## Stack

- Based on #212.

## Testing

- `make prepush` on the stacked branch in #215.


server/src/agent_control_server/auth_framework/providers/local_jwt.py
```diff
@@ -6,7 +6,7 @@
 returns a :class:`Principal` carrying the bound target. When a
 ``context_builder`` on the dependency surfaces ``target_type`` /
 ``target_id``, the provider also enforces that they match the token's
-binding — runtime endpoints get the request-target check for free.
+binding - runtime endpoints get the request-target check for free.
```
namrataghadi-galileo: P1: 
Line 58-76:
Target-binding check is conditional on the request supplying a target, not on the token carrying one. Today a runtime token bound to (env, prod) will pass authorization for /api/v1/evaluation if the caller simply omits target_type/target_id from the body — _evaluation_context returns {"target_type": None, "target_id": None}, and lines 61/69 short-circuit on the requested_target_* is not None guard. EvaluationRequest.target_type is Optional ([models/evaluation.py:46], so Pydantic will not reject this. Net effect: a target-bound token can be used for a namespace-wide eval that ignores the binding.

Suggested fix — invert the check so the token's binding is the source of truth:

```
if claims.target_type is not None or claims.target_id is not None:
    if context is None:
        raise ForbiddenError(...)
    if context.get("target_type") != claims.target_type:
        raise ForbiddenError("target_type does not match")
    if context.get("target_id") != claims.target_id:
        raise ForbiddenError("target_id does not match")
```
Add a regression test in test_runtime_token_exchange_endpoint.py next to test_evaluation_rejects_runtime_jwt_for_wrong_target that omits target_type/target_id from the eval body and asserts 403.

server/src/agent_control_server/main.py
```diff
@@ -309,9 +311,10 @@ async def attach_version_header(request, call_next):  # type: ignore[no-untyped-
 app.include_router(
     evaluation_router,
     prefix=api_v1_prefix,
-    dependencies=[Depends(require_api_key)],
+    dependencies=[Depends(get_api_key_from_header)],
 )
 
+# Evaluator discovery still uses the local credential dependency.
```
namrataghadi-galileo: Two routers were left on the legacy Depends(require_api_key) gate — evaluator_router (with the new "Evaluator discovery still uses the local credential dependency." comment) and observability_router. That means AGENT_CONTROL_AUTH_MODE=none and AGENT_CONTROL_AUTH_MODE=http_upstream silently do not apply to either route. An operator running in upstream-auth mode will still find /api/v1/evaluators / /api/v1/observability/... requiring an X-API-Key, and a none-mode deployment will return 401 from these even though the rest of the API is open.

Either migrate them in this PR (each is a one-line require_operation swap with a new EVALUATORS_READ / OBSERVABILITY_WRITE Operation member), or land a follow-up issue link in the comment so the deferral is explicit, not implied.

namrataghadi-galileo: Same applies to server/endpoints/observability line 45

server/src/agent_control_server/endpoints/evaluation.py
```diff
@@ -118,6 +117,20 @@ def _sanitize_evaluation_response(response: EvaluationResponse) -> EvaluationRes
     )
 
 
+async def _evaluation_context(request: Request) -> dict[str, object]:
```
namrataghadi-galileo, line 121: except Exception swallows every parse failure and returns {}, so a malformed body removes the target check from the runtime authorizer entirely (cf. Comment 1 above). FastAPI's body validator will eventually 422 the request, but the authorizer has already accepted the token by then — useful signal lost from logs, and harder to spot if the auth ordering ever changes.

Suggested: narrow the catch to (json.JSONDecodeError, UnicodeDecodeError), log at debug, and once Comment 1 lands the fail-open hole closes for free. Don't widen _evaluation_context to enforce — keep the policy in LocalJwtVerifyProvider, the context builder should just supply data.

server/tests/test_principal_namespace_flow.py
```diff
@@ -0,0 +1,168 @@
...
+
+    policy = ns_a.put(
+        "/api/v1/policies",
+        json={"name": f"policy-{uuid.uuid4().hex[:12]}"},
+    )
+    assert policy.status_code == 200, policy.text
+    policy_id = int(policy.json()["policy_id"])
+    attach_to_policy = ns_a.post(f"/api/v1/policies/{policy_id}/controls/{control_id}")
+    assert attach_to_policy.status_code == 200, attach_to_policy.text
+
```
namrataghadi-galileo, line 106: The new test pins create / list / get / delete cross-namespace isolation, but skips the mutating verbs on controls (PATCH /controls/{id}, PUT /controls/{id}/data), the binding by-key paths (PUT /control-bindings/by-key, POST /control-bindings/by-key:delete), and the agent association endpoints (POST /agents/{name}/policies/{policy_id} etc.). Each of those threads principal.namespace_key independently and a future refactor that drops the kwarg on one of them won't be caught.

Suggested: parametrize a single test_namespace_isolation_for_writes over (method, url, payload) tuples covering PATCH /controls, PUT /controls/{id}/data, PUT /control-bindings/by-key, POST /agents/{name}/policies/{policy_id}, and POST /agents/{name}/controls/{control_id} — assert each rejects (404 or empty list) when called from ns-b for an ns-a resource. ~40 lines.

server/src/agent_control_server/services/controls.py
```diff
@@ -128,10 +134,13 @@ async def get_control_or_404(
         self,
         control_id: int,
         *,
+        namespace_key: str | None = None,
```
namrataghadi-galileo: The sibling get_active_control_or_404 was tightened to make namespace_key required in this PR ([line 163], but get_control_or_404 still accepts namespace_key: str | None = None. Both internal callers (lines 240, 279) pass it, so making it required is a no-op today but closes the door on a future caller forgetting it.

Suggested: drop the default and make the parameter required, same as get_active_control_or_404.

server/src/agent_control_server/auth_framework/config.py
```diff
@@ -183,28 +204,97 @@ def _build_default_provider() -> RequestAuthorizer:
         timeout = float(os.environ.get(_UPSTREAM_TIMEOUT_ENV, "5.0"))
         token = os.environ.get(_UPSTREAM_TOKEN_ENV)
         token_header = os.environ.get(_UPSTREAM_TOKEN_HEADER_ENV, "X-Agent-Control-Service-Token")
+        extra_forward_headers = _parse_extra_forward_headers(
```
namrataghadi-galileo, line 216: _build_default_provider accepts header as an alias for api_key (good, backwards-compatible), but the RuntimeError lists only 'none', 'api_key', or 'http_upstream' — operators reading the error after a typo will think header is invalid. Same nit for _resolve_runtime_mode. Add 'header' to the error string, or drop the alias from the accepted set if you intend to retire it.

server/src/agent_control_server/auth_framework/providers/local_jwt.py
abhinav-galileo: Fixed in d0dd12c1. The verifier now treats the token binding as authoritative, so missing or mismatched request target context is denied. Added a regression test for target-less evaluation.

server/src/agent_control_server/main.py
abhinav-galileo: Agreed this should be explicit. I am leaving evaluator discovery and observability on the existing local-credential gate in this stack because migrating them adds new generic operation names to the upstream auth contract. I will track that as a separate auth-framework follow-up.

abhinav-galileo: Covered in the parent thread. Same follow-up: observability needs a generic operation added before moving it to the provider contract.

server/src/agent_control_server/endpoints/evaluation.py
abhinav-galileo: Fixed in d0dd12c1. The context builder now only catches JSON/Unicode decode failures and logs at debug; target enforcement stays in the JWT provider.

server/tests/test_principal_namespace_flow.py
abhinav-galileo: Added coverage in d0dd12c1 for the cross-namespace write paths: control patch/data update, binding by-key upsert/delete, and agent policy/control associations.

server/src/agent_control_server/services/controls.py
abhinav-galileo: Fixed in d0dd12c1. get_control_or_404 now requires namespace_key and always scopes the query by it.

server/src/agent_control_server/auth_framework/config.py
abhinav-galileo: Fixed in d0dd12c1. The accepted-mode error text now includes header wherever the alias is accepted.

server/src/agent_control_server/main.py
abhinav-galileo: Done in this stack. Evaluator discovery and observability now declare explicit auth-framework operations (`evaluators.read`, `observability.read`, `observability.write`). The router-level API-key extractor is kept only for OpenAPI security metadata, while local API-key mode preserves the previous authenticated-access behavior.

namrataghadi-galileo (approved):

server/src/agent_control_server/observability/store/base.py
```diff
@@ -119,11 +121,17 @@ class EventStore(ABC):
     """
 
     @abstractmethod
-    async def store(self, events: list[ControlExecutionEvent]) -> int:
+    async def store(
+        self,
+        events: list[ControlExecutionEvent],
+        *,
+        namespace_key: str = DEFAULT_NAMESPACE_KEY,
```
namrataghadi-galileo: `[server/src/agent_control_server/observability/store/base.py:124, 149, 167]` — and matching implementations in `[observability/store/postgres.py:110, 176, 397]` 

All three abstract methods (store, query_stats, query_events) declare namespace_key: str = DEFAULT_NAMESPACE_KEY as a kw-only default. This is the wrong default for a tenancy boundary — any future caller (a background job, a test, a new endpoint) that forgets to pass it gets silent cross-tenant access or writes to "default" with no warning. Make it required (namespace_key: str with no default) so the type system enforces it. The endpoint call sites already pass it explicitly, so this should be a no-op change today.

abhinav-galileo: Fixed in 902dbd3: observability store and ingestor methods now require an explicit namespace_key, with callers updated.

PR #215, opened 2026-05-07
feat(sdk): add runtime token auth

abhinav-galileo:
## Summary

- Add Python SDK runtime auth modes: `auto`, `none`, `api_key`, and `jwt`.
- Exchange target-bound runtime tokens for target-bearing evaluation requests.
- Cache runtime tokens per target, refresh before expiry, and retry once after a rejected runtime JWT.
- Keep no-target, no-auth, and API-key evaluation paths unchanged.
- Make the API-key header name configurable.

## Stack

- Based on #214.

## Testing

- `make prepush`

### DIFF
--- a/sdks/python/src/agent_control/runtime_auth.py
+++ b/sdks/python/src/agent_control/runtime_auth.py
@@ -1,0 +1,194 @@
+"""Runtime-token cache helpers for the Agent Control SDK."""
+
+from __future__ import annotations
+
+import asyncio
+import threading
+from collections.abc import Mapping, Sequence
+from dataclasses import dataclass
+from datetime import UTC, datetime, timedelta
+from typing import Literal
+
+RuntimeAuthMode = Literal["auto", "none", "api_key", "jwt"]
+
+_TokenKey = tuple[str, str, str]
+_DEFAULT_MAX_CACHE_ENTRIES = 256
+
+
+@dataclass(frozen=True)
+class RuntimeToken:
+    """Short-lived runtime token bound to one target."""
+
+    token: str
+    expires_at: datetime
+    server_url: str
+    target_type: str
+    target_id: str
+    scopes: tuple[str, ...]
+
+    def is_fresh(self, *, refresh_margin_seconds: int) -> bool:
+        """Return whether the token is usable beyond the refresh margin."""
+        refresh_at = datetime.now(UTC) + timedelta(seconds=refresh_margin_seconds)
+        return self.expires_at > refresh_at
+
+
+class RuntimeTokenCache:
+    """Thread-safe runtime token cache keyed by server and target."""
+
+    def __init__(self, *, max_entries: int = _DEFAULT_MAX_CACHE_ENTRIES) -> None:
+        if max_entries < 1:
+            raise ValueError("max_entries must be >= 1.")
+        self._max_entries = max_entries
+        self._tokens: dict[_TokenKey, RuntimeToken] = {}
+        self._jwt_unavailable = False
+        self._jwt_unavailable_targets: set[_TokenKey] = set()
+        self._exchange_locks: dict[_TokenKey, asyncio.Lock] = {}
+        self._lock = threading.Lock()
+
+    def get(
+        self,
+        server_url: str,
+        target_type: str,
+        target_id: str,
+        *,
+        refresh_margin_seconds: int,
+    ) -> RuntimeToken | None:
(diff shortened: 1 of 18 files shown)


sdks/python/src/agent_control/runtime_auth.py
```diff
@@ -0,0 +1,204 @@
...
+        *,
+        server_url: str | None = None,
+        target_type: str | None = None,
+        target_id: str | None = None,
+        globally: bool = False,
+        ttl_seconds: int = _DEFAULT_JWT_UNAVAILABLE_TTL_SECONDS,
+    ) -> None:
+        """Record that JWT runtime auth should not be attempted."""
+        with self._lock:
+            if globally:
```
namrataghadi-galileo, line 120: A single 404 on /auth/runtime-token-exchange in auto mode sets _jwt_unavailable = True for the lifetime of the cache. Only clear() (i.e., init()) resets it — the per-target marker has a TTL but the global one does not. A brief 404 during a deploy or proxy misroute permanently downgrades the SDK to API-key auth for the rest of the process, with no log line and no recovery path. Either give the global flag the same TTL behavior as the per-target marker, or narrow _GLOBAL_RUNTIME_TOKEN_FALLBACK_STATUSES so transient/edge 404s don't latch.

sdks/python/src/agent_control/runtime_auth.py
```diff
@@ -0,0 +1,204 @@
...
+
+    def __init__(self, *, max_entries: int = _DEFAULT_MAX_CACHE_ENTRIES) -> None:
+        if max_entries < 1:
+            raise ValueError("max_entries must be >= 1.")
+        self._max_entries = max_entries
+        self._tokens: dict[_TokenKey, RuntimeToken] = {}
+        self._jwt_unavailable = False
+        self._jwt_unavailable_targets: dict[_TokenKey, datetime] = {}
+        self._exchange_locks: dict[_LockKey, asyncio.Lock] = {}
+        self._lock = threading.Lock()
```
namrataghadi-galileo, line 57: [sdks/python/src/agent_control/runtime_auth.py:48, 137-141]
```
self._exchange_locks: dict[_LockKey, asyncio.Lock] = {}
...
def exchange_lock(self, server_url, target_type, target_id) -> asyncio.Lock:
    key = (server_url, target_type, target_id, id(asyncio.get_running_loop()))
    with self._lock:
        lock = self._exchange_locks.get(key)
        if lock is None:
            lock = asyncio.Lock()
            self._exchange_locks[key] = lock
        return lock
```
Tokens are LRU-capped by max_entries, but _exchange_locks has no cap and no eviction path. [test_runtime_token_cache_token_eviction_preserves_exchange_lock] explicitly pins this behavior, so a long-running process that talks to many distinct targets leaks asyncio.Lock objects (one per (server, target_type, target_id, loop_id)) forever. Either cap the lock dict with the same LRU policy, or wrap it in a WeakValueDictionary so unused locks get collected.

sdks/python/src/agent_control/runtime_auth.py
```diff
@@ -0,0 +1,204 @@
...
+    def clear(self) -> None:
+        """Clear every cached token and fallback marker."""
+        with self._lock:
+            self._tokens.clear()
+            self._jwt_unavailable = False
+            self._jwt_unavailable_targets.clear()
+
+    def exchange_lock(self, server_url: str, target_type: str, target_id: str) -> asyncio.Lock:
+        """Return the async exchange lock for one server and target."""
+        key = (server_url, target_type, target_id, id(asyncio.get_running_loop()))
```
namrataghadi-galileo: id() is only unique for currently alive objects. When an event loop is garbage-collected and Python reuses the address for a new loop (common in test suites that create/teardown loops per test, and possible in long-running servers that rebuild loops on failure), the new loop will inherit the prior loop's asyncio.Lock and asyncio will raise RuntimeError: <Lock ...> is bound to a different event loop on first use. Keep a strong reference to the loop via a WeakKeyDictionary[asyncio.AbstractEventLoop, dict[...]], or rebuild the lock when Lock._loop no longer matches.

sdks/python/src/agent_control/runtime_auth.py
abhinav-galileo: Fixed in b8e4ab5: runtime-auth fallback markers are TTL-bounded and scoped instead of process-permanent.

namrataghadi-galileo (approved):

PR #224, opened 2026-05-15
fix(server): prevent migration lock transactions

abhinav-galileo:
## Summary
- run Alembic with transaction_per_migration to match Galileo API migration behavior
- keep the migration advisory-lock connection in AUTOCOMMIT so it does not block CREATE INDEX CONCURRENTLY
- assert the advisory-lock engine is created with AUTOCOMMIT

## Validation
- uv run --package agent-control-server pytest server/tests/test_migrate.py -q
- uv run ruff check --config pyproject.toml server/alembic/env.py server/src/agent_control_server/migrate.py server/tests/test_migrate.py
- git diff --check

### DIFF
--- a/server/alembic/env.py
+++ b/server/alembic/env.py
@@ -1,9 +1,8 @@
 import os
 import sys
 
-from sqlalchemy import create_engine, pool
-
 from alembic import context
+from sqlalchemy import create_engine, pool
 
 sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), "..", "src")))
 
@@ -30,6 +29,7 @@ def run_migrations_offline() -> None:
         target_metadata=target_metadata,
         literal_binds=True,
         compare_type=True,
+        transaction_per_migration=True,
     )
     with context.begin_transaction():
         context.run_migrations()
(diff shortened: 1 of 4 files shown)


server/alembic/env.py
```diff
@@ -39,7 +39,12 @@ def run_migrations_online() -> None:
     url = _get_migration_url()
     connectable = create_engine(url, future=True, poolclass=pool.NullPool)
     with connectable.connect() as connection:
-        context.configure(connection=connection, target_metadata=target_metadata, compare_type=True)
+        context.configure(
```
namrataghadi-galileo, line 42: Previously, an upgrade head that traversed N migrations ran them in a single transaction — any failure rolled back everything. With transaction_per_migration=True, each migration commits independently, so a failure in migration #3 of 5 leaves migrations #1-2 committed and the DB in a partially-upgraded state. That's the right behavior for the production case (and a prerequisite if any future migration needs CREATE INDEX CONCURRENTLY or ALTER TYPE … ADD VALUE), but it changes recovery semantics. Please call this out in CHANGELOG and in the operator-facing migration docs so deployers know to check alembic current after a failed run, not assume the DB is at the pre-run revision.

abhinav-galileo: Good callout. This does change failed-upgrade recovery semantics: after a multi-revision upgrade failure, earlier revisions in that run may already be committed. I added an operator migration note in server/README.md to check the current Alembic revision before retrying, rather than assuming the DB rolled back to the pre-run revision. I kept this out of CHANGELOG.md since that file appears release-generated.

namrataghadi-galileo (approved):

PR #227, opened 2026-05-19
feat(server): allow host-owned logging setup

abhinav-galileo:
## Summary

- Add `AGENT_CONTROL_CONFIGURE_LOGGING`, defaulting to `true`, so OSS/default server behavior remains unchanged.
- When `AGENT_CONTROL_CONFIGURE_LOGGING=false`, Agent Control skips installing its own logging handlers.
- In host-owned logging mode, pass `log_config=None` to Uvicorn so embedding hosts keep their existing logging configuration.
- Add focused coverage for the new config setting, logging no-op behavior, and Uvicorn startup kwargs.

## Usage

Default behavior is unchanged:

```bash
AGENT_CONTROL_CONFIGURE_LOGGING=true
```

Embedding hosts that own process logging can disable Agent Control's logging setup:

```bash
AGENT_CONTROL_CONFIGURE_LOGGING=false
```

With that setting disabled:

- `configure_logging()` returns without changing root or Uvicorn handlers.
- `run()` passes `log_config=None` to `uvicorn.run(...)`.
- Existing host logging configuration remains in place.

## Validation

- `make server-test`
- `uv run --package agent-control-server ruff check --config pyproject.toml server/src server/tests/test_config.py server/tests/test_logging_utils.py server/tests/test_main_lifespan.py`

### DIFF
--- a/server/src/agent_control_server/main.py
+++ b/server/src/agent_control_server/main.py
@@ -37,7 +37,7 @@ from .errors import (
     http_exception_handler,
     validation_exception_handler,
 )
-from .logging_utils import configure_logging, get_uvicorn_log_level_name
+from .logging_utils import configure_logging, get_uvicorn_log_level_name, should_configure_logging
 from .observability.ingest import DirectEventIngestor
 from .observability.sinks import (
     EventStoreControlEventSink,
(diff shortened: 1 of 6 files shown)


server/src/agent_control_server/main.py
namrataghadi-galileo: configure_logging(default_level=_default_log_level())   # module top-level
This runs the first time anything imports agent_control_server.main. The new should_configure_logging() guard does the right thing — but only if AGENT_CONTROL_CONFIGURE_LOGGING=false is already in os.environ at import time. A host process that does:


import agent_control_server.main           # logging gets clobbered here
os.environ["AGENT_CONTROL_CONFIGURE_LOGGING"] = "false"
will lose its handlers once, silently. This is the most likely operational footgun for the actual use case this PR exists to support.

Recommendations (pick one):

- Document the contract in the new env var's description / in the PR description: "must be exported in the process environment before agent_control_server is imported".
- Better: move the module-level call into a lazy initializer (e.g., a _configure_logging_once() called from lifespan and run() only). Then the env var can be set up to the moment run() is invoked.
- Even better: add a regression test that imports the module with the env var set vs unset and asserts the root logger handler list is untouched in the false case.

namrataghadi-galileo (approved):

PR #232, opened 2026-05-25
fix(server): scope auth upstream CA to HTTP provider

abhinav-galileo:
## Summary
- add AGENT_CONTROL_AUTH_UPSTREAM_CA_FILE for http_upstream auth
- build an upstream-specific SSLContext instead of relying on process-wide SSL_CERT_FILE
- cover provider and env wiring behavior in auth framework tests

## Testing
- uv run --package agent-control-server pytest server/tests/test_auth_framework.py -q
- uv run --package agent-control-server ruff check --config pyproject.toml server/src/agent_control_server/auth_framework server/tests/test_auth_framework.py
- uv run --package agent-control-server mypy --config-file pyproject.toml server/src/agent_control_server/auth_framework
- pre-push hook ran make lint and make typecheck


server/src/agent_control_server/auth_framework/config.py
namrataghadi-galileo, line 184: Currently, a bad path in AGENT_CONTROL_AUTH_UPSTREAM_CA_FILE propagates as a raw FileNotFoundError/ssl.SSLError from ssl.create_default_context(cafile=...). The surrounding code already raises a typed RuntimeError for the missing-URL case (config.py:213) — this should match that pattern so operators get a message naming the env var and path, not a stack trace from ssl internals.

Suggested: wrap the ssl.create_default_context call (or do a Path(ca_file).is_file() precheck) in _build_default_provider and raise RuntimeError(f"{_UPSTREAM_CA_FILE_ENV}={ca_file!r} not found or unreadable.").

abhinav-galileo: Fixed. Bad `AGENT_CONTROL_AUTH_UPSTREAM_CA_FILE` values now raise a clear `RuntimeError` naming the env var/path, and I added a regression test for that case.

namrataghadi-galileo (approved):

PR #234, opened 2026-06-03
fix(evaluators)!: remove Luna2 evaluator

namrataghadi-galileo:
Remove legacy Luna2 evaluator and only support Luna evaluator with new API.


namrataghadi-galileo:
Screenshot 2026-06-03 at 6 14 59 PM

abhinav-galileo:
The `galileo.luna` create flow should seed the same defaults as the form. Right now it falls through to `{}` in `DEFAULT_EVALUATOR_CONFIGS`, so `fromConfig({})` opens with a blank threshold even though the default operator is `gte`. Please add a `galileo.luna` default config or make `fromConfig` fall back to `initialValues`.

abhinav-galileo:
Since `galileo.luna2` is a shipped evaluator id, this hard removal should be treated as a breaking release. Please mark the merge/release commit accordingly (for example `fix(evaluators)!: remove Luna2 evaluator` or a `BREAKING CHANGE:` footer) and include migration guidance, such as migrating to `galileo.luna` or pinning `agent-control-evaluator-galileo <8`.

abhinav-galileo (approved):

PR #235, opened 2026-06-05
fix(server): harden auth upstream connection handling

abhinav-galileo:
## Summary
- set auth upstream HTTP keepalive expiry to avoid reusing stale idle sockets while preserving HTTPX's default connection-count capacity
- expose generic auth upstream HTTP pool tuning env vars for keepalive expiry and connection limits
- add low-cardinality Prometheus metrics for auth upstream attempts and latency
- cover connection tuning validation and single-attempt network error behavior in auth framework tests

## Non-goals
- no retry logic
- no caching
- no API changes
- no deployment-specific behavior

## Tests
- uv run --package agent-control-server pytest -q --noconftest server/tests/test_auth_framework.py -k http_upstream
- uv run --package agent-control-server ruff check --config pyproject.toml server/src/agent_control_server/auth_framework/providers/http_upstream.py server/src/agent_control_server/auth_framework/config.py server/tests/test_auth_framework.py
- uv run --package agent-control-server mypy --config-file pyproject.toml server/src/agent_control_server/auth_framework/providers/http_upstream.py server/src/agent_control_server/auth_framework/config.py

### DIFF
--- a/server/src/agent_control_server/auth_framework/config.py
+++ b/server/src/agent_control_server/auth_framework/config.py
@@ -50,6 +50,11 @@ _UPSTREAM_TOKEN_ENV = "AGENT_CONTROL_AUTH_UPSTREAM_SERVICE_TOKEN"
 _UPSTREAM_TOKEN_HEADER_ENV = "AGENT_CONTROL_AUTH_UPSTREAM_SERVICE_TOKEN_HEADER"
 _UPSTREAM_EXTRA_FORWARD_HEADERS_ENV = "AGENT_CONTROL_AUTH_UPSTREAM_EXTRA_FORWARD_HEADERS"
 _UPSTREAM_CA_FILE_ENV = "AGENT_CONTROL_AUTH_UPSTREAM_CA_FILE"
+_UPSTREAM_KEEPALIVE_EXPIRY_ENV = "AGENT_CONTROL_AUTH_UPSTREAM_KEEPALIVE_EXPIRY_SECONDS"
+_UPSTREAM_MAX_CONNECTIONS_ENV = "AGENT_CONTROL_AUTH_UPSTREAM_MAX_CONNECTIONS"
+_UPSTREAM_MAX_KEEPALIVE_CONNECTIONS_ENV = (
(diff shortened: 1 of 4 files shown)


server/src/agent_control_server/auth_framework/config.py
```diff
@@ -219,6 +224,11 @@ def _build_default_provider() -> RequestAuthorizer:
             os.environ.get(_UPSTREAM_EXTRA_FORWARD_HEADERS_ENV)
         )
         ca_file = (os.environ.get(_UPSTREAM_CA_FILE_ENV) or "").strip() or None
+        keepalive_expiry_seconds = float(os.environ.get(_UPSTREAM_KEEPALIVE_EXPIRY_ENV, "1.0"))
```
namrataghadi-galileo: keepalive_expiry_seconds = float(os.environ.get(_UPSTREAM_KEEPALIVE_EXPIRY_ENV, "1.0"))
max_connections = int(os.environ.get(_UPSTREAM_MAX_CONNECTIONS_ENV, "100"))
max_keepalive_connections = int(
    os.environ.get(_UPSTREAM_MAX_KEEPALIVE_CONNECTIONS_ENV, "20")
)
A non-numeric env value (e.g. MAX_CONNECTIONS=abc) raises a raw ValueError: invalid literal for int().
A cross-field violation (e.g. MAX_KEEPALIVE_CONNECTIONS=30 with MAX_CONNECTIONS=20) raises the ValueError from [http_upstream.py:185-192].
Neither is caught by the except (OSError, ssl.SSLError) at [config.py:248], so the operator gets an opaque traceback instead of the actionable RuntimeError naming the env var that every other misconfig in _build_default_provider produces. Fix: wrap lines 227-244 (parse + HttpUpstreamConfig(...) construction) and re-raise as RuntimeError citing the offending variable.

abhinav-galileo: Fixed in 391954e: upstream numeric env parsing now raises RuntimeError with the env var name, connection-limit cross-field validation reports both relevant env vars, and HttpUpstreamConfig construction errors are wrapped during env-based startup.

namrataghadi-galileo (approved):

PR #236, opened 2026-06-08
feat(server): bound database connection pool

abhinav-galileo:
## Summary
- Add explicit async SQLAlchemy pool settings for non-SQLite databases.
- Bound pool size, overflow, checkout timeout, connect timeout, and statement timeout.
- Track checked-out SQLAlchemy connections with a Prometheus gauge.
- Release evaluation DB sessions before evaluator execution and dispose the engine on shutdown.

## Defaults
- `AGENT_CONTROL_DB_POOL_SIZE=5`
- `AGENT_CONTROL_DB_MAX_OVERFLOW=10`
- `AGENT_CONTROL_DB_POOL_TIMEOUT_SECONDS=5.0`
- `AGENT_CONTROL_DB_CONNECT_TIMEOUT_SECONDS=5`
- `AGENT_CONTROL_DB_STATEMENT_TIMEOUT_SECONDS=50.0`

## Validation
- `uv run --package agent-control-server ruff check --config pyproject.toml server/src server/tests/test_config.py server/tests/test_db.py server/tests/test_evaluation_error_handling.py`
- GitHub CI passing

### DIFF
--- a/server/src/agent_control_server/db.py
+++ b/server/src/agent_control_server/db.py
@@ -2,7 +2,6 @@ from collections.abc import AsyncGenerator
 from typing import Any
 
 from prometheus_client import Gauge
-from sqlalchemy import event
 from sqlalchemy.engine.url import make_url
 from sqlalchemy.ext.asyncio import AsyncSession, async_sessionmaker, create_async_engine
 from sqlalchemy.ext.asyncio.engine import AsyncEngine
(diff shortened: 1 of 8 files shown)


server/src/agent_control_server/db.py
```diff
@@ -13,11 +18,83 @@ class Base(DeclarativeBase):
...
+        connect_args: dict[str, Any] = {"connect_timeout": config.connect_timeout_seconds}
+        if statement_timeout_ms:
+            connect_args["options"] = f"-c statement_timeout={statement_timeout_ms}"
+        return connect_args
+    if driver == "asyncpg":
+        connect_args = {"timeout": float(config.connect_timeout_seconds)}
+        if statement_timeout_ms:
+            connect_args["server_settings"] = {"statement_timeout": str(statement_timeout_ms)}
+        return connect_args
+    return {}
```
namrataghadi-galileo, line 58: _build_connect_args returns {} for any driver other than psycopg/asyncpg, while pool kwargs still apply. It's documented ("Drivers without known timeout args get none") and a bare postgresql:// URL would fail create_async_engine anyway, so this is mostly theoretical — but a future valid async driver would lose timeouts without warning. A debug-level log when connect_args come back empty for a non-sqlite backend would make that visible.

server/src/agent_control_server/db.py
```diff
@@ -13,11 +18,83 @@ class Base(DeclarativeBase):
...
+    connect_args = _build_connect_args(url, config)
+    if connect_args:
+        kwargs["connect_args"] = connect_args
+    return kwargs
+
+
+def _instrument_connection_pool(engine: AsyncEngine) -> None:
+    """Track checked-out connections from the async engine's underlying pool."""
+    # Create the labeled series eagerly so idle processes scrape as 0, not absent.
+    SQLALCHEMY_CHECKED_OUT_CONNECTIONS.labels("default").set(0)
```
namrataghadi-galileo: The checkout/checkin event pair is the standard approach, but on a connection that's invalidated rather than returned, the matching checkin may not fire, leaking the counter over a long-lived process. pool_pre_ping=True reduces the frequency. If exactness matters, gauging off pool.checkedout() at scrape time avoids the drift; otherwise fine as-is.

server/src/agent_control_server/db.py
abhinav-galileo: Addressed in 80dcd0b: added a debug-level log when a non-sqlite backend has no known driver-level timeout connect args. It logs only backend/driver, not the DB URL.

namrataghadi-galileo (approved):

PR #237, opened 2026-06-11
fix(evaluators): configure Luna scorer API URL

abhinav-galileo:
## Summary
- Prefer optional `GALILEO_LUNA_API_URL` before `GALILEO_API_URL` for scorer API calls.
- Infer auth from exactly one configured credential; keep legacy `GALILEO_LUNA_AUTH_MODE` supported but undocumented.
- Emit a log warning when the deprecated auth-mode env var is used.
- Support optional CA verification and bounded HTTP connection pooling.
- Clarify examples: OSS users provide an API key; deployment-managed internal auth settings are injected by the deployment.

## Validation
- `uv run ruff check evaluators/contrib/galileo/src/agent_control_evaluator_galileo/luna/client.py evaluators/contrib/galileo/tests/test_luna_evaluator.py`
- `env PYTHONPATH=evaluators/contrib/galileo/src:evaluators/builtin/src:models/src uv run pytest evaluators/contrib/galileo/tests/test_luna_evaluator.py::TestGalileoLunaClient::test_client_warns_when_deprecated_auth_mode_env_is_set`
- Local route introspection confirms `/api/v1/agents/initAgent` and `/api/v1/agents/{agent_name}` are registered under CI DB env.

GitHub CI currently has one red `ci` job in `server/tests/test_init_agent.py::test_init_agent_route_exists`; rerun reproduced it. That server assertion is outside this PR diff.

### DIFF
--- a/evaluators/contrib/galileo/src/agent_control_evaluator_galileo/luna/client.py
+++ b/evaluators/contrib/galileo/src/agent_control_evaluator_galileo/luna/client.py
@@ -178,10 +178,9 @@ class GalileoLunaClient:
     """Thin HTTP client for Galileo Luna direct scorer invocation.
 
     Environment Variables:
-        GALILEO_API_SECRET_KEY or GALILEO_API_SECRET: Galileo API internal JWT signing secret.
+        GALILEO_API_SECRET_KEY: Galileo API internal JWT signing secret.
         GALILEO_API_KEY: Galileo API key fallback for public scorer invocation.
         GALILEO_LUNA_API_URL: Galileo Luna scorer invoke API URL override.
-        GALILEO_API_CLUSTER_URL: Internal Galileo API URL used in internal auth mode.
         GALILEO_API_URL: Galileo API URL fallback.
         GALILEO_LUNA_CA_FILE: CA bundle used to verify the scorer API endpoint, for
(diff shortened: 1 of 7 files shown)


evaluators/contrib/galileo/src/agent_control_evaluator_galileo/luna/client.py
```diff
@@ -56,6 +63,13 @@ def _env_auth_mode() -> AuthMode | None:
     value = os.getenv("GALILEO_LUNA_AUTH_MODE")
     if value is None or value.strip() == "":
         return None
+    warnings.warn(
```
namrataghadi-galileo: warnings.warn(..., DeprecationWarning) is filtered out by default in production Python, so an operator still relying on GALILEO_LUNA_AUTH_MODE likely never sees the notice. Consider also emitting logger.warning(...) so the deprecation actually surfaces in logs (the class already logs the selected auth mode right after).

abhinav-galileo: Addressed in 5ee5fc3 by emitting `logger.warning(...)` in addition to the `DeprecationWarning`, with a focused test asserting the warning reaches logs.

namrataghadi-galileo (approved):

PR #239, opened 2026-06-15
feat(engine): add evaluation runtime tuning

abhinav-galileo:
Summary:
- Exposes evaluation concurrency as a configurable engine setting.
- Adds Luna HTTP connection and client-pool tuning.
- Preserves Luna HTTP error metadata for evaluator failures.

Validation:
- git diff --check passed for the split diff.
- make engine-test did not reach tests because uv timed out fetching greenlet from the package index.

### DIFF
--- a/evaluators/contrib/galileo/src/agent_control_evaluator_galileo/luna/client.py
+++ b/evaluators/contrib/galileo/src/agent_control_evaluator_galileo/luna/client.py
@@ -6,6 +6,7 @@ import logging
 import os
 import ssl
 import warnings
+from asyncio import Lock
 from base64 import urlsafe_b64encode
 from hashlib import sha256
 from hmac import new as hmac_new
@@ -27,9 +28,11 @@ DEFAULT_INTERNAL_TOKEN_TTL_SECS = 3600
 DEFAULT_KEEPALIVE_EXPIRY_SECS = 1.0
 DEFAULT_MAX_CONNECTIONS = 100
 DEFAULT_MAX_KEEPALIVE_CONNECTIONS = 20
+DEFAULT_CLIENT_POOL_SIZE = 1
 LUNA_KEEPALIVE_EXPIRY_ENV = "GALILEO_LUNA_KEEPALIVE_EXPIRY_SECONDS"
 LUNA_MAX_CONNECTIONS_ENV = "GALILEO_LUNA_MAX_CONNECTIONS"
 LUNA_MAX_KEEPALIVE_CONNECTIONS_ENV = "GALILEO_LUNA_MAX_KEEPALIVE_CONNECTIONS"
+LUNA_CLIENT_POOL_SIZE_ENV = "GALILEO_LUNA_CLIENT_POOL_SIZE"
 PUBLIC_SCORER_INVOKE_PATH = "/scorers/invoke"
 INTERNAL_SCORER_INVOKE_PATH = "/internal/scorers/invoke"
 AuthMode = Literal["public", "internal"]
@@ -106,6 +109,7 @@ def _validate_connection_config(
     keepalive_expiry_seconds: float,
     max_connections: int,
     max_keepalive_connections: int,
+    client_pool_size: int,
 ) -> None:
     if keepalive_expiry_seconds < 0:
         raise ValueError(
@@ -124,6 +128,8 @@ def _validate_connection_config(
             f"{LUNA_MAX_KEEPALIVE_CONNECTIONS_ENV}={max_keepalive_connections} "
             f"must be less than or equal to {LUNA_MAX_CONNECTIONS_ENV}={max_connections}."
         )
+    if client_pool_size <= 0:
+        raise ValueError(f"{LUNA_CLIENT_POOL_SIZE_ENV}={client_pool_size} must be greater than 0.")
 
 
 def _as_float_or_none(value: JSONValue) -> float | None:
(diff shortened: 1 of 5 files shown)


evaluators/contrib/galileo/src/agent_control_evaluator_galileo/luna/client.py
```diff
@@ -316,26 +393,58 @@ def _derive_api_url(self, console_url: str) -> str:
...
+            timeout=httpx.Timeout(DEFAULT_TIMEOUT_SECS),
+            limits=httpx.Limits(
+                max_connections=self.max_connections,
+                max_keepalive_connections=self.max_keepalive_connections,
+                keepalive_expiry=self.keepalive_expiry_seconds,
+            ),
+            verify=verify,
+        )
+
+    def _select_pooled_client(self) -> httpx.AsyncClient:
```
namrataghadi-galileo, line 415: zeroDivisionError when close() races with the fast-path pool check

_get_client checks len(self._clients) == self.client_pool_size and then calls _select_pooled_client() — all without holding _client_lock. If close() runs concurrently and sets self._clients = [] between those two lines, _select_pooled_client executes self._clients[self._next_client_index % len(self._clients)] with an empty list, producing a ZeroDivisionError (modulo by zero).

This is realistic any time the client is shut down while requests are in flight (e.g. graceful shutdown with active evaluations).

evaluators/contrib/galileo/src/agent_control_evaluator_galileo/luna/client.py
```diff
@@ -78,6 +84,54 @@ def _env_auth_mode() -> AuthMode | None:
     raise ValueError("GALILEO_LUNA_AUTH_MODE must be either 'public' or 'internal'.")
 
 
+def _load_float_env(env_name: str, default: float) -> float:
```
namrataghadi-galileo, line 87: Empty string env vars crash instead of falling back to default

_load_float_env and _load_int_env check only if raw is None. If any of the GALILEO_LUNA_* vars is set to an empty string (common in Kubernetes secret rotation or template placeholders), float("") / int("") raises ValueError at client construction. The engine's _env_positive_int in [core.py:36](vscode-webview://16t29hbh4on5cp4pake9ice2nupl41l709oqr7nj6qcls0tqi2vl/engine/src/agent_control_engine/core.py#L36) explicitly skips empty strings with value.strip() == "". These should behave the same way.

evaluators/contrib/galileo/src/agent_control_evaluator_galileo/luna/client.py
```diff
@@ -316,26 +393,58 @@ def _derive_api_url(self, console_url: str) -> str:
...
-                    max_keepalive_connections=DEFAULT_MAX_KEEPALIVE_CONNECTIONS,
-                    keepalive_expiry=DEFAULT_KEEPALIVE_EXPIRY_SECS,
-                ),
-                verify=verify,
-            )
-        return self._client
+        """Get or create the next HTTP client."""
+        if self._client is not None and not self._client.is_closed:
+            return self._client
+        if self._client is not None and self._client.is_closed:
```
namrataghadi-galileo: self._client = None written outside _client_lock

The nullification at lines 425–426 happens before the lock is acquired. A concurrent coroutine can pass the self._client is not None guard at line 423, then see None after this write, and enter the lock to create a new client — leaking the one the first coroutine just cleared. In a pure asyncio single-threaded loop there's no await here so no actual preemption, but asyncio.Lock does not prevent re-ordering in a thread executor context and the pattern makes the double-check locking inside the lock incomplete: the in-lock re-check (line 436) re-reads _client after it may have been concurrently nulled by another caller, and for pool_size > 1 there's no re-check of _clients at all.

evaluators/contrib/galileo/src/agent_control_evaluator_galileo/luna/client.py
namrataghadi-galileo: Additionaly,
select_pooled_client called both inside and outside the lock — _next_client_index counter not synchronized

The fast path (line 433) calls it lock-free; the slow path (line 447) calls it while holding the lock. Two concurrent async tasks on the fast path read and increment _next_client_index without any guard. In the current CPython asyncio event loop there's no actual preemption across sequential statements, but the design breaks if _get_client is ever called from a thread executor, collapsing round-robin so all calls hit the same client.

namrataghadi-galileo:
[client.py:87 / core.py:31 / server] Three divergent env-int parsers across packages

_load_int_env (client.py), _env_positive_int (core.py), and presumably a third in the server package all parse integers from env vars but differ on: empty-string handling, exception type (ValueError vs RuntimeError), and positivity validation. Any fix to one is invisible to the others.

evaluators/contrib/galileo/src/agent_control_evaluator_galileo/luna/client.py
abhinav-galileo: Addressed in 35a12c1: pooled client selection now happens under _client_lock, so close() cannot clear _clients between the readiness check and selection. The same lock also guards _next_client_index updates.

evaluators/contrib/galileo/src/agent_control_evaluator_galileo/luna/client.py
abhinav-galileo: Addressed in 35a12c1: close/reset now runs under _client_lock, so _client is no longer nulled outside the lock and concurrent client creation sees a consistent state.

evaluators/contrib/galileo/src/agent_control_evaluator_galileo/luna/client.py
abhinav-galileo: Addressed in 35a12c1: Luna tuning env vars now treat empty/whitespace values as unset and fall back to defaults.

abhinav-galileo:
For the env parser consistency note: I fixed the immediate Luna parser inconsistency for empty values in 35a12c1. I did not extract a shared env parser across packages in this PR to avoid broadening the runtime-tuning change.

namrataghadi-galileo (approved):

PR #246, opened 2026-06-29
feat(sdk): add out of the box controls - part 1

namrataghadi-galileo:
## Summary
- Added Phase 1 out-of-box controls bootstrap tooling so startup can safely seed controls later without duplicating rows or blocking pod startup.
- Added evaluator gating, namespace-aware seeding, initial version creation, idempotency, duplicate-name race handling, and fail-open lifespan integration.

## Scope
- User-facing/API changes: None.
- Internal changes: New `agent_control_server.bootstrap` module, startup seeding hook, and bootstrap tests.
- Out of scope: Actual OOTB control definitions, Phase 2 regex/json/list templates, and Phase 3 Luna metadata seeding.

## Risk and Rollout
- Risk level: low
- Rollback plan: Remove the lifespan call to `seed_out_of_box_controls` and/or revert the bootstrap module and related tests.

## Testing
- [ ] Ran `make check` (not run because `uv` dependency resolution hit private index 401s)
- [x] Manually verified behavior via targeted tests, server lint, and server mypy using the existing `.venv`

## Checklist
- [x] Linked issue/spec (Phase 1 from OOTB controls technical spec)
- [x] Updated docs/examples for user-facing changes: N/A, no user-facing changes
- [ ] Included any required follow-up tasks in .md format: Phase 2 follow-up not yet added as a separate `.md` file

### DIFF
--- a/server/alembic/versions/d7e4a9b2c6f1_out_of_box_control_seed_index.py
+++ b/server/alembic/versions/d7e4a9b2c6f1_out_of_box_control_seed_index.py
@@ -9,8 +9,7 @@ Create Date: 2026-08-10 12:00:00.000000
 from __future__ import annotations
 
 import sqlalchemy as sa
-
-from alembic import op
+from alembic import context, op
 
 # revision identifiers, used by Alembic.
 revision = "d7e4a9b2c6f1"
@@ -37,7 +36,10 @@ def _index_is_invalid() -> bool:
 
 def upgrade() -> None:
     with op.get_context().autocommit_block():
-        if _index_is_invalid():
+        # Offline generation has no live PostgreSQL catalog to inspect. Emitting
+        # CREATE IF NOT EXISTS is safe there; invalid-index recovery remains an
+        # online-only retry path.
+        if not context.is_offline_mode() and _index_is_invalid():
             op.execute(f"DROP INDEX CONCURRENTLY IF EXISTS {_INDEX_NAME}")
         op.execute(
             f"""
--- a/server/alembic/versions/d7e4a9b2c6f1_out_of_box_control_seed_index.py
+++ b/server/alembic/versions/d7e4a9b2c6f1_out_of_box_control_seed_index.py
@@ -1,0 +1,53 @@
+"""add out-of-box control seed index
+
+Revision ID: d7e4a9b2c6f1
+Revises: f3a1c8d7e2b4
+Create Date: 2026-08-10 12:00:00.000000
+
+"""
+
+from __future__ import annotations
+
+import sqlalchemy as sa
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "d7e4a9b2c6f1"
+down_revision = "f3a1c8d7e2b4"
+branch_labels = None
+depends_on = None
+
+_INDEX_NAME = "idx_controls_namespace_seed_source"
+
+
+def _index_is_invalid() -> bool:
+    result = op.get_bind().execute(
+        sa.text(
+            """
+            SELECT NOT pg_index.indisvalid
+            FROM pg_index
+            WHERE pg_index.indexrelid = to_regclass(:index_name)
+            """
+        ),
+        {"index_name": _INDEX_NAME},
+    )
+    return bool(result.scalar_one_or_none())
+
+
+def upgrade() -> None:
+    with op.get_context().autocommit_block():
+        if _index_is_invalid():
+            op.execute(f"DROP INDEX CONCURRENTLY IF EXISTS {_INDEX_NAME}")
+        op.execute(
+            f"""
+            CREATE UNIQUE INDEX CONCURRENTLY IF NOT EXISTS {_INDEX_NAME}
+            ON controls (namespace_key, seed_source_id)
+            WHERE seed_source_id IS NOT NULL
+            """
+        )
+
+
+def downgrade() -> None:
+    with op.get_context().autocommit_block():
+        op.execute(f"DROP INDEX CONCURRENTLY IF EXISTS {_INDEX_NAME}")
--- a/server/alembic/versions/f3a1c8d7e2b4_out_of_box_control_seed_identity.py
+++ b/server/alembic/versions/f3a1c8d7e2b4_out_of_box_control_seed_identity.py
@@ -24,16 +24,18 @@ def upgrade() -> None:
         "controls",
         sa.Column("seed_opted_out_at", sa.DateTime(timezone=True), nullable=True),
     )
-    op.create_index(
-        "idx_controls_namespace_seed_source",
-        "controls",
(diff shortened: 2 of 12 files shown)


server/src/agent_control_server/bootstrap/out_of_box_controls.py
```diff
@@ -0,0 +1,235 @@
...
+
+
+async def _seed_one_control(
+    session: AsyncSession,
+    *,
+    namespace_key: str,
+    template: OutOfBoxControlTemplate,
+) -> str:
+    control_service = ControlService(session)
+    if await control_service.active_control_name_exists(template.name, namespace_key=namespace_key):
```
abhinav-galileo, line 199: Using an active, mutable name as the seed identity resurrects deleted controls and duplicates renamed ones on the next standalone startup. Could we persist an immutable seed/source ID plus an explicit opt-out tombstone, and cover delete/rename followed by reseeding?

server/src/agent_control_server/main.py
```diff
@@ -142,6 +146,24 @@ async def lifespan(app: FastAPI) -> AsyncGenerator[None, None]:
     available = list(list_evaluators().keys())
     logger.info(f"Evaluator discovery complete. Available evaluators: {available}")
 
+    try:
+        seed_result = await seed_out_of_box_controls(
```
abhinav-galileo: The try/except is fail-open only after this await returns. A database lock wait can hold startup before lifespan yields, especially because statement timeouts may be disabled. Please bound the whole bootstrap, or set a local lock timeout and retry later.

server/src/agent_control_server/bootstrap/out_of_box_controls.py
namrataghadi-galileo: Implemented. Out-of-box controls now persist an immutable, namespace-scoped source_id, and deleting a seeded control records an explicit opt-out tombstone. Reseeding resolves by source ID, so renamed controls are preserved and deleted controls are not resurrected. Added regression coverage for both rename → reseed and delete → reseed scenarios.

server/src/agent_control_server/main.py
namrataghadi-galileo: Addressed. The full OOTB bootstrap await is now bounded by a configurable timeout (10 seconds by default). If it blocks, the operation is cancelled and startup continues fail-open with a warning. Added lifespan coverage using a seed coroutine that never returns.

server/src/agent_control_server/bootstrap/out_of_box_controls.py
```diff
@@ -0,0 +1,260 @@
...
+
+    source_id: str
+    name: str
+    control: ControlDefinition
+    required_evaluators: frozenset[str] = field(default_factory=frozenset)
+
+    def __post_init__(self) -> None:
+        object.__setattr__(self, "source_id", _SLUG_NAME_ADAPTER.validate_python(self.source_id))
+        object.__setattr__(self, "name", _SLUG_NAME_ADAPTER.validate_python(self.name))
+        if not self.required_evaluators:
```
abhinav-galileo: [P2] Do not let explicit requirements replace condition dependencies

When `required_evaluators` is non-empty, this skips derivation from the control leaves, so a regex control with an extra Luna requirement can seed on a pod that has Luna but not regex. Always union explicit requirements with evaluator names derived from the condition, and add a mixed-requirement test.

server/alembic/versions/f3a1c8d7e2b4_out_of_box_control_seed_identity.py
```diff
@@ -0,0 +1,39 @@
...
+depends_on = None
+
+
+def upgrade() -> None:
+    op.add_column("controls", sa.Column("seed_source_id", sa.String(length=255), nullable=True))
+    op.add_column(
+        "controls",
+        sa.Column("seed_opted_out_at", sa.DateTime(timezone=True), nullable=True),
+    )
+    op.create_index(
```
abhinav-galileo: [P2] Build the seed index without blocking writes

This emits a transactional `CREATE UNIQUE INDEX`, which holds a write-blocking table lock for the scan; the preceding clone-lineage migration already uses `CREATE INDEX CONCURRENTLY` for this rollout pattern. Use the same autocommit/concurrent approach for upgrade and downgrade.

abhinav-galileo: [P2] Make the concurrent migration retry-safe

Entering `autocommit_block()` commits both unguarded `ADD COLUMN` statements before the index build and revision stamp. If the concurrent build or stamp fails, rerunning stops on `DuplicateColumn`; a failed concurrent build can also leave an invalid index that `IF NOT EXISTS` skips by name. Split the column and index changes into retry-safe revisions and recreate any invalid existing index.

server/src/agent_control_server/bootstrap/out_of_box_controls.py
namrataghadi-galileo: Fixed in a64020b. Explicit required_evaluators are now unioned with evaluator dependencies derived from condition leaves. I also added a regression test covering a regex control with an additional Luna requirement, ensuring it is skipped when regex is unavailable.

server/alembic/versions/d7e4a9b2c6f1_out_of_box_control_seed_index.py
```diff
@@ -0,0 +1,53 @@
...
+        sa.text(
+            """
+            SELECT NOT pg_index.indisvalid
+            FROM pg_index
+            WHERE pg_index.indexrelid = to_regclass(:index_name)
+            """
+        ),
+        {"index_name": _INDEX_NAME},
+    )
+    return bool(result.scalar_one_or_none())
```
abhinav-galileo, line 34: [P2] Keep offline migration generation working

`agent-control-migrate upgrade --sql` has no live result object, so this catalog probe reaches `scalar_one_or_none()` on `None` and aborts before emitting the index DDL. Branch on Alembic offline mode and emit an offline-safe index path, with a focused `--sql` regression test.

abhinav-galileo (approved):
If you want to address this - https://github.com/agentcontrol/agent-control/pull/246#discussion_r3758886694

server/alembic/versions/d7e4a9b2c6f1_out_of_box_control_seed_index.py
namrataghadi-galileo: Fixed. The migration now checks Alembic offline mode before querying pg_index. Offline upgrade --sql skips the catalog probe and emits CREATE UNIQUE INDEX CONCURRENTLY IF NOT EXISTS directly, while online execution retains invalid-index recovery. Added a focused offline SQL-generation regression test.

PR #247, opened 2026-06-29
feat(sdk): out of box controls - part 2

namrataghadi-galileo:
## Summary
- Added the Phase 2 static out-of-box control catalog so new installs seed useful non-Luna controls at startup.
- Included regex PII/shell controls, JSON approval controls, and list-based RBAC/tool allowlist controls using the Phase 1 idempotent bootstrap path.

## Scope
- User-facing/API changes: No API contract changes. Seeded controls will appear in the existing Controls tab/list after startup.
- Internal changes: Populated `OUT_OF_BOX_CONTROL_TEMPLATES` and expanded bootstrap/evaluator behavior tests.
- Out of scope: Phase 3 Luna scorer metadata lookup and org-scoped Luna control generation.

## Risk and Rollout
- Risk level: low
- Rollback plan: Revert the Phase 2 catalog additions in `server/src/agent_control_server/bootstrap/out_of_box_controls.py`, or remove specific templates from `OUT_OF_BOX_CONTROL_TEMPLATES`.

## Testing
- [ ] Ran `make check` (not run because `uv` dependency resolution hit private index/auth issues previously)
- [x] Manually verified behavior via targeted server tests, full server test suite, server lint, and server mypy using the existing `.venv`

## Checklist
- [x] Linked issue/spec (Phase 2 from OOTB controls technical spec)
- [x] Updated docs/examples for user-facing changes: N/A, no API/docs contract changes
- [x] Included any required follow-up tasks: Phase 3 Luna controls remain follow-up work


server/src/agent_control_server/bootstrap/out_of_box_controls.py
```diff
@@ -104,7 +101,250 @@ def skipped_count(self) -> int:
...
+        data=_leaf_control_payload(
+            description="Block tool commands matching common destructive shell operations.",
+            selector_path="input.command",
+            evaluator_name="regex",
+            evaluator_config={
+                "pattern": (
+                    r"\b(?:rm\s+-rf\s+(?:/|~|\$HOME)|sudo\s+rm\s+-rf|"
+                    r"mkfs(?:\.[a-z0-9]+)?|dd\s+if=[^\s]+\s+of=/dev/[^\s]+|"
+                    r"chmod\s+-R\s+777\s+/|chown\s+-R\s+[^|;&]*\s+/|"
+                    r"shutdown\s+(?:-h\s+)?now|reboot)\b"
```
abhinav-galileo: The shared trailing `\b` cannot match after `/` or `~`, so `rm -rf /`, `rm -rf ~`, `chmod -R 777 /`, and `chown -R root /` pass this control. Please use explicit end/separator handling per alternative and add these cases to the test.

server/src/agent_control_server/bootstrap/out_of_box_controls.py
```diff
@@ -104,7 +101,250 @@ def skipped_count(self) -> int:
...
+                        },
+                        {
+                            "required": ["amount"],
+                            "properties": {
+                                "amount": {"type": "number", "exclusiveMinimum": 10000}
+                            },
+                            "anyOf": [
+                                {
+                                    "required": ["approved"],
+                                    "properties": {"approved": {"const": True}},
```
abhinav-galileo: Could we avoid treating a caller-controlled `approved` flag as authorization evidence? A prompt-injected agent can set or replay it, while tools without this field cannot complete the flow. This should use a trusted approval artifact or host workflow bound to the action; the outbound template below has the same issue.

server/src/agent_control_server/bootstrap/out_of_box_controls.py
```diff
@@ -104,7 +101,250 @@ def skipped_count(self) -> int:
...
+                "recipient and message before sending."
+            ),
+            tags=["tool", "approval", "exfiltration", "json"],
+        ),
+    ),
+    OutOfBoxControlTemplate.from_payload(
+        name="oob-sensitive-tool-requires-approved-role",
+        data=_leaf_control_payload(
+            description="Deny sensitive tool use when runtime context has an unapproved role.",
+            selector_path="context.user.role",
```
abhinav-galileo: This fails open when `context.user.role` is missing: the selector returns `None`, and `ListEvaluator` returns `matched=False` before applying `match_on="no_match"`. Please fail closed for missing/unknown roles and source the role from trusted principal context.

server/src/agent_control_server/bootstrap/out_of_box_controls.py
```diff
@@ -104,7 +101,250 @@ def skipped_count(self) -> int:
...
+        ),
+    ),
+    OutOfBoxControlTemplate.from_payload(
+        name="oob-only-approved-tools-may-run",
+        data=_leaf_control_payload(
+            description="Deny tool calls whose step name is not in the approved tool list.",
+            selector_path="name",
+            evaluator_name="list",
+            evaluator_config={
+                "values": ["search", "web_search", "retrieve", "calculator"],
```
abhinav-galileo, line 332: Exact bare names do not match integrations that qualify tool names. Google ADK emits names such as `<agent>.web_search`, so this denies an explicitly approved tool. Please normalize tool identity across integrations or match a canonical raw name, with an integration-level test.

server/src/agent_control_server/endpoints/controls.py
```diff
@@ -257,6 +261,78 @@ def _validate_attachment_filters(
...
+async def _seed_out_of_box_controls_for_namespace(
+    db: AsyncSession,
+    *,
+    namespace_key: str,
+) -> None:
+    """Best-effort namespace seeding for browse/list surfaces."""
+    try:
+        if namespace_key == default_out_of_box_namespace_key():
+            return
+        if await _namespace_has_active_controls(db, namespace_key=namespace_key):
```
abhinav-galileo: Checking for any active control makes partial seeding permanent: templates commit separately, so one transient failure leaves rows that cause every later request to return here. It also skips tenants with one custom control. Please check expected seed identities or always run the idempotent per-template pass.

server/src/agent_control_server/endpoints/controls.py
```diff
@@ -257,6 +261,78 @@ def _validate_attachment_filters(
...
+            return
+        await seed_out_of_box_controls(
```
abhinav-galileo: This path is reached from a GET authorized only with `controls.read`, but it commits `Control` and `ControlVersion` rows. Could seeding move to trusted tenant provisioning, or require `controls.create` resolved to the same namespace, so read-only callers cannot trigger writes?

server/src/agent_control_server/bootstrap/out_of_box_controls.py
namrataghadi-galileo: Addressed by adding a canonical step identity to the shared Step contract. Integrations can retain qualified names for registration while controls evaluate the integration-independent tool name. Google ADK now sends web_search as canonical identity for names such as writer.web_search; Strands is wired similarly. Added ADK integration coverage and fallback tests for unqualified names.

server/src/agent_control_server/bootstrap/out_of_box_controls.py
namrataghadi-galileo: Fixed.

server/src/agent_control_server/bootstrap/out_of_box_controls.py
namrataghadi-galileo: Addressed by removing caller-controlled approval flags as bypasses from both templates. Qualifying high-value and outbound actions now always steer to a trusted host workflow, which must bind its approval artifact to the exact action. Tests verify that both top-level and nested approval flags cannot bypass the controls.

server/src/agent_control_server/bootstrap/out_of_box_controls.py
namrataghadi-galileo: Agreed. AC does not yet provide a trusted principal-role contract, so this control cannot securely distinguish authorized roles at the control layer. Making the selector fail closed would address missing values but would still trust caller-controlled context. We’ll remove this OOTB control for now and reintroduce it when RBAC/trusted principal context is supported.

server/src/agent_control_server/endpoints/controls.py
namrataghadi-galileo: Fixed by removing the namespace-wide active-control check and always running the idempotent per-template seed pass. This allows partially seeded namespaces to recover and ensures a tenant’s custom controls do not suppress OOTB seeding. Also added regression coverage for both scenarios.

server/src/agent_control_server/endpoints/controls.py
namrataghadi-galileo: This write-on-GET behavior is intentional for OOTB controls. The Galileo All controls page only calls GET /api/v1/controls?cloned=false; there is currently no tenant-provisioning call, and the organization namespace is resolved from the existing controls.read authorization. We need the fixed, server-owned OOTB controls to appear on the first page load, regardless of agent or logstream attachment. The reconciliation is idempotent, uses immutable seed identities, and respects deletion tombstones; callers cannot provide or modify the seeded payloads. Standalone/open-source deployments are seeded at startup, while this path covers dynamically discovered organization namespaces without an additional authorization request.

server/src/agent_control_server/bootstrap/out_of_box_controls.py
abhinav-galileo: [P1] Keep older SDKs compatible

This fixes current ADK calls, but pre-PR clients still send `writer.web_search` without `canonical_name`; the fallback returns that qualified name, so this exact allowlist matches `no_match` and denies an approved tool. Please normalize the legacy qualified form server-side or gate this control on an enforced minimum SDK version, and add a mixed-version test.

server/src/agent_control_server/bootstrap/out_of_box_controls.py
abhinav-galileo: [P1] Cover equivalent destructive command forms

The revised pattern still misses `rm -rf "$HOME"`, `rm -rf ~/`, and `rm -fr /`, so simple quoting, a trailing slash, or equivalent flag ordering bypasses this deny control. Please normalize command arguments or cover these equivalent forms in both the pattern and regression tests.

server/src/agent_control_server/bootstrap/out_of_box_controls.py
```diff
@@ -109,7 +106,209 @@ def skipped_count(self) -> int:
...
+            ),
+            tags=["tool", "approval", "exfiltration", "json"],
+        ),
+    ),
+    OutOfBoxControlTemplate.from_payload(
+        source_id="oob-only-approved-tools-may-run",
+        name="oob-only-approved-tools-may-run",
+        data=_leaf_control_payload(
+            description="Deny tool calls whose step name is not in the approved tool list.",
+            selector_path="canonical_name",
```
abhinav-galileo, line 329: [P1] Keep seeded controls rollback-compatible

This persists a selector that PR 246 rejects, while reconciliation never removes existing seed identities, so removing the template leaves previously seeded controls in place and a full code rollback makes their stored definitions invalid. Please add an explicit cleanup/data migration or establish a rollback floor that retains `canonical_name` support.

server/src/agent_control_server/endpoints/controls.py
```diff
@@ -1234,6 +1285,21 @@ async def list_controls(
...
+        cloned=cloned,
+        step_type=step_type,
+        stage=stage,
+        execution=execution,
+        tag=tag,
+        include_attachments=include_attachments,
+        attachment_target_type=attachment_target_type,
+        attachment_target_id=attachment_target_id,
+    ):
+        await _seed_out_of_box_controls_for_namespace(namespace_key=namespace_key)
```
abhinav-galileo, line 1346: [P2] Bound lazy namespace reconciliation

Every qualifying first-page GET waits for a fresh reconciliation; already-seeded namespaces issue one serial existence query per template, while cold or concurrent requests add per-template queries and commits. Bulk-fetch existing seed identities and bound or deduplicate the whole reconciliation so one controls read cannot absorb multiple statement timeouts.

server/src/agent_control_server/bootstrap/out_of_box_controls.py
abhinav-galileo: [P1] Normalize recursive-rm arguments

The new cases pass, but `rm -rf -- "$HOME"` and `rm --recursive --force "$HOME"` still bypass this deny rule, while the targetless `sudo` arm blocks safe `sudo rm -rf /tmp/cache`. Tokenize and normalize flags plus the actual target, including `--`, quoting, `sudo`, and long options, so the control distinguishes root or home deletion from scoped deletion.

server/src/agent_control_server/bootstrap/out_of_box_controls.py
abhinav-galileo: [P1] Make the rollback path run before old code serves

The added `downgrade()` does not run in a code rollback from PR 247 to PR 246 because both revisions remain `f3a1c8d7e2b4`, so old pods still encounter persisted `canonical_name` rows. Even an explicit downgrade filters by seed source, leaving clone-and-bind copies without `seed_source_id` incompatible. Stage reader support before seeding, or enforce a rollback floor or new migration boundary that cleans every persisted copy.

server/src/agent_control_server/bootstrap/out_of_box_controls.py
```diff
@@ -109,7 +110,300 @@ def skipped_count(self) -> int:
...
+    ),
+    OutOfBoxControlTemplate.from_payload(
+        source_id="oob-owasp-llm05-read-only-sql",
+        name="oob-owasp-llm05-read-only-sql",
+        data=_leaf_control_payload(
+            description=("Block SQL tool calls that are not a single read-only SELECT statement."),
+            selector_path="input.query",
+            evaluator_name="sql",
+            evaluator_config={
+                "allowed_operations": ["SELECT"],
```
abhinav-galileo, line 356: [P1] Treat SELECT INTO as a write

With this configuration, PostgreSQL `SELECT * INTO backup FROM users` returns no match even though it creates a table, because the SQL evaluator sees `SELECT` but does not classify the `INTO` node as DDL. Reject `SELECT ... INTO`, or defer this preset until the evaluator can, and add the exact regression before calling it read-only.

server/src/agent_control_server/bootstrap/out_of_box_controls.py
```diff
@@ -109,7 +110,300 @@ def skipped_count(self) -> int:
...
+    OutOfBoxControlTemplate.from_payload(
+        source_id="oob-owasp-llm10-bounded-sql-query",
+        name="oob-owasp-llm10-bounded-sql-query",
+        data=_leaf_control_payload(
+            description=("Block SQL queries without bounded results or with excessive complexity."),
+            selector_path="input.query",
+            evaluator_name="sql",
+            evaluator_config={
+                "require_limit": True,
+                "max_limit": 1000,
```
abhinav-galileo, line 377: [P1] Fail closed on indeterminate result bounds

`LIMIT $1`, `LIMIT (1000 + 1)`, and `LIMIT 1000 OFFSET $1` all pass this preset because the evaluator skips nonliteral limits and treats an unknown offset as zero. Reject indeterminate `LIMIT` or `OFFSET` expressions whenever a maximum is configured, or the 1000-row window remains bypassable.

models/src/agent_control_models/agent.py
```diff
@@ -150,6 +150,15 @@ class Step(BaseModel):
     name: str = Field(
         ..., min_length=1, description="Step name (tool name or model/chain id)"
     )
+    canonical_name: str | None = Field(
```
abhinav-galileo, line 153: [P2] Preserve the existing wildcard payload shape

`select_data(step, "*")` dumps the whole model, so current tool integrations now add `canonical_name` to every wildcard evaluation. Existing JSON controls with `additionalProperties: false` change from non-match to match and can deny previously valid calls; exclude this field from the legacy `*` view or version that selector shape.

server/src/agent_control_server/bootstrap/out_of_box_controls.py
namrataghadi-galileo: Full shell tokenization and argument normalization isn’t possible with the current out-of-box control setup, which only supports selector extraction followed by existing evaluators such as RE2 regex matching. RE2 cannot reliably model shell quoting, --, option normalization, and operand semantics. Supporting this correctly would require a dedicated shell-command evaluator or preprocessing layer, which is beyond this PR’s scope. I’ll track that separately rather than present the regex as comprehensive.

server/src/agent_control_server/bootstrap/out_of_box_controls.py
namrataghadi-galileo: Fixed in f94d14c. The server now derives canonical_name from the final segment of legacy qualified tool names when the field is absent, while preserving an explicitly supplied canonical name. Added mixed-version API coverage confirming both writer.web_search from older SDKs and current payloads with canonical_name="web_search" pass the same exact allowlist.

server/src/agent_control_server/bootstrap/out_of_box_controls.py
namrataghadi-galileo: PRs 246 and 247 are stacked and will be merged and deployed together. The deployment does not permit old and new application versions to serve concurrently, and this release establishes the minimum rollback version supporting canonical_name. Therefore, no unsupported reader will encounter the seeded selector.

server/src/agent_control_server/endpoints/controls.py
namrataghadi-galileo: Fixed in f94d14c. Reconciliation now bulk-loads all existing seed identities and conflicting names in one query instead of querying per template. Concurrent requests for the same namespace join one in-flight task, reconciliation is bounded by a 3-second timeout, and the global pending-task count is capped at 128. Failures remain best-effort so the controls read continues. Added tests for the single-query warm path, concurrent deduplication, and timeout cancellation.

server/src/agent_control_server/bootstrap/out_of_box_controls.py
namrataghadi-galileo: Fixed. PostgreSQL SELECT … INTO is now classified as a CREATE operation, so allowed_operations=["SELECT"] rejects it. Added the exact regression to both the SQL evaluator suite and the seeded read-only control coverage.

server/src/agent_control_server/bootstrap/out_of_box_controls.py
namrataghadi-galileo: Fixed. When max_limit or max_result_window is configured, nonliteral LIMIT values are rejected. Nonliteral OFFSET values are also rejected when a maximum result window is configured. Added the three exact bypass cases to evaluator and seeded-control regression coverage.

models/src/agent_control_models/agent.py
namrataghadi-galileo: Fixed. The * selector now excludes canonical_name, preserving its existing payload contract for strict JSON controls. The field remains available through the explicit canonical_name selector. Added unit coverage and an E2E regression using additionalProperties: false.

server/src/agent_control_server/bootstrap/out_of_box_controls.py
namrataghadi-galileo: Also, This cannot be implemented reliably with the current regex-only evaluator. Correctly handling --, quoting, sudo, flag ordering, and long options requires shell tokenization and normalized argument inspection, which is outside the current setup. I’ve fixed the cases that can be handled safely with the existing evaluator and will document the remaining limitation.

server/src/agent_control_server/bootstrap/out_of_box_controls.py
```diff
@@ -109,7 +110,299 @@ def skipped_count(self) -> int:
...
+    ),
+    OutOfBoxControlTemplate.from_payload(
+        source_id="oob-owasp-llm05-read-only-sql",
+        name="oob-owasp-llm05-read-only-sql",
+        data=_leaf_control_payload(
+            description=("Block SQL tool calls that are not a single read-only SELECT statement."),
+            selector_path="input.query",
+            evaluator_name="sql",
+            evaluator_config={
+                "allowed_operations": ["SELECT"],
```
abhinav-galileo, line 356: [P1] Do not label SELECT-only SQL as read-only

This allowlist still passes state-changing PostgreSQL calls such as `SELECT setval('seq', 42) LIMIT 1`, `SELECT pg_advisory_lock(42) LIMIT 1`, and executable user-defined functions, so the seeded "read-only" control does not provide that boundary. Rename and document it as SELECT-only, or require a read-only database role or transaction for the actual guarantee.

namrataghadi-galileo: Fixed. Renamed the preset to SELECT-only and updated its description to state that syntactic SELECT validation does not guarantee read-only execution. The description directs users to enforce a read-only database role or transaction for that boundary. Added regression coverage for setval, advisory locks, and user-defined SELECT functions.

PR #248, opened 2026-07-08
feat(evaluators): add defenseclaw evaluator

namrataghadi-galileo:
## Summary

Added two global external evaluator integrations:
- `defenseclaw.rule_pack`
- `defenseclaw.opa_policy`

- Added strict, typed Pydantic configuration models and distinct JSON Schemas.
- Added dedicated Agent Control UI forms with canonical serialization of hidden protocol fields.
- Both evaluator execution paths are intentionally no-ops and do not install or invoke DefenseClaw OSS packages.

## Scope

**User-facing/API changes:**
- Both evaluators appear independently in `GET /api/v1/evaluators`.
- Added typed UI forms for rule-pack and OPA-policy configuration.
- Rule-pack forms support adding and removing rules.
- Fixed protocol fields remain hidden in Form mode but are always serialized.
- DefenseClaw configurations are preserved in `get_server_controls()` output.

**Internal changes:**
- Added the `agent-control-evaluator-defenseclaw` contrib package.
- Added evaluator entry-point registration and package/release wiring.
- Added strict nested validation, including unique rule IDs and confidence bounds.
- Added default configurations and UI registry entries.
- Added focused backend and UI tests.

**Out of scope:**
- DefenseClaw transport, authentication, or provider invocation.
- Installing or calling DefenseClaw OSS packages.
- Changes to the separate `~/code/ui` Galileo UI repository.
- Orbit production packaging changes.

## Risk and Rollout

**Risk level:** medium

**Primary risk:**
- The evaluator package must be installed in the server runtime for catalog discovery and server execution.
- It must also be installed in SDK runtimes for controls using `execution: "sdk"`.

**Rollout requirements:**
- Publish `agent-control-evaluator-defenseclaw`.
- Add the package to the Orbit Agent Control image dependencies.
- Verify both evaluator identifiers appear in `GET /api/v1/evaluators` before activating controls.

**Rollback plan:**
- Remove or disable controls referencing the DefenseClaw evaluator names.
- Remove the DefenseClaw package from runtime dependencies.
- Revert the evaluator registry and UI form registration changes.

## Testing

- [x] Added or updated automated tests
  - Ran `make check` (full repository check was not run).
  - Ran 26 focused Python tests successfully.
  - Ran Ruff and mypy successfully.
  - Ran TypeScript, ESLint, and Prettier checks successfully.
  - Verified contrib package wiring, evaluator registration, and server catalog schemas.
  - Playwright could not run in the sandbox because it could not bind the local UI server port.

[...]

PR #249, opened 2026-07-09
feat(ui): show DefenseClaw enforcement spans in Monitor

vineethsai7:
> **Stack note:** This PR is intentionally based on `feature/68338-add-defense-claw-evaluator` and is a UI/observability follow-up to #248. It contains only the Monitor-tab commits. Once #248 lands, this PR can be retargeted to `main`.

## Problem

Agent Control's Monitor tab currently shows aggregate execution, trigger, and error statistics, but an operator cannot inspect the DefenseClaw enforcement event that caused a request to be blocked. That makes it difficult to verify the integration end to end or correlate a denial with its trace, span, request, rule, and blocked input.

## Solution

Query recent observability events for the selected agent and render them below the aggregate charts. The latest event is expanded automatically while still respecting a user's explicit accordion selection.

For DefenseClaw enforcement events, the view shows:

- trace, span, request, rule, action, and duration metadata;
- the exact blocked prompt, raw request body, and enforcement reason when DefenseClaw sends content;
- an `UNREDACTED` or `METADATA ONLY` badge so the privacy mode is explicit;
- a distinct event-query error state instead of presenting API failures as an empty result.

The query recomputes its relative start time on every refetch and does not retain a prior agent's event data while a new agent query is loading.

## What Changed

| Path | Change |
| --- | --- |
| `ui/src/core/api/client.ts` | Exposes the observability events query through the UI API client. |
| `ui/src/core/hooks/query-hooks/use-agent-events.ts` | Adds the agent-scoped rolling-window query hook. |
| `ui/src/core/page-components/agent-detail/monitor/index.tsx` | Integrates recent executions and event-query states into Monitor. |
| `ui/src/core/page-components/agent-detail/monitor/recent-executions.tsx` | Renders expandable enforcement events and exact blocked content. |
| `ui/tests/agent-stats.spec.ts` | Covers unredacted, metadata-only, error, selection, and refetch behavior. |
| `ui/tests/fixtures.ts` | Adds observability event fixtures and loads them in authenticated test setup. |
| `evaluators/contrib/defenseclaw/README.md` | Documents how DefenseClaw events appear in Agent Control. |

## Security and Privacy

- Event content is rendered as ordinary React text; no raw HTML rendering is used.
- Metadata-only events never render absent content fields.
- Removing previous-query retention prevents content from one agent flashing in another agent's view.
- Whether content is unredacted is controlled by the DefenseClaw integration; the UI labels the received mode explicitly.

## Review Feedback Addressed

- Authenticated Playwright fixtures now register the observability event route.
- Monitor imports follow the repository sort order.
- Accordion selection is derived without synchronously mutating state from an effect.
- Conditional rendering uses explicit `null` branches and cannot leak falsey numeric values into the DOM.

## Test Plan

- `npm run lint` — passed with no warnings or errors.
- `npm run typecheck` — passed.
- `npm run prettify:check` — passed.
- `npx playwright test tests/agent-stats.spec.ts tests/auth.spec.ts --reporter=line --workers=2` — 17 passed.
- Live DefenseClaw integration — a blocked request appeared in Monitor with its exact prompt, raw request, reason, trace ID, span ID, request ID, and matching rule IDs.

### DIFF
--- a/evaluators/contrib/defenseclaw/README.md
+++ b/evaluators/contrib/defenseclaw/README.md
@@ -14,3 +14,13 @@ pip install "agent-control-evaluators[defenseclaw]"
 The configuration contracts and JSON Schemas are implemented and discoverable. Both evaluator
 classes intentionally execute as no-ops: they return `matched=False` without contacting or
 installing any DefenseClaw runtime or OSS package.
+
+The complementary DefenseClaw watcher can emit post-decision
+`ControlExecutionEvent` records through the Agent Control SDK. The agent Monitor
+shows aggregate enforcement counts and a **Recent executions** drill-down with
+trace/span/request correlation, control and rule identity, action, and duration.
+When the DefenseClaw integration is enabled, it includes exact blocked input,
+raw request body, and enforcement reason by default. Monitor labels those spans
+`UNREDACTED` and renders the content in the execution drill-down. DefenseClaw
+operators can explicitly select metadata-only delivery. Treat access to these
+events as access to sensitive workload data.
--- a/ui/src/core/page-components/agent-detail/monitor/index.tsx
+++ b/ui/src/core/page-components/agent-detail/monitor/index.tsx
@@ -18,9 +18,11 @@ export const TIME_RANGE_SEGMENTS: TimeRangeOption[] = [
 ];
 
 import type { StatsResponse } from '@/core/hooks/query-hooks/use-agent-monitor';
+import { useAgentEvents } from '@/core/hooks/query-hooks/use-agent-events';
 import { useAgentMonitor } from '@/core/hooks/query-hooks/use-agent-monitor';
 
 import { ControlStatsTable } from './control-stats-table';
+import { RecentExecutions } from './recent-executions';
 import { SummaryCard } from './summary-card';
 import type { SummaryMetrics } from './types';
 import { mapTimeRangeTypeToTimeRange } from './utils';
@@ -49,6 +51,18 @@ function calculateSummary(
   };
 }
 
+const TIME_RANGE_MILLISECONDS: Record<string, number> = {
+  '1m': 60_000,
+  '5m': 5 * 60_000,
+  '15m': 15 * 60_000,
+  '1h': 60 * 60_000,
+  '24h': 24 * 60 * 60_000,
+  '7d': 7 * 24 * 60 * 60_000,
+  '30d': 30 * 24 * 60 * 60_000,
+  '180d': 180 * 24 * 60 * 60_000,
+  '365d': 365 * 24 * 60 * 60_000,
+};
+
 export function AgentsMonitor({
   agentUuid,
   timeRangeValue,
@@ -70,6 +84,13 @@ export function AgentsMonitor({
 
   // Calculate summary metrics
   const summary = useMemo(() => calculateSummary(stats), [stats]);
+  const eventWindowMs =
+    TIME_RANGE_MILLISECONDS[apiTimeRange] ?? TIME_RANGE_MILLISECONDS['1h'];
+  const { data: recentEvents, error: recentEventsError } = useAgentEvents(
+    agentUuid,
+    eventWindowMs,
+    { refetchInterval: 2000 }
+  );
 
   if (isLoading && !stats) {
     return (
@@ -126,6 +147,11 @@ export function AgentsMonitor({
           </Text>
         </Box>
       )}
+
+      <RecentExecutions
+        events={recentEvents?.events ?? []}
+        hasError={recentEventsError !== null}
+      />
     </Stack>
   );
 }
--- a/ui/src/core/page-components/agent-detail/monitor/index.tsx
+++ b/ui/src/core/page-components/agent-detail/monitor/index.tsx
@@ -17,8 +17,8 @@ export const TIME_RANGE_SEGMENTS: TimeRangeOption[] = [
   { label: '1Y', value: 'lastYear' },
 ];
 
-import type { StatsResponse } from '@/core/hooks/query-hooks/use-agent-monitor';
 import { useAgentEvents } from '@/core/hooks/query-hooks/use-agent-events';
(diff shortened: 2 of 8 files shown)


ui/tests/fixtures.ts
namrataghadi-galileo, line 1250: Can you also add `await mockRoutes.events(page);` here. 
Any auth-required test that reaches the monitor tab can escape mocks or show event API failures.

ui/src/core/page-components/agent-detail/monitor/index.tsx
```diff
@@ -18,9 +18,11 @@ export const TIME_RANGE_SEGMENTS: TimeRangeOption[] = [
 ];
 
 import type { StatsResponse } from '@/core/hooks/query-hooks/use-agent-monitor';
+import { useAgentEvents } from '@/core/hooks/query-hooks/use-agent-events';
```
namrataghadi-galileo, line 20: Nit: Can you change this order to:
```
import { useAgentEvents } from '@/core/hooks/query-hooks/use-agent-events';
import type { StatsResponse } from '@/core/hooks/query-hooks/use-agent-monitor';
import { useAgentMonitor } from '@/core/hooks/query-hooks/use-agent-monitor';
```

ui/src/core/page-components/agent-detail/monitor/recent-executions.tsx
```diff
@@ -0,0 +1,259 @@
...
+  hasError = false,
+}: RecentExecutionsProps) {
+  const latestKey = events.length > 0 ? executionKey(events[0]) : null;
+  const [openValue, setOpenValue] = useState<string | null>(latestKey);
+  const previousLatestRef = useRef<string | null>(latestKey);
+
+  useEffect(() => {
+    const previousLatest = previousLatestRef.current;
+    if (latestKey === null) {
+      setOpenValue(null);
```
namrataghadi-galileo: This component currently fails UI lint: React flags the synchronous setOpenValue calls inside useEffect, and jsx-no-leaked-render also flags the truthy conditional renders below. Can we remove the effect by deriving the Accordion value from selectedValue ?? latestKey, only update selectedValue from onChange, and replace {value && ...} renders with explicit value !== null ? ... : null checks?

evaluators/contrib/defenseclaw/README.md
```diff
@@ -14,3 +14,13 @@ pip install "agent-control-evaluators[defenseclaw]"
...
 classes intentionally execute as no-ops: they return `matched=False` without contacting or
 installing any DefenseClaw runtime or OSS package.
+
+The complementary DefenseClaw watcher can emit post-decision
+`ControlExecutionEvent` records through the Agent Control SDK. The agent Monitor
+shows aggregate enforcement counts and a **Recent executions** drill-down with
+trace/span/request correlation, control and rule identity, action, and duration.
+When the DefenseClaw integration is enabled, it includes exact blocked input,
+raw request body, and enforcement reason by default. Monitor labels those spans
+`UNREDACTED` and renders the content in the execution drill-down. DefenseClaw
```
yashsheth46, line 24: Why "UNREDACTED"? Is redaction the only action we usually take? How about UNCHANGED or ALLOWED?

yashsheth46 (commented):
1. Are we already logging some metadata from defenseclaw execution to be surfaced through agent-control monitoring?

2. Will that metadata be similar to what we populate in control spans, so splunk can pick it up as well?

evaluators/contrib/defenseclaw/README.md
vineethsai7: Good question. `UNREDACTED` is not the control action; it describes only the disclosure state of the event payload. DefenseClaw sets `metadata.content_unredacted=true` when the prompt, raw request body, and reason arrive without redaction. The separate action badge remains `DENY`. This feed reports final blocks, so `ALLOWED` would be incorrect, and `UNCHANGED` could be read as an enforcement decision.

There are actually three useful content states: `FULL CONTENT`, `REDACTED CONTENT`, and `METADATA ONLY` (`include_content=false`). Those labels are clearer than `UNREDACTED`, and the README should describe this as a content/privacy label rather than an action.

vineethsai7:
Thanks — answers to both questions:

1. **Yes, for final managed blocks.** The complementary DefenseClaw implementation in [cisco-ai-defense/defenseclaw#457](https://github.com/cisco-ai-defense/defenseclaw/pull/457) tails DefenseClaw's structured event stream and emits one SDK `ControlExecutionEvent` per Agent Control control whose managed rule IDs contributed to a final block. It deliberately does not report every DefenseClaw execution. The event carries the deterministic execution ID; trace/span/request correlation; Agent Control control ID/name; matching DefenseClaw rule IDs; pre/post stage; `llm_call`/`tool_call`; `action=deny`; confidence; timestamp and duration; evaluator; severity, direction, connector, policy/evaluation IDs when present; and, subject to DefenseClaw's redaction/content settings, the blocked input, raw body, and reason.

2. **The payload is compatible with Agent Control control spans and Splunk OTLP ingestion, but current delivery has an important boundary.** `ControlExecutionEvent` is the core observability model. The Python SDK's OTEL adapter converts it to an `agent_control.control_execution` span, preserves the supplied trace ID and parent span ID, maps the standard fields to `agent_control.*`, and maps event metadata to `agent_control.metadata.*`. Splunk Observability can ingest that shape through OTLP. DefenseClaw also sends the source structured verdict through its normal OTel/Splunk log fan-out, with trace/request/evaluation/rule correlation and DefenseClaw's global redaction policy.

The current DefenseClaw integration selects the SDK's default server sink so these events populate Agent Control Monitor. The built-in `otel` sink is an alternative selection, not an automatic dual-write. Therefore, today Agent Control Monitor receives the `ControlExecutionEvent`, while a configured Splunk destination receives DefenseClaw's native structured verdict event. If the requirement is for the *same* `agent_control.control_execution` span to go to both Monitor and Splunk simultaneously, we should add/configure an explicit composite fan-out sink rather than claim this PR already provides that dual delivery.

PR #253, opened 2026-07-23
feat: configurable runtime-token header to avoid gateway Authorization collision [HYBIM-866]

josjeon:
## Problem

When Agent Control runs behind the O11y gateway, the gateway overwrites `Authorization` with its own downstream identity JWT — clobbering AC's runtime-eval token on the hot path (HYBIM-866, under epic HYBIM-741). The runtime token and the gateway's identity JWT both want the `Authorization` header.

## Fix

Make the runtime token ride a **configurable header** on both sides, selected by `AGENT_CONTROL_RUNTIME_TOKEN_HEADER` (default `Authorization`). Behind the gateway, point both sides at a dedicated header (e.g. `X-Agent-Control-Runtime-Token`); the runtime token rides that header while the gateway keeps `Authorization` for its identity JWT — no collision.

- **`Authorization`**: keeps the mandatory `Bearer` scheme (existing contract).
- **Dedicated header**: carries the **raw** token (no `Bearer` prefix).
- **Default unchanged**: with the env unset, behavior is byte-identical to today.

### Server (verify side)

- `auth_framework/providers/local_jwt.py`: `LocalJwtVerifyProvider` takes a `header_name` (default `Authorization`) and reads the token from it. `Bearer` required only on `Authorization`; raw token accepted on a dedicated header.
- `auth_framework/config.py`: `_resolve_runtime_token_header()` reads `AGENT_CONTROL_RUNTIME_TOKEN_HEADER` (blank → default), passed into the provider when runtime mode is `jwt`.

### SDK (send side)

- `sdks/python/.../client.py`: `AgentControlClient` gains a `runtime_token_header` param (+ same env var). Sends the runtime token on the configured header — raw on a dedicated header, `Bearer` on `Authorization` (single `_format_runtime_token` helper is the sole authority for that rule).
- The API key is **preserved as the outer gateway credential**: when the runtime token rides a dedicated header, the API key still rides its own header (`X-API-Key`), so a request can authenticate at the gateway while the runtime JWT is verified by Agent Control. The existing same-header guard prevents any collision.
- High-level SDK: `agent_control.init()` accepts `runtime_token_header`, stores it in session state, and threads it into the evaluation clients (`evaluation.py`, `control_decorators.py`); it is cleared on reset.

## Configuration (behind the gateway)

```
# server
AGENT_CONTROL_RUNTIME_AUTH_MODE=jwt
AGENT_CONTROL_RUNTIME_TOKEN_SECRET=<secret>
AGENT_CONTROL_RUNTIME_TOKEN_HEADER=X-Agent-Control-Runtime-Token

# SDK (must match the server header)
AgentControlClient(..., runtime_token_header="X-Agent-Control-Runtime-Token")
# or agent_control.init(..., runtime_token_header="X-Agent-Control-Runtime-Token")
# or AGENT_CONTROL_RUNTIME_TOKEN_HEADER=X-Agent-Control-Runtime-Token
```

## Tests

- **Server** (`test_auth_framework.py`): default Bearer path; default rejects raw on `Authorization`; dedicated header reads raw token and coexists with a gateway `Authorization` JWT; Bearer also accepted on dedicated header; missing-header error names the configured header; blank `header_name` rejected; env resolver (unset → default, set → trimmed, whitespace → default).
- **Server, app-level** (`test_runtime_token_exchange_endpoint.py`): end-to-end through `/api/v1/evaluation` exercising config wiring + `Operation.RUNTIME_USE` routing — runtime token on a dedicated header with an outer `Authorization` gateway JWT is accepted; a token presented on `Authorization` is rejected (401) when a custom header is configured.
- **SDK** (`test_client.py`): header resolution (param/env/default, blank rejected, whitespace-env fallback); raw token on dedicated header with `Authorization` free while the API key is preserved on its own header; default sends `Bearer` on `Authorization`; auto-mode fallback keeps the API key when the exchange is unavailable.
- **SDK, high-level** (`test_init_validation.py`): `init()` stores `runtime_token_header` in session state; defaults to `None` when unset; validates the header up front (blank and bad field-name rejected before state is mutated); a positional call through `target_id` proves the new param is appended last and does not shift any existing slot (`controls_file` onward).

## Live validation on lab0 (through the real O11y gateway)

Validated the branch on lab0 behind the actual O11y api-gateway (the gateway that
owns `Authorization` for its own identity JWT — the real collision scenario).

- **Deploy:** built the branch server image, pushed it to a personal
  `docker-test.repo.splunkdev.net/user-<name>/agent-control:<tag>` namespace (a
  path a personal Artifactory token can write and lab0 can pull), then
  `kubectl set image` the lab0 `agent-control` deploy at it with
  `AGENT_CONTROL_RUNTIME_TOKEN_HEADER=X-Agent-Control-Runtime-Token`. Rolled back
  to the released image afterward.
- **Full flow returns 200, with a real org and log_stream.** `runtime-token-exchange`
  returns 200 and mints a real server-issued runtime JWT for the log_stream; the
  evaluation call carries that token on the **custom header** through the gateway
  and returns **200** with a real result (`is_safe: true`).
- **The collision is handled.** The same token presented on `Authorization` is
  rejected with 401 ("Missing X-Agent-Control-Runtime-Token"), because the server
  reads only the dedicated header. So even when the gateway overwrites
  `Authorization` with its own identity JWT, the runtime JWT survives on
  `X-Agent-Control-Runtime-Token`.
- **Correction from an earlier draft of this PR:** an initial run hit `502`
  (upstream `422`) on the exchange, which looked like a capability/provisioning gap.
  That was a test-input error, not a real gap: the request used a made-up
  `target_id` that was not a real log_stream, so the exchange could not resolve it.
  With a real log_stream id the whole flow returns 200. Nothing is blocked on
  provisioning.
- **Not exercised in this run:** a control actually firing (steer/deny) and control
  spans landing in Galileo (no control was bound to the log_stream, so the eval ran
  clean). That is separate from the header change and was already shown on the
  devstack.

## Security notes

- **Header isolation**: with a dedicated header configured, the verifier reads only that header — a token presented on `Authorization` is ignored, so it can't be smuggled past the gateway boundary.
- **No token leakage**: auth error messages reference the header name only, never the token value.
- Signature / scope / target-binding checks (`verify_runtime_token`) are unchanged.

## Backward compatibility

Safe. No backward compatibility issue.

The change is default-preserving. `runtime_token_header` defaults to `Authorization` on both the SDK and server, and an unset env var falls back to that default. With no config, behavior is byte-identical to today: the SDK still sends `Bearer <token>` on `Authorization`, and the server still requires the Bearer scheme there. All four default-path branches were traced to confirm.

One behavior change on the default path, and it's an improvement: a malformed `Authorization: Bearer ` with only trailing whitespace used to return an empty token and fail deep in signature verification. Now it's rejected up front with a clean `AUTH_MISSING_KEY`. No valid request changes.

### Operational notes (expected tradeoffs, not design risks)

- **Two-sided header contract.** Setting `AGENT_CONTROL_RUNTIME_TOKEN_HEADER` on only one side breaks runtime auth (401 on every eval). This is intrinsic to any configurable-header feature, not a regression, and it fails closed: a mismatch denies auth, never bypasses it. Neither side can validate the other's value since they run as separate processes, so the intended safeguard is a startup log of the resolved header on both sides, making a mismatch a quick log diff rather than a debugging session.
- **API-key retention on a dedicated header.** `_AgentControlAuth` now leaves the API key in place when the runtime token rides a dedicated header, so the key can act as the outer gateway credential. This is deliberate and required by the two-credential model: the API key authenticates at the gateway, and the runtime JWT is verified by Agent Control. The default path is unchanged. It is called out only because it is the one change on every request's auth path, which makes it the right spot to focus review.

## Notes / scope

- Backwards compatible: unset env → identical behavior. No change to the API-key or `none` runtime modes.
- Design follows the O11y api-service precedent: support both auth methods, gateway stays neutral, opt-in, default preserved.
- Server-side tests require the repo's Postgres test fixture (run in CI); the SDK suite runs standalone.
- Validated end-to-end on a devstack: with the SDK sending the token on `X-Agent-Control-Runtime-Token` and the server reading the same header, runtime JWT exchange, control steering, and control-span ingestion all worked. This confirms the SDK and server agree on the custom header. It does not yet exercise the gateway-collision path (the devstack has no O11y gateway overwriting `Authorization`).
- Gateway-collision validation on lab0 is done and green end-to-end (see "Live validation on lab0" above): with a real org and log_stream, `runtime-token-exchange` returns 200, the evaluation call rides `X-Agent-Control-Runtime-Token` through the real O11y gateway and returns 200 (`is_safe: true`), and the same token on `Authorization` is rejected 401. Control firing and span ingestion were not exercised in this run (no control bound to the log_stream), and are separate from the header change.

### DIFF
--- a/sdks/python/src/agent_control/__init__.py
+++ b/sdks/python/src/agent_control/__init__.py
@@ -113,6 +113,7 @@ from .observability import (
     write_events,
 )
 from .otel_sink import control_event_to_otel_span
+from .runtime_auth import validate_http_field_name
 from .tracing import (
     get_current_span_id,
     get_current_trace_id,
@@ -562,6 +563,16 @@ def init(
         raise ValueError(
             "target_type and target_id must be supplied together."
         )
+    # Validate the runtime-token header up front, before stopping the refresh
+    # loop or mutating shared session state. An explicit blank or a
+    # syntactically-invalid header must fail here rather than on the first
+    # later evaluation. (A None/unset value is fine; AgentControlClient resolves
+    # the env-var fallback and default when the eval client is built.)
+    if runtime_token_header is not None:
+        if not runtime_token_header.strip():
+            raise ValueError("runtime_token_header must not be blank.")
+        validate_http_field_name(runtime_token_header.strip())
+
     resolved_api_key_header = (
         api_key_header
         or os.getenv(AgentControlClient.API_KEY_HEADER_ENV_VAR)
--- a/sdks/python/src/agent_control/__init__.py
+++ b/sdks/python/src/agent_control/__init__.py
@@ -448,6 +448,7 @@ def init(
     server_url: str | None = None,
     api_key: str | None = None,
     api_key_header: str | None = None,
+    runtime_token_header: str | None = None,
     controls_file: str | None = None,
     steps: list[StepSchemaDict] | None = None,
     conflict_mode: Literal["strict", "overwrite"] = "overwrite",
@@ -488,6 +489,12 @@ def init(
         api_key: Optional API key for authentication (defaults to AGENT_CONTROL_API_KEY env var)
         api_key_header: Optional HTTP header name for API key authentication
             (defaults to AGENT_CONTROL_API_KEY_HEADER env var or X-API-Key)
+        runtime_token_header: Optional HTTP header the runtime token is sent on
+            for evaluation requests (defaults to AGENT_CONTROL_RUNTIME_TOKEN_HEADER
+            env var or Authorization). Point this at a dedicated header (e.g.
+            X-Agent-Control-Runtime-Token) when the server runs behind a gateway
+            that reserves Authorization for its own identity JWT. The server must
+            be configured to read the same header.
         controls_file: Optional explicit path to controls.yaml (auto-discovered if not provided)
         steps: Optional list of step schemas for registration:
                [{"type": "tool", "name": "search", "input_schema": {...}, "output_schema": {...}}]
@@ -588,6 +595,9 @@ def init(
         state.server_url = server_url or os.getenv('AGENT_CONTROL_URL') or 'http://localhost:8000'
         state.api_key = api_key
         state.api_key_header = resolved_api_key_header
+        # Stored raw (may be None); AgentControlClient resolves the env-var
+        # fallback and blank-handling so the rule stays in one place.
+        state.runtime_token_header = runtime_token_header
         state.runtime_token_cache.clear()
         state.target_type = target_type
         state.target_id = target_id
@@ -746,6 +756,7 @@ def _reset_state() -> None:
         state.server_url = None
         state.api_key = None
         state.api_key_header = None
+        state.runtime_token_header = None
         state.runtime_token_cache.clear()
         state.target_type = None
         state.target_id = None
--- a/sdks/python/src/agent_control/__init__.py
+++ b/sdks/python/src/agent_control/__init__.py
@@ -448,7 +448,6 @@ def init(
     server_url: str | None = None,
     api_key: str | None = None,
     api_key_header: str | None = None,
-    runtime_token_header: str | None = None,
     controls_file: str | None = None,
     steps: list[StepSchemaDict] | None = None,
     conflict_mode: Literal["strict", "overwrite"] = "overwrite",
@@ -459,6 +458,7 @@ def init(
     policy_refresh_interval_seconds: int = 60,
     target_type: str | None = None,
     target_id: str | None = None,
+    runtime_token_header: str | None = None,
     **kwargs: object
 ) -> Agent:
     """
@@ -489,12 +489,6 @@ def init(
         api_key: Optional API key for authentication (defaults to AGENT_CONTROL_API_KEY env var)
         api_key_header: Optional HTTP header name for API key authentication
             (defaults to AGENT_CONTROL_API_KEY_HEADER env var or X-API-Key)
-        runtime_token_header: Optional HTTP header the runtime token is sent on
-            for evaluation requests (defaults to AGENT_CONTROL_RUNTIME_TOKEN_HEADER
-            env var or Authorization). Point this at a dedicated header (e.g.
-            X-Agent-Control-Runtime-Token) when the server runs behind a gateway
-            that reserves Authorization for its own identity JWT. The server must
-            be configured to read the same header.
         controls_file: Optional explicit path to controls.yaml (auto-discovered if not provided)
         steps: Optional list of step schemas for registration:
                [{"type": "tool", "name": "search", "input_schema": {...}, "output_schema": {...}}]
@@ -513,6 +507,12 @@ def init(
             returned set when both are present.
         target_id: Optional opaque target identifier. Required iff target_type
             is also supplied.
+        runtime_token_header: Optional HTTP header the runtime token is sent on
+            for evaluation requests (defaults to AGENT_CONTROL_RUNTIME_TOKEN_HEADER
+            env var or Authorization). Point this at a dedicated header (e.g.
+            X-Agent-Control-Runtime-Token) when the server runs behind a gateway
+            that reserves Authorization for its own identity JWT. The server must
+            be configured to read the same header.
         **kwargs: Additional metadata to store with the agent
 
     Returns:
--- a/sdks/python/src/agent_control/client.py
+++ b/sdks/python/src/agent_control/client.py
@@ -15,6 +15,7 @@ from .runtime_auth import (
     RuntimeTokenCache,
     normalize_runtime_auth_mode,
     parse_runtime_token_exchange_response,
+    resolve_runtime_token_header,
 )
 
 _logger = logging.getLogger(__name__)
@@ -167,16 +168,11 @@ class AgentControlClient:
         self._runtime_auth_mode = normalize_runtime_auth_mode(configured_runtime_mode)
         # Explicit blank param is a hard error; a blank env var falls back to
         # the default (mirrors the server's _resolve_runtime_token_header).
-        if runtime_token_header is not None and not runtime_token_header.strip():
-            raise ValueError("runtime_token_header must not be blank.")
-        env_runtime_token_header = os.environ.get(_RUNTIME_TOKEN_HEADER_ENV_VAR)
-        if env_runtime_token_header is not None and not env_runtime_token_header.strip():
-            env_runtime_token_header = None
-        self._runtime_token_header = (
-            runtime_token_header
-            or env_runtime_token_header
-            or _DEFAULT_RUNTIME_TOKEN_HEADER
-        ).strip()
+        self._runtime_token_header = resolve_runtime_token_header(
+            runtime_token_header,
+            os.environ.get(_RUNTIME_TOKEN_HEADER_ENV_VAR),
+            default=_DEFAULT_RUNTIME_TOKEN_HEADER,
+        )
         # Bearer only on Authorization; a dedicated header carries the raw token
         # so it can't collide with the gateway's Authorization JWT. Must stay in
         # sync with LocalJwtVerifyProvider._require_bearer on the server.
@@ -305,7 +301,14 @@ class AgentControlClient:
             headers=request_headers,
         )
 
-        if _should_refresh_runtime_token(response) and runtime_authorization is not None:
+        runtime_token_on_dedicated_header = not self._runtime_token_use_bearer
+        if (
+            _should_refresh_runtime_token(
+                response,
+                runtime_token_on_dedicated_header=runtime_token_on_dedicated_header,
+            )
+            and runtime_authorization is not None
+        ):
             await response.aread()
             if target_type is not None and target_id is not None:
                 self._runtime_token_cache.remove(
@@ -327,7 +330,10 @@ class AgentControlClient:
                 headers=request_headers,
             )
             if (
-                _should_refresh_runtime_token(response)
+                _should_refresh_runtime_token(
+                    response,
+                    runtime_token_on_dedicated_header=runtime_token_on_dedicated_header,
+                )
                 and target_type is not None
                 and target_id is not None
             ):
@@ -507,10 +513,39 @@ class AgentControlClient:
         return token.token
 
 
-def _should_refresh_runtime_token(response: httpx.Response) -> bool:
+def _authenticate_flags_runtime_token(response: httpx.Response) -> bool:
+    """True when the response's WWW-Authenticate marks the token invalid.
+
+    Looks for the RFC 6750 ``error="invalid_token"`` challenge the runtime
+    verifier emits, so a refresh is driven by the server naming the token as
+    the problem, not by a bare status code.
+    """
+    return "invalid_token" in response.headers.get("WWW-Authenticate", "").lower()
+
+
+def _should_refresh_runtime_token(
+    response: httpx.Response,
+    *,
+    runtime_token_on_dedicated_header: bool = False,
+) -> bool:
+    """Decide whether a runtime-token refresh is warranted for this response.
+
+    When the runtime token rides ``Authorization`` (the default), a 401 is
+    unambiguous: the only credential on that header is the runtime token, so
+    treat it as expired and refresh.
+
+    When the token rides a dedicated header (behind a gateway that owns
+    ``Authorization`` for its own identity JWT), a bare 401 is ambiguous: it
+    may be the gateway rejecting its own credential, not our runtime token.
+    Evicting a still-valid token there just forces another exchange that hits
+    the same gateway 401 and hides the original error. So in that case only
+    refresh when the server explicitly flags the runtime token as invalid via
+    WWW-Authenticate (or a 403 invalid_token challenge).
+    """
     if response.status_code == 401:
+        if runtime_token_on_dedicated_header:
+            return _authenticate_flags_runtime_token(response)
         return True
     if response.status_code != 403:
         return False
-    authenticate = response.headers.get("WWW-Authenticate", "")
-    return "invalid_token" in authenticate.lower()
+    return _authenticate_flags_runtime_token(response)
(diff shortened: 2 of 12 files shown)


sdks/python/src/agent_control/client.py
```diff
@@ -35,19 +37,39 @@ def _runtime_cache_identity(api_key: str | None, api_key_header: str) -> str:
...
-            if self._header_name not in request.headers:
-                request.headers[self._header_name] = self._api_key
+        runtime_token_on_dedicated_header = (
+            self._runtime_token_header is not None
+            and self._runtime_token_header in request.headers
+        )
+        if (
+            self._api_key
+            and "Authorization" not in request.headers
+            and not runtime_token_on_dedicated_header
```
abhinav-galileo: [P1] Preserve the outer gateway credential

When `api_key_header` carries an outer gateway credential, this suppresses it as soon as the dedicated runtime-token header is present. The request then cannot authenticate at the gateway before Agent Control can verify its runtime JWT. Please retain the configured request credential when it uses a different header from the runtime token.

sdks/python/src/agent_control/client.py
```diff
@@ -99,6 +121,7 @@ def __init__(
         api_key: str | None = None,
         api_key_header: str | None = None,
         runtime_auth_mode: RuntimeAuthMode | str | None = None,
+        runtime_token_header: str | None = None,
```
abhinav-galileo: [P1] Preserve the existing positional argument order

This inserts `runtime_token_header` before `runtime_token_cache`, so an existing positional call binds a `RuntimeTokenCache` here and fails on `.strip()`. Please append the new option after the existing parameters, or otherwise preserve the prior positional slots.

sdks/python/src/agent_control/client.py
```diff
@@ -121,6 +144,13 @@ def __init__(
                 request auth when the exchange endpoint is unavailable. ``jwt``
                 requires a successful exchange. ``api_key`` and ``none`` keep
                 evaluation requests on the normal request-auth path.
+            runtime_token_header: HTTP header the runtime token is sent on.
```
abhinav-galileo, line 147: [P2] Expose this through the high-level SDK flow

`agent_control.init()` and the evaluation clients it creates do not accept or retain `runtime_token_header`, so high-level users can configure it only through a process-wide environment variable. Can we add it to the session state and pass it into those clients?

server/tests/test_auth_framework.py
```diff
@@ -1764,3 +1764,134 @@ async def test_teardown_auth_clears_registry():
...
+    with pytest.raises(AuthenticationError):
+        await provider.authorize(
+            _build_request(headers={"Authorization": token}),
+            Operation.RUNTIME_USE,
+            context={"target_type": "log_stream", "target_id": "ls-1"},
+        )
+
+
+@pytest.mark.asyncio
+async def test_local_jwt_custom_header_reads_raw_token():
```
abhinav-galileo, line 1819: [P2] Exercise the configured header through the runtime endpoint

These tests exercise `LocalJwtVerifyProvider` directly, but this change also depends on configuration wiring and `Operation.RUNTIME_USE` routing. Can we retain an app-level `/api/v1/evaluation` test that sends both an outer `Authorization` value and the dedicated runtime-token header?

sdks/python/src/agent_control/client.py
josjeon: Fixed in 1c9d292. `_AgentControlAuth` no longer suppresses the API key when the runtime token rides a dedicated header. The pre-existing `self._header_name not in request.headers` guard already prevents a same-header collision, so the extra suppression only ever fired when the two credentials were on *different* headers — exactly the case you flagged. Removed that special-casing (and the `runtime_token_header` param it needed), so the API key stays available on its own header as the outer gateway credential. Updated `test_runtime_evaluation_sends_raw_token_on_custom_header` to assert `X-API-Key` is now preserved.

sdks/python/src/agent_control/client.py
josjeon: Fixed in 1c9d292. Moved `runtime_token_header` to the end of `AgentControlClient.__init__` (after `transport`), so `runtime_token_cache` and the other existing parameters keep their prior positional slots. Docstring reordered to match.

sdks/python/src/agent_control/client.py
josjeon: Done in 1c9d292. `agent_control.init()` now accepts `runtime_token_header`, stores it in session state (`_state.py`), and threads it into the evaluation clients created in `evaluation.py` and `control_decorators.py`; `_reset_state()` clears it. Raw value is stored (env/blank resolution stays solely in `AgentControlClient`). Added `test_init_stores_runtime_token_header_in_state` and `test_init_defaults_runtime_token_header_to_none`.

server/tests/test_auth_framework.py
josjeon: Added in 1c9d292, in `test_runtime_token_exchange_endpoint.py` (which already drives the app + DB). `test_evaluation_accepts_runtime_jwt_on_configured_header_with_gateway_authorization` exchanges a real token and POSTs to `/api/v1/evaluation` with the token on `X-Agent-Control-Runtime-Token` and an outer `Authorization` gateway JWT — exercising config wiring + `Operation.RUNTIME_USE` routing end-to-end. Added a companion negative test asserting a token on `Authorization` is rejected (401) when a custom header is configured. Kept the provider-level tests here in `test_auth_framework.py` as well.

sdks/python/src/agent_control/__init__.py
```diff
@@ -448,6 +448,7 @@ def init(
     server_url: str | None = None,
     api_key: str | None = None,
     api_key_header: str | None = None,
+    runtime_token_header: str | None = None,
```
abhinav-galileo: [P1] Preserve `init()`'s positional argument order

This inserts `runtime_token_header` before the existing `controls_file` slot, so existing positional calls bind a controls path as the HTTP header and shift every later argument. Please append the new option after the existing parameters, as done for `AgentControlClient`, and add a positional compatibility test.

josjeon: Fixed in 270f42a. Moved `runtime_token_header` to the end of `init()`'s signature (after `target_id`, before `**kwargs`), so existing positional callers keep their slots — `controls_file` and everything after it are no longer shifted. Docstring reordered to match, and added `test_init_preserves_positional_argument_order` which does a positional call through `api_key_header` and asserts the values land in their original slots with `runtime_token_header` unset. Mirrors the earlier AgentControlClient fix.

sdks/python/src/agent_control/__init__.py
```diff
@@ -588,6 +595,9 @@ async def handle(message: str):
         state.server_url = server_url or os.getenv('AGENT_CONTROL_URL') or 'http://localhost:8000'
         state.api_key = api_key
         state.api_key_header = resolved_api_key_header
+        # Stored raw (may be None); AgentControlClient resolves the env-var
+        # fallback and blank-handling so the rule stays in one place.
+        state.runtime_token_header = runtime_token_header
```
namrataghadi-galileo, line 611: [P2] Validate before replacing the active SDK session

This stores runtime_token_header without validation, and the registration client does not consume it. As a result, a value such as whitespace lets init() succeed after stopping the previous refresh loop and replacing shared state; only the first evaluation later raises ValueError. Please normalize and validate the header before _stop_policy_refresh_loop() and before mutating session state.

sdks/python/src/agent_control/client.py
```diff
@@ -295,18 +341,22 @@ async def post_runtime_evaluation(
...
     ) -> dict[str, str] | None:
-        """Merge caller headers with an optional Bearer token."""
+        """Merge caller headers with an optional runtime token header."""
         if headers is None and runtime_authorization is None:
             return None
 
         merged = dict(headers or {})
         if runtime_authorization is not None:
-            merged["Authorization"] = runtime_authorization
+            merged[self._runtime_token_header] = runtime_authorization
```
namrataghadi-galileo, line 365: [P1] Distinguish gateway 401s from runtime-token failures

This makes the runtime token independent from Authorization, so Authorization can now carry a separate gateway identity. However, _should_refresh_runtime_token() below still treats every 401 as an expired runtime token. A gateway-generated 401 will therefore evict a valid token and force another exchange, which may fail with the same gateway 401 and obscure the original response. Please refresh only when WWW-Authenticate or the structured error identifies the Agent Control runtime token as invalid.

server/src/agent_control_server/auth_framework/config.py
```diff
@@ -378,12 +384,28 @@ def _build_runtime_provider(
...
+    """Header the runtime JWT verifier reads the token from (default ``Authorization``).
+
+    Behind a gateway that overwrites ``Authorization`` with its own identity
+    JWT, set a dedicated header so the two tokens don't collide. Blank falls
+    back to the default.
+    """
+    raw = os.environ.get(_RUNTIME_TOKEN_HEADER_ENV)
+    if raw is None or not raw.strip():
+        return DEFAULT_RUNTIME_TOKEN_HEADER
+    return raw.strip()
```
namrataghadi-galileo: [P2] Validate HTTP field-name syntax at startup

This accepts every nonblank value, including header names containing spaces, colons, or unsupported Unicode. The server then starts with a runtime-auth configuration that compliant clients cannot use, causing request-time authentication failures instead of a clear startup error. Please validate this against the HTTP field-name token grammar and apply the same validation in the SDK.

josjeon: Fixed in 5f76cf1. Added an RFC 7230 field-name validator (`validate_http_field_name`) and applied it to both sides: server `_resolve_runtime_token_header` and `LocalJwtVerifyProvider`, and the SDK (`resolve_runtime_token_header` in runtime_auth, used by `AgentControlClient` and `init()`). A header with spaces, colons, control chars, or non-ASCII now fails at config/startup and at client construction, not at request time. Added tests on both sides (server: provider + resolver reject invalid names; SDK: param + env).

sdks/python/src/agent_control/__init__.py
josjeon: Fixed in 5f76cf1. `init()` now validates runtime_token_header (blank + HTTP field-name grammar) in the early-validation block, before `_stop_policy_refresh_loop()` and before mutating shared session state. So an invalid/whitespace header raises up front instead of leaving a stopped refresh loop + replaced state that only errors on the first later evaluation.

sdks/python/src/agent_control/client.py
josjeon: Fixed in 5f76cf1. `_should_refresh_runtime_token` now takes `runtime_token_on_dedicated_header`. When the runtime token rides a dedicated header (so Authorization can carry a separate gateway identity), a bare 401 no longer evicts the token — it only refreshes if WWW-Authenticate flags the runtime token invalid (RFC 6750 error="invalid_token"), same as the 403 path. When the token is on Authorization (default), a 401 still refreshes as before. Added unit tests covering both header modes + the 403/other-status cases.

sdks/python/tests/test_init_validation.py
```diff
@@ -48,3 +53,87 @@ def test_init_exports_control_match() -> None:
...
+            # Positional call matching the pre-change signature order:
+            # agent_name, agent_description, agent_version, server_url,
+            # api_key, api_key_header
+            agent_control.init(
+                f"agent-{uuid4().hex[:12]}",
+                "desc",
+                "1.0.0",
+                "http://localhost:8000",
+                "key",
+                "X-API-Key",
```
abhinav-galileo, line 138: [P2] Exercise the shifted positional slot

This call stops at `api_key_header`, but the compatibility regression began at the next positional argument, `controls_file`, so this test would also pass with the broken signature. Please include a distinct seventh positional value and assert that it binds to `controls_file`, or cover the remaining pre-existing slots through `target_id`.

sdks/python/src/agent_control/client.py
abhinav-galileo: [P1] Complete the runtime-token 401 contract

This branch waits for `invalid_token`, but `LocalJwtVerifyProvider` raises `AuthenticationError` and the API error handler serializes every 401 as `WWW-Authenticate: ApiKey`. AC therefore never emits the signal needed to evict a rejected dedicated-header token; conversely, a gateway's own generic `invalid_token` challenge could trigger the wrong refresh. Please add an AC-runtime-specific marker, match that exact marker here, and cover AC rejection -> refresh and gateway 401 -> no refresh end to end.

sdks/python/tests/test_init_validation.py
josjeon: Good catch. Extended the positional call through `target_id` (16th slot) and now assert `target_type="env"` / `target_id="prod"` bind to their pre-change slots, so it transitively proves none of the intervening slots (`controls_file` onward) shifted. Confirmed the test fails if `runtime_token_header` is inserted before `target_id`. Fixed in fc5faf7.

sdks/python/src/agent_control/client.py
abhinav-galileo: [P2] Treat runtime-token rejection recovery as non-blocking

Revisiting the severity: the custom-header happy path works, and normal expiry is refreshed from the cached lifetime, so this is not a merge blocker. It would still be useful to add an Agent Control-specific rejection signal and retry contract for signing-secret rotation or other early token invalidation, either here or as a follow-up.

sdks/python/src/agent_control/__init__.py
abhinav-galileo: [P3] Resolve the environment fallback before replacing the session

The explicit value is now validated early, but the `None` path still resolves `AGENT_CONTROL_RUNTIME_TOKEN_HEADER` in the registration client after session state has changed, so an invalid environment value is treated as a connection failure. As a follow-up, consider resolving parameter over environment over default once before `_stop_policy_refresh_loop()` and passing that value to the internal clients.

abhinav-galileo (approved):

namrataghadi-galileo (approved):

josjeon:
https://github.com/agentcontrol/agent-control/pull/258 <-- merged with a new MR

PR #259, opened 2026-08-18
feat(evaluators): add AgentControl support for Orbit preset SLM scorer invocation

namrataghadi-galileo:
## Summary

- Adds AgentControl support for Orbit preset SLM scorer invocation and is paired with rungalileo/orbit#1720.
- Requires `scorer_id`; `scorer_label` is optional display metadata and `scorer_version_id` is a deprecated compatibility identifier.
- Calls `/api/v1/scorers/invoke` with a scoped internal HS256 JWT (`scorers.invoke`).
- Dual-writes legacy `inputs` and a structured `record` during the rollout, including selector-aligned input/output plus Step context, tools, and ground truth.
- Sends an Orbit execution timeout below the AgentControl HTTP deadline (8 seconds for the default 10-second deadline).

## Scope

- Orbit owns feature-flag routing between RuntimeScorer and the legacy adapter.
- Nine flat preset SLM scorers are validated for direct runtime execution.
- Existing Step context propagation through models, engine, integrations, and SDKs is retained.
- Broader preset SLM runtime coverage is deferred to Jira phases 2 and 3.
- Customer-owned Luna scorers are not supported in this release.
- No backend selector, Protect payload construction, Wizard configuration, trusted tenant identity, or Galileo record-tree construction is added.

## Risk and Rollout

- Risk level: medium.
- Deploy Orbit #1720 and keep its legacy adapter available for the tandem rollout.
- Roll back AgentControl to the legacy `inputs.query`/`inputs.response` contract if needed.
- Retain dual-write until older supported Orbit versions are retired.

## Testing

- [x] Galileo evaluator tests (116 passed)
- [x] Models tests (71 passed)
- [x] Engine tests (117 passed)
- [x] Affected Python SDK tests (103 passed)
- [x] Built-in evaluator tests (310 passed)
- [x] TypeScript SDK tests, lint, and typecheck
- [x] Repository lint, mypy, and `git diff --check`
- [ ] Full Python SDK target requires local PostgreSQL on `localhost:5432`

## Follow-up

Orbit #1720 must update all six AgentControl git references to commit `8a2b5d52717090ed7c65b627035f73d89c46f5ff` and regenerate requirements before the tandem merge.

### DIFF
--- a/engine/src/agent_control_engine/core.py
+++ b/engine/src/agent_control_engine/core.py
@@ -320,7 +320,7 @@ class ControlEngine:
                     timeout = DEFAULT_EVALUATOR_TIMEOUT
 
                 result = await asyncio.wait_for(
-                    evaluator.evaluate(data),
+                    evaluator.evaluate_with_context(data, request.step),
                     timeout=timeout,
                 )
         except TimeoutError:
--- a/evaluators/contrib/galileo/src/agent_control_evaluator_galileo/luna/client.py
+++ b/evaluators/contrib/galileo/src/agent_control_evaluator_galileo/luna/client.py
@@ -11,11 +11,14 @@ from hashlib import sha256
 from hmac import new as hmac_new
 from json import dumps
 from time import time
+from typing import Literal, cast, get_args
 from urllib.parse import urlsplit
 
 import httpx
 from agent_control_models import JSONObject, JSONValue, Step
 from pydantic import BaseModel, Field, PrivateAttr, model_validator
+
+from .config import ScorerInvokeConfig
 
 logger = logging.getLogger(__name__)
 
@@ -39,6 +42,20 @@ LUNA_KEEPALIVE_EXPIRY_ENV = "GALILEO_LUNA_KEEPALIVE_EXPIRY_SECONDS"
 LUNA_MAX_CONNECTIONS_ENV = "GALILEO_LUNA_MAX_CONNECTIONS"
 LUNA_MAX_KEEPALIVE_CONNECTIONS_ENV = "GALILEO_LUNA_MAX_KEEPALIVE_CONNECTIONS"
 LUNA_CLIENT_POOL_SIZE_ENV = "GALILEO_LUNA_CLIENT_POOL_SIZE"
+
+# These values mirror Orbit's StepType discriminator. Agent Control's generic
+# Step remains extensible; only the Galileo transport boundary is constrained.
+ScorerInvokeRecordType = Literal[
+    "llm",
+    "retriever",
+    "tool",
+    "workflow",
+    "agent",
+    "control",
+    "trace",
+    "session",
+]
+SUPPORTED_SCORER_INVOKE_RECORD_TYPES = frozenset(get_args(ScorerInvokeRecordType))
 
 
 def _b64url(data: bytes) -> str:
@@ -148,6 +165,25 @@ def _has_value(value: JSONValue) -> bool:
     return True
 
 
+def _serialize_dataset_output(value: JSONValue) -> str | None:
+    """Map Agent Control ground truth to Orbit's string dataset-output field.
+
+    Orbit accepts the original JSON value in the legacy ``inputs.ground_truth``
+    field, but its shared record hierarchy represents ``dataset_output`` as a
+    string. Compact, sorted JSON keeps the structured value lossless and lets
+    Orbit's semantic dual-write validator compare both representations.
+
+    Args:
+        value: JSON-compatible ground truth from the Agent Control step.
+
+    Returns:
+        The original string, compact JSON for a structured value, or ``None``.
+    """
+    if value is None or isinstance(value, str):
+        return value
+    return dumps(value, ensure_ascii=False, separators=(",", ":"), sort_keys=True)
+
+
 class ScorerInvokeInputs(BaseModel):
     """Input values sent to the Luna scorer invoke endpoint."""
 
@@ -158,14 +194,19 @@ class ScorerInvokeInputs(BaseModel):
 
 
 class ScorerInvokeRecord(BaseModel):
-    """Structured runtime record sent alongside legacy scorer inputs."""
+    """Caller-controlled subset of Orbit's partial runtime-record contract.
 
-    type: str = Field(min_length=1)
+    Identity, ownership, persistence, and execution IDs are intentionally not
+    represented here. Orbit hydrates those fields from trusted server context.
+    """
+
+    type: ScorerInvokeRecordType
+    name: str | None = None
     input: JSONValue = None
     output: JSONValue = None
     context: JSONObject | None = None
     tools: list[JSONObject] | None = None
-    dataset_output: JSONValue = None
+    dataset_output: str | None = None
 
 
 class ScorerInvokeRequest(BaseModel):
@@ -176,6 +217,7 @@ class ScorerInvokeRequest(BaseModel):
         scorer_version_id: Optional pinned scorer version identifier.
         scorer_label: Optional display/metadata label.
         inputs: Selected scorer input values.
+        record: Optional Orbit-compatible structured runtime record.
         config: Scorer-specific configuration, always emitted.
     """
 
@@ -184,7 +226,7 @@ class ScorerInvokeRequest(BaseModel):
     scorer_label: str | None = Field(default=None, min_length=1)
     inputs: ScorerInvokeInputs
     record: ScorerInvokeRecord | None = None
-    config: JSONObject = Field(default_factory=dict)
+    config: ScorerInvokeConfig = Field(default_factory=ScorerInvokeConfig)
 
     @model_validator(mode="after")
     def ensure_required_values(self) -> ScorerInvokeRequest:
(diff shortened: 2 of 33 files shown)


quinn-galileo (approved):

abhinav-galileo (commented):
**[P1][pre-existing] Make cached async clients event-loop-safe**

Cached evaluator instances retain pooled `httpx.AsyncClient` objects, while the synchronous decorator creates a new event loop per call. Reusing a keep-alive connection on a later loop can fail with `RuntimeError: Event loop is closed`, and cache eviction does not close these clients. This predates this PR and is not a regression introduced here; please track loop-scoped ownership or a persistent-loop lifecycle with awaited cleanup separately.

engine/src/agent_control_engine/core.py
```diff
@@ -320,7 +320,7 @@ async def _evaluate_leaf(
                     timeout = DEFAULT_EVALUATOR_TIMEOUT
 
                 result = await asyncio.wait_for(
-                    evaluator.evaluate(data),
+                    evaluator.evaluate_with_context(data, request.step),
```
abhinav-galileo, line 323: [P1] Raise dependency floors with the new evaluator hook

The engine now calls a method introduced by this change unconditionally, but the package constraints still allow evaluator releases that do not define it. An existing lockfile can therefore resolve successfully and then fail every control with `AttributeError`; raise the engine, SDK/server, and contributed evaluator/model floors to the first coordinated release.

evaluators/contrib/galileo/src/agent_control_evaluator_galileo/luna/client.py
```diff
@@ -185,6 +239,48 @@ def to_dict(self) -> JSONObject:
...
+    if step is None or step.type not in SUPPORTED_SCORER_INVOKE_RECORD_TYPES:
+        return None
+
+    # The membership check above narrows the runtime value to Orbit's known
+    # discriminator set, but static type checkers cannot infer that relationship.
+    record_type = cast(ScorerInvokeRecordType, step.type)
+    return ScorerInvokeRecord(
+        type=record_type,
+        name=step.name,
+        input=selected_input if selected_input is not None else step.input,
```
abhinav-galileo, line 296: [P1] Preserve the original structured record

Selector output may already have been coerced to legacy text before it reaches this helper. Writing it into `record.input` or `record.output` replaces the original `Step` side, so structured messages, tool calls, or nested objects can be lost. Keep `step.input` and `step.output` intact in `record`, and carry selector-derived legacy values separately with explicit server handling for intentional differences.

evaluators/contrib/galileo/src/agent_control_evaluator_galileo/luna/client.py
```diff
@@ -369,30 +466,47 @@ async def invoke(
...
+        # Accept dictionaries for source compatibility with the original client,
+        # but validate them against Orbit's authoritative allowlist locally.
+        invoke_config = (
+            ScorerInvokeConfig.model_validate(config)
+            if config is not None
+            else ScorerInvokeConfig()
+        )
         request_body = ScorerInvokeRequest(
             scorer_id=scorer_id,
             scorer_version_id=scorer_version_id,
```
abhinav-galileo, line 521: [P1] Verify the scorer version that actually executed

This sends a requested version, but the response model does not expose the resolved scorer ID/version and result metadata later records the configured value as executed. If the service resolves a different version, audit data is incorrect. Parse the executed identity and reject a mismatch, or label this value as requested until the endpoint returns it.

evaluators/contrib/galileo/src/agent_control_evaluator_galileo/luna/config.py
```diff
@@ -5,15 +5,38 @@
...
+        request_timeout_seconds: Optional upper bound for scorer execution in
+            Orbit. The Agent Control HTTP and evaluator deadlines must remain
+            longer than this value.
+    """
+
+    model_config = ConfigDict(extra="forbid")
+
+    threshold: float | None = None
+    score_threshold: float | None = None
+    request_timeout_seconds: float | None = Field(default=None, gt=0)
```
abhinav-galileo, line 37: [P1] Enforce an ordered deadline budget

`request_timeout_seconds` can equal or exceed `timeout_ms`, and when omitted the remote execution default can outlive the 10-second client and engine deadline. A valid score can be discarded while server work continues consuming capacity. Derive or validate `server execution < HTTP timeout < engine timeout` with explicit slack.

sdks/python/src/agent_control/integrations/google_adk/plugin.py
```diff
@@ -662,6 +667,33 @@ def _iter_tools(self, agent: Any) -> Iterable[Any]:
...
+            if not isinstance(agent_name, str) or not agent_name:
+                continue
+            step_name = self._resolve_step_name(
+                agent_name,
+                step_type="llm",
+                callback_context=None,
+                agent=agent,
+            )
+            definitions: list[dict[str, Any]] = []
+            for tool in self._iter_tools(agent):
```
abhinav-galileo, line 697: [P1] Capture tools resolved for this invocation

This iterates the raw tools bound to the agent during `bind`, but ADK entries can be callables or `BaseToolset` instances that are resolved asynchronously and may vary by request context. Tool-aware controls can therefore receive a stale list or a wrapper definition instead of the tools offered to the model. Build the payload from the current `LlmRequest.tools_dict` or the ADK async resolver, and retain the matching snapshot for the post-call check.

evaluators/contrib/galileo/src/agent_control_evaluator_galileo/luna/config.py
```diff
@@ -5,15 +5,38 @@
...
+    Attributes:
+        threshold: Legacy threshold accepted by Orbit. Agent Control still
+            applies its evaluator threshold locally.
+        score_threshold: Legacy score threshold accepted by Orbit.
+        request_timeout_seconds: Optional upper bound for scorer execution in
+            Orbit. The Agent Control HTTP and evaluator deadlines must remain
+            longer than this value.
+    """
+
+    model_config = ConfigDict(extra="forbid")
```
abhinav-galileo, line 33: [P2] Preserve forward-compatible scorer configuration

This replaces the prior JSON object with a closed three-key model, so persisted controls and future scorer-specific options fail locally before reaching the scorer. Keep the transport field JSON-extensible, or preserve unknown keys during migration and validate them at the scorer adapter that owns them.

models/src/agent_control_models/agent.py
```diff
@@ -142,6 +142,8 @@ def validate_type(cls, v: str) -> str:
 class Step(BaseModel):
     """Runtime payload for an agent step invocation."""
 
+    model_config = {"frozen": True}
```
abhinav-galileo, line 145: [P2] Snapshot Step at the evaluator boundary

`frozen=True` changes the public Python API while providing only shallow immutability: nested dictionaries and lists remain writable, and the same `Step` is shared across concurrent evaluators. Keep the public model source-compatible and pass a deep per-evaluation snapshot or a recursively immutable view at the engine boundary.

namrataghadi-galileo:
@abhinav-galileo Review follow-up for `8a2b5d5`:

- [Structured record](#discussion_r3865624446): addressed using Orbit #1720 as the authoritative contract. Orbit rejects conflicts between legacy `inputs.query`/`inputs.response` and `record.input`/`record.output`, so selector-selected values are intentionally dual-written. The full `Step` still supplies context, tools, dataset output, type, and name; the contract test verifies those structured fields are preserved.
- [Scorer version identity](#discussion_r3865624452): addressed. Metadata now uses `requested_scorer_version_id`, and docs state that Orbit executes the scorer's current default version.
- [Timeout ordering](#discussion_r3865624457): addressed. The server timeout is 80% of the HTTP deadline, and equal/greater explicit values fail locally before any request. Tests cover all branches.

The dependency-floor, invocation-resolved ADK tools, forward-compatible config, and deep `Step` snapshot threads are not addressed in this scoped change and remain open. Config strictness currently mirrors Orbit #1720's `extra="forbid"` contract.

namrataghadi-galileo:
@abhinav-galileo Follow-up on the four remaining review threads:

- Dependency floors (#discussion_r3865624436): addressed in `bdd30d5`. The coordinated 8.6.0 floors now cover engine models/evaluators, builtin models and the Galileo extra, Galileo models/evaluators, and SDK/server runtime evaluators. `evaluate_with_context(data, step)` remains the additive hook, so evaluators implementing only `evaluate(data)` stay compatible.
- Invocation-resolved ADK tools (#discussion_r3865624464): addressed in `bdd30d5`. Model callbacks now prefer the exact current `LlmRequest.tools_dict`, treat an empty resolved map as authoritative, fall back to the bind-time registry only for older ADK versions, and reuse the captured set for the matching post-model evaluation. Tests cover dynamic override and an empty invocation set.
- Forward-compatible scorer config (#discussion_r3865624472): no code change is needed for this rollout. Orbit #1720 is authoritative and defines `extra="forbid"` with only `threshold`, `score_threshold`, and `request_timeout_seconds`. AgentControl intentionally mirrors that allowlist and should extend it in lockstep with Orbit.
- Deep Step snapshots (#discussion_r3865624478): no code change is needed for the scorer-invoke rollout. Luna only reads the trusted request-scoped Step. True evaluator isolation would require copying both selector-selected data and the full Step per evaluator in the hot path, so that should be a separate design/performance change if we decide to guarantee deep isolation.

namrataghadi-galileo:
@abhinav-galileo Follow-up on the four remaining review threads:

- Dependency floors (#discussion_r3865624436): addressed in `bdd30d5`. The coordinated 8.6.0 floors now cover engine models/evaluators, builtin models and the Galileo extra, Galileo models/evaluators, and SDK/server runtime evaluators. `evaluate_with_context(data, step)` remains the additive hook, so evaluators implementing only `evaluate(data)` stay compatible.
- Invocation-resolved ADK tools (#discussion_r3865624464): addressed in `bdd30d5`. Model callbacks now prefer the exact current `LlmRequest.tools_dict`, treat an empty resolved map as authoritative, fall back to the bind-time registry only for older ADK versions, and reuse the captured set for the matching post-model evaluation. Tests cover dynamic override and an empty invocation set.
- Forward-compatible scorer config (#discussion_r3865624472): no code change is needed for this rollout. Orbit #1720 is authoritative and defines `extra="forbid"` with only `threshold`, `score_threshold`, and `request_timeout_seconds`. AgentControl intentionally mirrors that allowlist and should extend it in lockstep with Orbit.
- Deep Step snapshots (#discussion_r3865624478): no code change is needed for the scorer-invoke rollout. Luna only reads the trusted request-scoped Step. True evaluator isolation would require copying both selector-selected data and the full Step per evaluator in the hot path, so that should be a separate design/performance change if we decide to guarantee deep isolation.

engine/src/agent_control_engine/core.py
namrataghadi-galileo: Addressed in `bdd30d5`. All coordinated package floors now require 8.6.0, while `evaluate_with_context(data, step)` remains additive so existing `evaluate(data)` implementations stay compatible.

sdks/python/src/agent_control/integrations/google_adk/plugin.py
namrataghadi-galileo: Addressed in `bdd30d5`. The callbacks now use the invocation-resolved `LlmRequest.tools_dict`, preserve an authoritative empty set, use bind-time tools only as an older-ADK fallback, and retain the exact set for post-model scoring. Added tests cover both runtime override and empty resolution.

evaluators/contrib/galileo/src/agent_control_evaluator_galileo/luna/config.py
namrataghadi-galileo: No code change is needed for this rollout. Orbit #1720 is the schema source of truth and its `ScoringInvokeConfig` deliberately uses `extra="forbid"` with only `threshold`, `score_threshold`, and `request_timeout_seconds`. AgentControl mirrors that exact allowlist and will extend it in lockstep when Orbit adds supported fields.

models/src/agent_control_models/agent.py
namrataghadi-galileo: No code change is needed for this scorer-invoke rollout. `Step` is trusted request-scoped evaluator context and Luna only reads it, so shallow nested mutability does not affect the Orbit contract. True isolation would require copying both selector-selected data and the full Step per evaluator in the latency-sensitive fan-out; that should be a separate design/performance change if deep isolation becomes a supported guarantee.

abhinav-galileo (approved):

PR #260, opened 2026-08-21
feat(sdk): add optional traceprovider parameter

namrataghadi-galileo:
## Summary
- The built-in OTEL observability sink always created its own `TracerProvider`, exporter, and span processor. Applications that already run OTEL (e.g. Google ADK via `GoogleADKInstrumentor`) ended up with two disconnected pipelines, splitting application and control spans across separate export requests.
- Adds an optional `otel_tracer_provider` param to `agent_control.init()` so the built-in `otel` sink can reuse an application-owned or globally-registered SDK provider instead of always creating its own.

## Scope
- **User-facing/API changes:**
  - New optional `otel_tracer_provider` kwarg on `agent_control.init()` / `init_observability()`.
  - Applied only when `observability_sink_name="otel"`; ignored for `default`, `registered`, and other named sinks.
  - Provider resolution order: explicit param → valid globally-registered OTEL SDK `TracerProvider` (via `trace.get_tracer_provider()`) → Agent Control-owned provider (existing behavior).
  - The default OTEL proxy/no-op provider is never treated as a configured provider.
  - README documents a Google ADK shared-provider example.
- **Internal changes:**
  - `otel_sink.py`: `create_otel_control_event_sink()` accepts a `tracer_provider` kwarg, tracks `provider_owned`, and only attaches an Agent Control exporter/`BatchSpanProcessor` when it creates the provider itself. OTLP exporter import is now lazy and skipped entirely on the external-provider path.
  - `observability.py`: provider is threaded through a runtime-only global (`_configured_otel_tracer_provider`), not the JSON-typed `observability_sink_config`. Changing the explicit provider on reinitialization invalidates only the cached OTEL sink, leaving other sink caches untouched.
  - `OTELControlEventSink.close()` flushes external providers but no longer shuts them down; Agent Control-created providers are still shut down as before.
- **Out of scope:** Orbit, ingest-service, Galileo Python SDK, and customer application code were not touched. No changes to `default`, `registered`, or other named custom sink behavior.

## Risk and Rollout
- **Risk level:** low — the new parameter defaults to `None` and only affects behavior when `observability_sink_name="otel"`; all other sink paths are unchanged.
- **Rollback plan:** revert the commit; no data migrations, schema changes, or server-side changes are involved (Python SDK only).

## Testing
- [x] Added or updated automated tests — `test_otel_sink.py` (explicit-provider precedence, provider-owned vs. external, proxy/no-op rejection, no duplicate processor, trace/parent correlation, lifecycle/shutdown ownership, reinitialization invalidation) and `test_init_validation.py` (runtime-only path, not leaked into `sink_config`).
- [x] Ran equivalent of `make check`: `uv run pytest` (644 passed, 1 pre-existing unrelated failure requiring a live server, confirmed failing on the prior commit too), `uv run mypy src/agent_control` (clean), `uv run ruff check src/` (clean, matches the `make lint` target).
- [x] Manually verified behavior — traced through provider resolution order, shutdown/ownership branches, and cache invalidation logic against each acceptance criterion in the spec; no regressions found in default/registered/custom sink paths.

## Checklist
- [x] Linked issue/spec (SAO-16152 — reuse existing OTEL provider in Agent Control Python SDK)
- [x] Updated docs/examples for user-facing changes (README: Google ADK shared-provider example)
- [ ] Included any required follow-up tasks — none identified

### DIFF
--- a/sdks/python/README.md
+++ b/sdks/python/README.md
@@ -24,3 +24,38 @@ async def search_database(query: str) -> str:
 ```
 
 Controls are defined centrally and enforced automatically at runtime. See [Python SDK Documentation](https://docs.agentcontrol.dev/sdk/python-sdk) for complete reference.
+
+## Sharing an OpenTelemetry provider with Google ADK
+
+When Google ADK and Agent Control should export through the same OpenTelemetry
+pipeline, configure one SDK `TracerProvider` and pass it to both components:
+
+```python
+import agent_control
+from opentelemetry import trace
+from opentelemetry.exporter.otlp.proto.http.trace_exporter import OTLPSpanExporter
+from openinference.instrumentation.google_adk import GoogleADKInstrumentor
+from opentelemetry.sdk.resources import Resource
+from opentelemetry.sdk.trace import TracerProvider
+from opentelemetry.sdk.trace.export import BatchSpanProcessor
+
+resource = Resource.create({"service.name": "my-google-adk-agent"})
+provider = TracerProvider(resource=resource)
+provider.add_span_processor(BatchSpanProcessor(OTLPSpanExporter()))
+trace.set_tracer_provider(provider)
+
+GoogleADKInstrumentor().instrument(tracer_provider=provider)
+agent_control.init(
+    agent_name="my-google-adk-agent",
+    observability_enabled=True,
+    observability_sink_name="otel",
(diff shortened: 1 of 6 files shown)


savula15 (approved):
LGTM

soumya-io (approved):

PR #262, opened 2026-09-17
feat(sdk): add step_type as input parameter to the control decorator and evaluate_controls function

namrataghadi-galileo:
## Summary
- Added support for arbitrary custom step types such as retriever, trace, and session across the Python SDK, telemetry, observability models, TypeScript SDK, and UI API types.
Existing llm/tool inference remains unchanged.
This is to accomodate controls at different levels especially Luna but not restricted to Galileo levels only.

## Scope
User-facing/API changes:- Added step_type: str | None to @agent_control.control().
- Widened evaluate_controls(step_type) to accept custom strings.
- Updated generated TypeScript/UI API artifacts.

Internal changes:- Preserved custom decorator arguments and telemetry classifications.
- Added step-type normalization and validation.
- Added endpoint and regression tests.

Out of scope:- Google ADK and Strands adapters remain intentionally limited to LLM/tool lifecycle events.

## Risk and Rollout
Risk level: low
Rollback plan: Revert this commit.

## Testing
- [ x] Added or updated automated tests
- [ x] Ran `make check` (or explained why not)
- [x ] Manually verified behavior

## Checklist
- [ x] Updated docs/examples for user-facing changes


josjeon (commented):
Question:

So my understanding is that an unknown type/typo like step_type="sesion" passes through and matches no controls, and process() early returns is_safe=True, confidence=1.0 like:
  https://github.com/agentcontrol/agent-control/blob/bd90753a900f446f92bc82694f90823ed11acf64/engine/src/agent_control_engine/core.py#L642-L649
  so it reads the same as a genuine pass. Assuming that's intentional for now?

namrataghadi-galileo:
@josjeon Yes—that is intentional with the current generic string-based design.
step_type="sesion" is accepted as a non-empty custom type. If no controls apply to that type, the engine returns:
- is_safe=True
- confidence=1.0 when there are no precheck errors
- no matches
This preserves support for arbitrary custom step types, but means typos can silently bypass controls scoped to "session". Controls without a step-type restriction may still apply. A future strict mode or registered step-type vocabulary could detect typos, but that is not part of this change.

sdks/python/src/agent_control/validation.py
```diff
@@ -20,3 +20,10 @@ def ensure_agent_name(value: str, field_name: str = "agent_name") -> str:
             f"{field_name} may only contain lowercase letters, digits, ':', '_' or '-'"
         )
     return normalized
+
+
+def ensure_step_type(value: str, field_name: str = "step_type") -> str:
+    """Return a normalized non-empty step type or raise ``ValueError``."""
+    if not isinstance(value, str) or not value.strip():
+        raise ValueError(f"{field_name} must be a non-empty string")
+    return value.strip()
```
wrisa, line 29: normalize to lowercase ?

sdks/python/src/agent_control/evaluation_events.py
```diff
@@ -62,9 +61,13 @@ def observability_metadata(
...
 
-def map_applies_to(step_type: str) -> Literal["llm_call", "tool_call"]:
+def map_applies_to(step_type: str) -> str:
     """Map Agent Control step types to observability applies_to values."""
-    return "tool_call" if step_type == "tool" else "llm_call"
+    if step_type == "tool":
+        return "tool_call"
+    if step_type == "llm":
+        return "llm_call"
+    return step_type
```
wrisa: `_call` suffix for consistency ?

sdks/python/src/agent_control/integrations/google_adk/plugin.py
```diff
@@ -58,7 +58,12 @@
 
 
 class AgentControlPlugin(BasePlugin):
-    """Google ADK plugin that enforces Agent Control across model and tool hooks."""
+    """Google ADK plugin that enforces Agent Control across model and tool hooks.
+
+    Google ADK lifecycle callbacks expose model and tool operations only, so
+    this adapter intentionally uses the built-in ``llm`` and ``tool`` step
+    types. Custom step types remain available through the core SDK APIs.
+    """
```
wrisa, line 66: Do we need to modify `step_type` logic as well ?

sdks/python/src/agent_control/integrations/strands/plugin.py
```diff
@@ -71,6 +71,9 @@ class AgentControlPlugin(Plugin):
 
     The Agent Control server is required for control distribution and policy assignment.
     Controls may specify execution="sdk" or execution="server".
+    Strands lifecycle events handled by this adapter are model and tool events,
+    so its internal step-type annotations intentionally remain limited to
+    ``llm`` and ``tool``. Custom types remain available through the core SDK APIs.
```
wrisa, line 76: Do we need to modify `step_type` logic as well ?

sdks/python/src/agent_control/validation.py
namrataghadi-galileo: I’m intentionally normalizing whitespace only. Built-in inferred types are already lowercase (llm and tool), while custom step types remain case-sensitive and are matched exactly. Lowercasing would need to be applied consistently to Step.type, ControlScope.step_types, registry keys, and telemetry; otherwise values such as "Retriever" could become mismatched. We can add case-insensitive normalization as a separate API contract if desired.

sdks/python/src/agent_control/integrations/google_adk/plugin.py
namrataghadi-galileo: No additional change is needed here. Google ADK currently exposes model and tool lifecycle callbacks only, and each callback explicitly evaluates with step_type="llm" or step_type="tool". Custom step types remain supported through the core SDK decorator and direct evaluate_controls() API. We can extend this adapter if ADK later exposes custom lifecycle callbacks.

sdks/python/src/agent_control/integrations/strands/plugin.py
namrataghadi-galileo: No additional change is needed here. The Strands adapter currently handles model and tool lifecycle events only, so its internal step-type annotations intentionally remain limited to llm and tool. Custom step types remain available through control(step_type=...) and direct evaluate_controls() calls. We can extend this adapter if Strands adds custom lifecycle event types.

sdks/python/src/agent_control/validation.py
namrataghadi-galileo: I’m intentionally normalizing whitespace only. Built-in inferred types are already lowercase (llm and tool), while custom step types remain case-sensitive and are matched exactly. Lowercasing would need to be applied consistently to Step.type, ControlScope.step_types, registry keys, and telemetry; otherwise values such as "Retriever" could become mismatched. We can add case-insensitive normalization as a separate API contract if desired.

sdks/python/src/agent_control/evaluation_events.py
namrataghadi-galileo: Updated observability event classification so custom step types use the same _call suffix convention as the built-in types. For example, retriever, trace, and session now emit as retriever_call, trace_call, and session_call. Existing llm_call and tool_call behavior is unchanged. Control matching continues to use the raw step types. Added coverage for retriever, trace, and session classifications.

josjeon (approved):

wrisa (approved):

